PR #4, opened 2017-08-24
Add support for tokens.

andrewmchen:


databricks_cli/configure/cli.py
```diff
@@ -47,10 +55,19 @@ def configure_cli():
                             confirmation_prompt=True)
     if password == default_password:
         password = conf.password
-    config = DatabricksConfig(host, username, password)
+    config = DatabricksConfig.construct_from_password(host, username, password)
     config.overwrite()
 
 
+@click.command(context_settings=CONTEXT_SETTINGS)
+@click.option('--token', show_default=True, is_flag=True, default=False)
```
loneknightpy: how could user know there is an option to use token?

databricks_cli/configure/config.py
```diff
@@ -49,49 +50,73 @@ def decorator(*args, **kwargs):
...
+
+    @property
+    def is_valid_with_token(self):
+        return self._config.has_option(DEFAULT_SECTION, TOKEN) and \
+            self._config.has_option(DEFAULT_SECTION, HOST)
 
     @property
     def host(self):
-        return self.config.get(DEFAULT_SECTION, HOST) if self.is_valid else None
+        return self._config.get(DEFAULT_SECTION, HOST) if self.is_valid else None
```
loneknightpy: I don't see why we need is_valid check here.

loneknightpy: If a key is not what `self._config.get()` will return?

tests/configure/test_config.py
```diff
@@ -55,10 +57,46 @@ def test_function(x):
 
 
 class TestDatabricksConfig(object):
+    databricks_config_from_password = config.DatabricksConfig.construct_from_password('h', 'u', 'p')
```
loneknightpy: can you use more meaningful strings? e.g. `test-host`, `test-user`, etc.

databricks_cli/configure/cli.py
andrewmchen: Let's document it better in `README.rst`. I added documentation pointing users to the two ways to configure the CLI for authentication.

databricks_cli/configure/config.py
andrewmchen: The checks seem to be necessary here. If the key does not exist the get function will raise a `NoOptionError` exception.

tests/configure/test_config.py
andrewmchen: Sure.

loneknightpy:
lgtm

Issue #5, opened 2017-09-11
Add support for the Workspace API.


andrewmchen:
https://docs.databricks.com/api/latest/workspace.html to enable CI/CD with Databricks.

PR #6, opened 2017-09-11
Add support for the Workspace API

andrewmchen:
https://github.com/databricks/databricks-cli/issues/5

### DIFF
--- a/databricks_cli/sdk/service.py
+++ b/databricks_cli/sdk/service.py
@@ -457,3 +457,60 @@ class DbfsService(object):
             self.add_block(handle, base64.encodestring(block))
         self.close(handle)
         input_file.close()
+
+
+class WorkspaceService(object):
+    def __init__(self, client):
+        self.client = client
+
+    def mkdirs(self, path):
+        _data = {}
+        if path is not None:
+            _data['path'] = path
+        return self.client.perform_query('POST', '/workspace/mkdirs', data=_data)
+    
+    def list(self, path):
+        _data = {}
+        if path is not None:
+            _data['path'] = path
+        return self.client.perform_query('GET', '/workspace/list', data=_data)
+    
+    def import_workspace(self, path, format=None, language=None, content=None, overwrite=None):
+        _data = {}
+        if path is not None:
+            _data['path'] = path
+        if format is not None:
+            _data['format'] = format
+        if language is not None:
+            _data['language'] = language
+        if content is not None:
+            _data['content'] = content
+        if overwrite is not None:
+            _data['overwrite'] = overwrite
+        return self.client.perform_query('POST', '/workspace/import', data=_data)
+    
+    def export_workspace(self, path, format=None, direct_download=None):
+        _data = {}
+        if path is not None:
+            _data['path'] = path
+        if format is not None:
+            _data['format'] = format
+        if direct_download is not None:
+            _data['direct_download'] = direct_download
+        return self.client.perform_query('GET', '/workspace/export', data=_data)
+    
+    def delete(self, path, recursive=None):
+        _data = {}
+        if path is not None:
+            _data['path'] = path
+        if recursive is not None:
+            _data['recursive'] = recursive
+        return self.client.perform_query('POST', '/workspace/delete', data=_data)
+    
+    def get_status(self, path):
+        _data = {}
+        if path is not None:
+            _data['path'] = path
+        return self.client.perform_query('GET', '/workspace/get-status', data=_data)
+     
+
--- a/databricks_cli/workspace/api.py
+++ b/databricks_cli/workspace/api.py
@@ -1,0 +1,126 @@
+# Databricks CLI
+# Copyright 2017 Databricks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"), except
+# that the use of services to which certain application programming
+# interfaces (each, an "API") connect requires that the user first obtain
+# a license for the use of the APIs from Databricks, Inc. ("Databricks"),
+# by creating an account at www.databricks.com and agreeing to either (a)
+# the Community Edition Terms of Service, (b) the Databricks Terms of
+# Service, or (c) another written agreement between Licensee and Databricks
+# for the use of the APIs.
+#
+# You may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+from base64 import b64encode, b64decode
+
+import click
+
+from databricks_cli.configure.config import get_workspace_client
+from databricks_cli.dbfs.exceptions import LocalFileExistsException
+
+
+DIRECTORY = 'DIRECTORY'
+NOTEBOOK = 'NOTEBOOK'
+LIBRARY = 'LIBRARY'
+
+
+class WorkspaceFileInfo(object):
+    def __init__(self, path, object_type, language=None):
+        self.path = path
+        self.object_type = object_type
+        self.language = language
+
+    def to_row(self, is_long_form, is_absolute):
+        path = self.path if is_absolute else self.basename
+        if self.is_dir:
+            stylized_path = click.style(path, 'cyan')
+        elif self.is_library:
+            stylized_path = click.style(path, 'green')
+        else:
+            stylized_path = path
+        if is_long_form:
+            return [self.object_type, stylized_path, self.language]
+        else:
+            return [stylized_path]
+
+    @property
+    def is_dir(self):
+        return self.object_type == DIRECTORY
+
+    @property
+    def is_notebook(self):
+        return self.object_type == NOTEBOOK
+
+    @property
+    def is_library(self):
+        return self.object_type == LIBRARY
+
+    @property
+    def basename(self):
+        return os.path.basename(self.path)
+
+    @classmethod
+    def from_json(cls, deserialized_json):
+        return cls(**deserialized_json)
+
+
+def get_status(workspace_path):
+    workspace_client = get_workspace_client()
+    return WorkspaceFileInfo.from_json(workspace_client.get_status(workspace_path))
+
+
+def list_objects(workspace_path):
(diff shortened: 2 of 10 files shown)


databricks_cli/sdk/service.py
```diff
@@ -457,3 +457,60 @@ def upload_file(self, dbfs_path, local_path, overwrite=False):
             self.add_block(handle, base64.encodestring(block))
         self.close(handle)
         input_file.close()
+
+
+class WorkspaceService(object):
```
loneknightpy: Is this generated?

databricks_cli/workspace/api.py
```diff
@@ -0,0 +1,124 @@
...
+        else:
+            return [stylized_path]
+
+    @property
+    def is_dir(self):
+        return self.object_type == DIRECTORY
+
+    @property
+    def is_notebook(self):
+        return self.object_type == NOTEBOOK
```
loneknightpy: I think list api also return libraries.

databricks_cli/workspace/api.py
```diff
@@ -0,0 +1,124 @@
...
+
+def list_files(workspace_path):
+    workspace_api = get_workspace_client()
+    response = workspace_api.list(workspace_path)
+    # This case is necessary when we list an empty dir in the workspace.
+    # TODO(andrewmchen): We should make our API respond with a json with 'objects' field even
+    # in this case.
+    if 'objects' not in response:
+        return []
+    files = response['objects']
```
loneknightpy: files -> objects?

databricks_cli/workspace/api.py
```diff
@@ -0,0 +1,124 @@
...
+    return [WorkspaceFileInfo.from_json(f) for f in files]
+
+
+def mkdirs(workspace_path):
+    workspace_api = get_workspace_client()
+    workspace_api.mkdirs(workspace_path)
+
+
+def import_workspace(source_path, target_path, language, fmt, is_overwrite):
```
loneknightpy: fmt -> format?

databricks_cli/workspace/api.py
```diff
@@ -0,0 +1,124 @@
...
+    workspace_api = get_workspace_client()
+    if os.stat(source_path).st_size >= SIZE_LIMIT:
```
loneknightpy: Why do we handle size limit here? We may change the limit in server side.

databricks_cli/workspace/api.py
```diff
@@ -0,0 +1,124 @@
...
+    def basename(self):
+        return os.path.basename(self.path)
+
+    @classmethod
+    def from_json(cls, deserialized_json):
+        return cls(**deserialized_json)
+
+
+def get_status(workspace_path):
+    workspace_api = get_workspace_client()
```
loneknightpy: Can we rename workspace_api to workspace_client?

databricks_cli/workspace/types.py
```diff
@@ -0,0 +1,97 @@
...
+        ext = cls.get_extension(path)
+        if ext == '.scala':
+            return cls.SCALA
+        elif ext == '.py':
+            return cls.PYTHON
+        elif ext == '.sql':
+            return cls.SQL
+        elif ext == '.r':
+            return cls.R
+        elif ext == '.R':
```
loneknightpy: Why is R special?

databricks_cli/workspace/api.py
```diff
@@ -0,0 +1,124 @@
...
+    return WorkspaceFileInfo.from_json(workspace_api.get_status(workspace_path))
+
+
+def list_files(workspace_path):
```
loneknightpy: can we just use `list`? They are not really files.

databricks_cli/workspace/cli.py
```diff
@@ -0,0 +1,212 @@
...
+@click.command(context_settings=CONTEXT_SETTINGS)
+@click.argument('workspace_path')
+@require_config
+@eat_exceptions
+def mkdirs_cli(workspace_path):
+    """
+    Make directories in the Databricks Workspace.
+
+    Note that this function is subtly different than mkdir. Mkdir requires
+    the parent directory of the input to already exist. Mkdirs will create directories
```
loneknightpy: I think we can skip the first sentence.

databricks_cli/workspace/cli.py
```diff
@@ -0,0 +1,212 @@
...
+@click.argument('source_path')
+@click.argument('target_path')
+@click.option('--language', '-l', required=True, type=LanguageClickType())
+@click.option('--format', '-f', default=WorkspaceFormat.SOURCE, type=FormatClickType())
+@click.option('--overwrite', '-o', is_flag=True, default=False)
+@require_config
+@eat_exceptions
+def import_workspace_cli(source_path, target_path, language, format, overwrite): # NOQA
+    """
+    TODO
```
loneknightpy: why TODO?

tests/workspace/test_api.py
```diff
@@ -0,0 +1,127 @@
...
+            f.write('test')
+        api.import_workspace(test_file_path, TEST_WORKSPACE_PATH, TEST_LANGUAGE, TEST_FMT, is_overwrite=False)
+        import_workspace_mock = get_workspace_client.return_value.import_workspace
+        assert import_workspace_mock.call_count == 1
+        assert import_workspace_mock.call_args[0][0] == TEST_WORKSPACE_PATH
+        assert import_workspace_mock.call_args[0][1] == TEST_FMT
+        assert import_workspace_mock.call_args[0][2] == TEST_LANGUAGE
+        assert import_workspace_mock.call_args[0][3] == b64encode('test')
+        assert import_workspace_mock.call_args[0][4] == False
+
```
loneknightpy: add new line

tests/workspace/test_api.py
```diff
@@ -0,0 +1,127 @@
...
+def test_mkdirs():
+    with mock.patch('databricks_cli.workspace.api.get_workspace_client') as get_workspace_client:
+        api.mkdirs(TEST_WORKSPACE_PATH)
+        mkdirs_mock = get_workspace_client.return_value.mkdirs
+        assert mkdirs_mock.call_count == 1
+        assert mkdirs_mock.call_args[0][0] == TEST_WORKSPACE_PATH
+
+
+def test_import_workspace(tmpdir):
```
loneknightpy: can we also test import/export dir?

databricks_cli/sdk/service.py
andrewmchen: Yes it is.

databricks_cli/workspace/api.py
andrewmchen: list is a reserved keyword in python so it's best not to use it if possible.

databricks_cli/workspace/api.py
andrewmchen: Sure.

databricks_cli/workspace/api.py
andrewmchen: I can rename to `list_objects` though.

databricks_cli/workspace/api.py
andrewmchen: Format is a special keyword in python so I'd prefer not to use it.

tests/workspace/test_api.py
andrewmchen: Forgot to add this file.

databricks_cli/workspace/api.py
```diff
@@ -0,0 +1,126 @@
...
+            is_overwrite)
+
+
+def export_workspace(source_path, target_path, fmt, is_overwrite):
+    if os.path.exists(target_path) and not is_overwrite:
+        raise LocalFileExistsException()
+    workspace_client = get_workspace_client()
+    output = workspace_client.export_workspace(source_path, fmt)
+    content = output['content']
+    # Will truncate target_path.
```
loneknightpy: What do you mean truncate?

databricks_cli/workspace/cli.py
```diff
@@ -0,0 +1,224 @@
...
+@eat_exceptions
+def ls_cli(l, absolute, workspace_path):
+    """
+    List files in the Databricks Workspace
+    """
+    if len(workspace_path) == 0:
+        workspace_path = '/'
+    else:
+        workspace_path = workspace_path[0]
+    files = list_objects(workspace_path)
```
loneknightpy: objects?

databricks_cli/workspace/cli.py
loneknightpy: Mkdirs is clearly different from mkdir. I don't think we need to explain so much. `Mkdirs will create directories along the path to the argument directory.` should be enough.

databricks_cli/workspace/types.py
loneknightpy: Can you comment on this? Why do we only support upper case for R? I feel it is better to keep consistent.

andrewmchen: This was brought up in the design doc by @mengxr. I don't have a strong preference either way.

databricks_cli/workspace/cli.py
andrewmchen: Fixed.

databricks_cli/workspace/api.py
andrewmchen: I meant it will overwrite the target path. I can make it more clear.

databricks_cli/workspace/types.py
loneknightpy: We can make all of them case insensitive.

databricks_cli/workspace/cli.py
```diff
@@ -0,0 +1,221 @@
...
+def ls_cli(l, absolute, workspace_path):
+    """
+    List objects in the Databricks Workspace
+    """
+    if len(workspace_path) == 0:
+        workspace_path = '/'
+    else:
+        workspace_path = workspace_path[0]
+    objects = list_objects(workspace_path)
+    table = tabulate([o.to_row(is_long_form=l, is_absolute=absolute) for o in objects],
```
loneknightpy: o is hard to read, can you at least use obj?

databricks_cli/workspace/cli.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def _export_dir_helper(source_path, target_path, overwrite):
+    if os.path.isfile(target_path):
+        click.echo('{} exists as a file. Skipping this subtree {}'
+                   .format(target_path, source_path))
+        return
+    if not os.path.isdir(target_path):
+        os.makedirs(target_path)
+    for o in list_objects(source_path):
```
loneknightpy: ditto.

PR #9, opened 2017-09-18
Dockerfile

morkot:
Originally I was thinking about dev container that would help to avoid pollution of dev workstation. Later I could make more product-ish container (smaller image without unneeded dev tools etc.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,9 @@ WORKDIR /usr/src/databricks-cli
 COPY . .
 
 RUN pip install --upgrade pip && \
-    pip install -r dev-requirements.txt
-
-RUN ./lint.sh && \
+    pip install -r dev-requirements.txt && \
+    ./lint.sh && \
     pip install . && \
     pytest tests
+
+ENTRYPOINT [ "databricks" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM python:2.7
+
+WORKDIR /usr/src/databricks-cli
+
+COPY . .
+
+RUN pip install --upgrade pip && \
+    pip install -r dev-requirements.txt
+
+RUN ./lint.sh && \
+    pip install . && \
+    pytest tests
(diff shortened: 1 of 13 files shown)


andrewmchen:
Hi @morkot, thanks for making a PR! What's the purpose of this Docker image? Do you use it to run the tests?

morkot:
@andrewmchen tests are run only on the building stage. Then I use the container to run CLI in different environments without any collisions of python and libraries versions. Also, I plan to use it in Jenkins.

README.rst
```diff
@@ -42,6 +42,18 @@ and follow the prompts. The second way is to use a access token generated inside
 Databricks. To configure the CLI to use an access token run ``dbfs configure --token``
 and follow the prompts.
 
+Using Docker
+------------
+
+To build image run
+``docker build -t $(pwd | sed 's@^.*/\([^$]*\)$@\1@') .``
```
andrewmchen: could we use `basename $(pwd)` instead?

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+WORKDIR /usr/src/databricks-cli
+
+COPY . .
+
+RUN pip install --upgrade pip && \
+    pip install -r dev-requirements.txt
+
+RUN ./lint.sh && \
+    pip install . && \
+    pytest tests
```
andrewmchen: Can we combine the two RUNs and add a CMD for `databricks`?

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:2.7
+
+WORKDIR /usr/src/databricks-cli
+
+COPY . .
```
andrewmchen: nit: would it be more conventional to specify the target path absolutely?

morkot: Considering WORKDIR instruction above, I don't think so.

morkot:
@andrewmchen are going to merge it? :)

andrewmchen:
Done. Thanks for the contribution!

morkot:
np. 

Do you know anything about API/CLI for DBFS mount of S3 buckets? 
https://github.com/databricks/databricks-cli/issues/10

PR #20, opened 2017-10-10
Make exporting to directory munge target_path

andrewmchen:
We should make `databricks workspace export` accept a directory name as the target path. If the target is detected as a directory, we should munge the target_path to include the basename of the source.

For example `databricks workspace export /some-notebook .` should write a file `./some-notebook.scala` if some-notebook is a scala notebook.

### DIFF
--- a/databricks_cli/workspace/cli.py
+++ b/databricks_cli/workspace/cli.py
@@ -112,7 +112,8 @@ def export_workspace_cli(source_path, target_path, format, overwrite): # NOQA
     """
     if os.path.isdir(target_path):
         file_info = get_status(source_path)
-        assert file_info.is_notebook, 'Export can only be called on a notebook'
+        if not file_info.is_notebook:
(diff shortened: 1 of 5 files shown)


mengxr:
Implementation LGTM. Needs to fix tests.

databricks_cli/workspace/cli.py
```diff
@@ -110,6 +110,11 @@ def export_workspace_cli(source_path, target_path, format, overwrite): # NOQA
     format is documented at
     https://docs.databricks.com/api/latest/workspace.html#notebookexportformat.
     """
+    if os.path.isdir(target_path):
+        file_info = get_status(source_path)
+        assert file_info.is_notebook, 'Export can only be called on a notebook'
```
mengxr: This should always lead to a runtime error, but `assert` is not guaranteed to run.

mengxr:
LGTM

PR #22, opened 2017-10-25
Jobs CLI

andrewmchen:


databricks_cli/jobs/api.py
```diff
@@ -0,0 +1,48 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from databricks_cli.configure.config import get_jobs_client
+
+
+def create_job(json):
+    return get_jobs_client().client.perform_query('POST', '/jobs/create', data=json)
```
loneknightpy: why do we need to call perform_query directly?

databricks_cli/jobs/cli.py
```diff
@@ -0,0 +1,131 @@
...
+@require_config
+@eat_exceptions
+def get_cli(job_id):
+    click.echo(pretty_format(get_job(job_id)))
+
+
+@click.command(context_settings=CONTEXT_SETTINGS)
+@click.option('--job-id', required=True, type=int)
+@click.option('--jar-params', default=None)
+@click.option('--notebook-params', default=None)
```
loneknightpy: how about python jobs and spark submit job?

databricks_cli/jobs/api.py
andrewmchen: The jobs client doesn't have a nice way to invoke an API given the entire JSON. You always have to unpack the top level fields unfortunately.

databricks_cli/jobs/cli.py
andrewmchen: Let's add them once we document these fields in the API.

PR #23, opened 2017-10-25
Runs CLI

andrewmchen:


databricks_cli/utils.py
```diff
@@ -58,3 +58,17 @@ def error_and_quit(message):
 
 def pretty_format(json):
     return json_dumps(json, indent=2)
+
+
+def json_base(json_file, json, api):
```
loneknightpy: It doesn't make send to name it `json_base`? How could people know it actually send api request?

PR #24, opened 2017-10-25
Clusters CLI

andrewmchen:


databricks_cli/clusters/cli.py
```diff
@@ -0,0 +1,188 @@
...
+@click.option('--cluster-id', required=True)
+@require_config
+@eat_exceptions
+def restart_cli(cluster_id):
+    """
+    Restarts a Spark cluster given its id.
+
+    If the cluster is not currently in a RUNNING state, nothing will happen
+    """
+    restart_cluster(cluster_id)
```
loneknightpy: does it return different response to indicate success/failure?

andrewmchen: Yup it will use exitcode 1. This function does the magic

```
 36 def eat_exceptions(function):
 37     @six.wraps(function)
 38     def decorator(*args, **kwargs):
 39         try:
 40             return function(*args, **kwargs)
 41         except HTTPError as exception:
 42             if exception.response.status_code == 401:
 43                 error_and_quit('Your authentication information may be incorrect. Please '
 44                                'reconfigure with ``dbfs configure``')
 45             else:
 46                 error_and_quit(exception.response.content)
 47         except Exception as exception: # noqa
 48             if not DEBUG_MODE:
 49                 error_and_quit('{}: {}'.format(type(exception).__name__, str(exception)))
 50     decorator.__doc__ = function.__doc__
 51     return decorator```

andrewmchen:
No cluster edit was not in the design.

databricks_cli/clusters/cli.py
loneknightpy: I guess I mean how user could tell if the cluster is restarted or not. It seems the api doesn't tell you that.

loneknightpy:
lgtm

PR #25, opened 2017-10-26
Make databricks jobs reset match design.

andrewmchen:
`databricks jobs reset` should take two parameters --json and --job-id. Then it should materialize the request body sent to the reset endpoint with
```
{
  "job_id": job_id,
  "new_settings": json
}
```

### DIFF
--- a/databricks_cli/jobs/cli.py
+++ b/databricks_cli/jobs/cli.py
@@ -47,10 +47,25 @@ def create_cli(json_file, json):
 @click.option('--json-file', default=None,
               help='File containing json to POST to /jobs/create.')
 @click.option('--json', default=None, help='Displays absolute paths.')
+@click.option('--job-id', required=True, help='The job_id to reset')
 @require_config
 @eat_exceptions
-def reset_cli(json_file, json):
-    json_cli_base(json_file, json, reset_job)
+def reset_cli(json_file, json, job_id):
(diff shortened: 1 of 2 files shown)


databricks_cli/jobs/cli.py
```diff
@@ -47,10 +47,25 @@ def create_cli(json_file, json):
...
-    json_cli_base(json_file, json, reset_job)
+def reset_cli(json_file, json, job_id):
+    if not bool(json_file) ^ bool(json):
+        raise RuntimeError('Either --json-file or --json should be provided')
+    if json_file:
+        with open(json_file, 'r') as f:
+            json = f.read()
+    deser_json = json_loads(json)
+    if 'job_id' in deser_json:
+        raise RuntimeError('job_id should not be provided in the json spec. The json spec should ' +
```
loneknightpy: This check is not needed.

andrewmchen: It's a common error to put the job_id in the --json option so I thought it'd be nice to catch it in the client side.

loneknightpy: why is it invalid? If you send the request to server, it will accept it.

loneknightpy:
lgtm

fmms:
@andrewmchen , @loneknightpy this commit broke the documentation. Can you comment in #356 how to get that updated. Thanks.

PR #28, opened 2017-10-31
Update SDK

andrewmchen:
We should update the SDK to autogenerate code for new run-now endpoint parameters.


databricks_cli/sdk/objects.py
```diff
@@ -166,7 +166,8 @@ class JobSettings(dict):
     def __init__(self, name=None, existing_cluster_id=None, new_cluster=None, libraries=None,
```
loneknightpy: Is this file auto generated?

loneknightpy:
lgtm assuming the sdk is auto generated.

loneknightpy:
lgtm. Thanks for cleaning it up.

PR #31, opened 2017-10-31
Make jobs list sorted by name.

andrewmchen:


databricks_cli/jobs/cli.py
```diff
@@ -85,7 +85,7 @@ def _jobs_to_table(jobs_json):
     ret = []
     for j in jobs_json['jobs']:
         ret.append((j['job_id'], j['settings']['name']))
-    return ret
+    return sorted(ret, key=lambda t: t[1])
```
mengxr: sort by lowercase names (easier for human and same behavior as in UI)

Issue #36, opened 2017-11-06
Changing on filenames in importing


brenolacerda:
When I import my workspace from databricks, somes files have its name altered. Usually the last letter is missing.

andrewmchen:
Could you leave some more information? Are you moving files from Databricks to the local filesystem? The other way? What filenames are being altered?

brenolacerda:
So, I send some files with this command "databricks workspace import_dir -o PathOrigin /Users/myUserName/Projects/". But, I have a file called 'my_battery', when I import to databricks, this file have its name altered for 'my_batter'.

andrewmchen:
Is the file name exactly `my_battery`?

brenolacerda:
file names altered: run_ah_battery, ah_battery, ah_history

Seems to me databricks-cli deletes the last letter when this letter is 'y'.

brenolacerda:
We found the error.

If file name have last letters 'py' in python files, the last letters are deleted. (e.g. samplepy.py ~ sample.py)

The same error happens in sql files, and scala files.

andrewmchen:
I can reproduce the problem. Thanks for reporting it!

andrewmchen:
`'apy.py'.rstrip('.py') == 'a'` and not `apy`..

andrewmchen:
Closed by #46

PR #46, opened 2017-11-07
Fix bug in import_dir where we improperly strip off the extension.

andrewmchen:
Addresses #36.

### DIFF
--- a/tests/workspace/test_cli.py
+++ b/tests/workspace/test_cli.py
@@ -93,10 +93,10 @@ def test_import_dir_helper(tmpdir):
     """
     Copy from directory ``tmpdir`` with structure as follows
     - a (directory)
-      - b (scala)
-      - c (python)
-      - d (r)
-      - e (sql)
+      - b.scala (scala)
+      - c.py (python)
+      - d.r (r)
+      - e.sql (sql)
     - f (directory)
       - g (directory)
     """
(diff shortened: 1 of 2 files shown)


tests/workspace/test_cli.py
```diff
@@ -144,3 +144,25 @@ def test_import_dir_helper(tmpdir):
...
+
+def test_import_dir_rstrip(tmpdir):
+    """
+    Copy from directory ``tmpdir`` with structure as follows
+    - a (directory)
+      - test-py.py (python)
+    """
+    os.makedirs(os.path.join(tmpdir.strpath, 'a'))
+    with open(os.path.join(tmpdir.strpath, 'a', 'test-py.py'), 'wb') as f:
+        f.write('print 1 + 1')
```
loneknightpy: where did you check the content of this file?

andrewmchen: We don't check the content of the file. The test just checks to see if the paths have their extensions stripped correctly.

loneknightpy: If that is the case, you don't need any content here. You could just put `pass` here.

andrewmchen: Ok let's do that.

loneknightpy:
lgtm

Issue #37, opened 2017-11-06
databricks fs cp -r should have progress message.


andrewmchen:
We should make STDOUT for the recursive copy for dbfs resemble `databricks workspace import_dir`. In `databricks workspace import_dir` we show which files are being moved to which paths.

This progress message should also show which files were failed to be moved because the file already exists.

Issue #15, opened 2017-10-04
The cp command does not return an error code when the target resource already exists


edurdevic:
When copying a file to dbfs without `--overwrite` and the resource already exists on dbfs, no file is copied and no exception is raised.
This behaviour leads to expecting the cp was successful even when the cp action is not performed.

andrewmchen:
To be clear this is only for recursive copy right? Would https://github.com/databricks/databricks-cli/issues/37 solve the issue?

edurdevic:
Actually this problem is present even with non-recursive copy.
The following commands will not fail (will just output an error message), and the final content of 'dbfs:/test.txt' will be "v1".
`
echo "v1" > test.txt
dbfs cp test.txt dbfs:/
echo "v2" > test.txt
dbfs cp test.txt dbfs:/
echo $?
`
The command 'echo $?' will write 0, but it should be non zero.

andrewmchen:
Gotcha, this PR should address it. https://github.com/databricks/databricks-cli/pull/49

andrewmchen:
Should be addressed by https://github.com/databricks/databricks-cli/pull/49. Would appreciate if you tried QAing the change!

Build from master by doing
`pip install git+https://github.com/databricks/databricks-cli.git@master`

PR #49, opened 2017-11-08
Refactor DBFS module so we print progress for recursive copy.

andrewmchen:
We should attempt to refactor DBFS so commands like `dbfs cp -r local_dir dbfs:/andrew` will have a line printed for each file that is copied from local to dbfs.

A side effect of this refactoring is we now correctly return the 1 exit code for exceptional cases where we cannot copy the file because the remote file already exists. This addresses https://github.com/databricks/databricks-cli/issues/15 and https://github.com/databricks/databricks-cli/issues/37

### DIFF
--- a/databricks_cli/dbfs/cli.py
+++ b/databricks_cli/dbfs/cli.py
@@ -87,26 +87,6 @@ def rm_cli(recursive, dbfs_path):
     delete(dbfs_path, recursive)
 
 
-def copy_to_dbfs(src, dbfs_path_dst, overwrite):
-    try:
-        put_file(src, dbfs_path_dst, overwrite)
-    except HTTPError as e:
-        if e.response.json()['error_code'] == DbfsErrorCodes.RESOURCE_ALREADY_EXISTS:
-            click.echo(e.response.json())
-        else:
-            raise e
-
-
-def copy_from_dbfs(dbfs_path_src, dst, overwrite):
-    # Will truncate the local file if it exists.
-    try:
-        get_file(dbfs_path_src, dst, overwrite)
-    except LocalFileExistsException:
-        click.echo(('The host destination {} already exists. You should provide the '
-                   '--overwrite flag.')
-                   .format(dst))
-
-
 def copy_to_dbfs_non_recursive(src, dbfs_path_dst, overwrite):
     # Munge dst path in case dbfs_path_dst is a dir
     try:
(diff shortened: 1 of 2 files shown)


databricks_cli/dbfs/cli.py
```diff
@@ -140,7 +120,14 @@ def copy_to_dbfs_recursive(src, dbfs_path_dst, overwrite):
         if os.path.isdir(cur_src):
             copy_to_dbfs_recursive(cur_src, cur_dbfs_dst, overwrite)
         elif os.path.isfile(cur_src):
-            copy_to_dbfs(cur_src, cur_dbfs_dst, overwrite)
+            try:
+                put_file(cur_src, cur_dbfs_dst, overwrite)
+                click.echo('{} -> {}'.format(cur_src, cur_dbfs_dst))
+            except HTTPError as e:
+                if e.response.json()['error_code'] == DbfsErrorCodes.RESOURCE_ALREADY_EXISTS:
+                    click.echo(e.response.json())
```
mengxr: Shall we make the error message clearer? Right now it is

~~~
{u'message': u'A file or directory already exists at the input path dbfs:/home/meng/path.', u'error_code': u'RESOURCE_ALREADY_EXISTS'}
~~~

A more consistent message would be

~~~
dbfs:/home/meng/tmp/dbfsBytesRead/path already exists. Skip.
~~~

databricks_cli/dbfs/cli.py
```diff
@@ -87,26 +87,6 @@ def rm_cli(recursive, dbfs_path):
...
-            raise e
-
-
-def copy_from_dbfs(dbfs_path_src, dst, overwrite):
-    # Will truncate the local file if it exists.
-    try:
-        get_file(dbfs_path_src, dst, overwrite)
-    except LocalFileExistsException:
-        click.echo(('The host destination {} already exists. You should provide the '
-                   '--overwrite flag.')
```
mengxr: The `--overwrite` hint is missing w/ the new code, which could be quite helpful.

tests/dbfs/test_cli.py
```diff
@@ -0,0 +1,91 @@
...
+import mock
+from requests.exceptions import HTTPError
+from requests.models import Response
+
+import databricks_cli.dbfs.cli as cli
+from databricks_cli.dbfs.api import DbfsErrorCodes
+from tests.utils import get_callback, get_error_and_quit_mock, eat_error_and_quit
+
+
+def test_cp_cli_to_dbfs_non_recursive(tmpdir):
```
mengxr: The tests do not really belong to this PR. Could you create a separate PR?

PR #54, opened 2017-11-13
Added doc for how to find options to subcommands and how to run tests.

jric:


README.rst
```diff
@@ -74,6 +74,17 @@ reference `Aliasing Command Groups <#aliasing-command-groups>`_.
       mkdirs      Make directories in the Databricks Workspace.
       rm          Deletes objects from the Databricks...
 
+Each of the commands may have options of their own.  To see them, run, e.g.
+	$ datbricks workspace ls -h
```
andrewmchen: typo? Also could you drop all of this in a "code" block like above?

databricks_cli/workspace/cli.py
```diff
@@ -41,7 +41,7 @@
 @click.option('--absolute', is_flag=True, default=False,
               help='Displays absolute paths.')
 @click.option('-l', is_flag=True, default=False,
-              help='Displays full information including ObjectType, Language, Path')
+              help='Displays full information including ObjectType, Path, Language')
```
andrewmchen: Thanks!

tests/README.tests
```diff
@@ -0,0 +1,6 @@
+To run tests:
+
+* python -m pytest
+
+If using pydev you may see "undefined variable" errors when viewing some of the test files. Those can be
+safely ignored.
```
andrewmchen: Trailing newline.

tests/README.tests
```diff
@@ -0,0 +1,6 @@
+To run tests:
```
andrewmchen: Can you change this to a README.md?

andrewmchen: Or even better, I think it's best at TESTS.md

andrewmchen: Or even better CONTRIBUTING.md as top-level file looks pretty canonical.

tests/README.tests
```diff
@@ -0,0 +1,6 @@
...
+
```
andrewmchen: Could you also leave instructions on how to create a proper dev environment? Here is my attempt.

```
Setting up the dev environment
---------------------------------
To isolate dependencies from your global python installation, it is important to use a tool like
[virtualenv](https://virtualenv.pypa.io/en/stable/). With `virtualenv` you can install the dev environment by doing the following.

- `pip install -e .`
- `pip install -r dev-requirements.txt`

File: tests/README.tests [hunk 6]
```diff
@@ -0,0 +1,6 @@
...
+* python -m pytest
```
andrewmchen: If we put the file as a toplevel file. Invoking the tests with `python -m pytest` won't work. Can you put down the command `pytest tests` instead?

andrewmchen:
Thanks for the contribution! I've left some comments. LMK what you think.

PR #63, opened 2018-01-02
Fix build by explicit pylint version.

morkot:
If pylint version later then 1.7.4, then it complains with inconsistent-return-statements:

```
Messages
========

databricks_cli/configure/cli.py
  Line: 79
    pylint: inconsistent-return-statements / Either all return statements in a function should return an expression, or none of them should. (col 4)

databricks_cli/workspace/types.py
  Line: 36
    pylint: inconsistent-return-statements / Either all return statements in a function should return an expression, or none of them should. (col 4)
  Line: 52
    pylint: inconsistent-return-statements / Either all return statements in a function should return an expression, or none of them should. (col 4)
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ COPY . .
 
 RUN pip install --upgrade pip && \
     pip install -r dev-requirements.txt && \
+    pip list && \
     ./lint.sh && \
     pip install . && \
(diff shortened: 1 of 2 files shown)


andrewmchen:
LGTM

PR #64, opened 2018-01-02
Add support for write only configuration profiles.

andrewmchen:
This is part one of two for the configuration profile feature.

In this part, we add support for writing configurations with different profiles. In the next PR, we will add support for using this profile to authenticate to a Databricks workspace other than the one specified by the `DEFAULT` profile.

### DIFF
--- a/databricks_cli/configure/cli.py
+++ b/databricks_cli/configure/cli.py
@@ -25,8 +25,10 @@ import click
 
 from click import ParamType
 
+from databricks_cli.configure.provider import DatabricksConfig, update_and_persist_config, \
+    get_config_for_profile
 from databricks_cli.utils import CONTEXT_SETTINGS
-from databricks_cli.configure.config import DatabricksConfig, profile_option
+from databricks_cli.configure.config import profile_option
 
 
 PROMPT_HOST = 'Databricks Host (should begin with https://)'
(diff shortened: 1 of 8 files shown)


databricks_cli/configure/cli.py
```diff
@@ -35,41 +35,42 @@
 PROMPT_TOKEN = 'Token' #  NOQA
 
 
-def _configure_cli_token():
+def _configure_cli_token(profile):
```
loneknightpy: does it support consolidated shard? I think we need to set workspaceId for it.

databricks_cli/configure/config.py
```diff
@@ -51,30 +52,35 @@ def decorator(*args, **kwargs):
     return decorator
 
 
-def _get_api_client():
+def profile_option(f):
```
loneknightpy: Why don't we just document how to create config file? I don't think the command option is that useful for this. As a user I would be fine to copy paste and modify the token.

tests/configure/test_cli.py
```diff
@@ -0,0 +1,76 @@
...
+def test_configure_cli(tmpdir):
+    path = tmpdir.strpath
+    with mock.patch.object(DatabricksConfig, 'home', path):
+        runner = CliRunner()
+        runner.invoke(cli.configure_cli,
+                      input=(TEST_HOST + '\n' +
+                             TEST_USER + '\n' +
+                             TEST_PASSWORD + '\n' +
+                             TEST_PASSWORD + '\n'))
+        assert DatabricksConfig.fetch_from_fs().host(DEFAULT_SECTION) == TEST_HOST
```
loneknightpy: can we have some test that doesn't use DEFAULT_SECTION

tests/configure/test_config.py
```diff
@@ -57,42 +57,45 @@ def test_function(x):
...
+                                                   'test-token')
 
     def test_is_valid_true(self):
         assert self.databricks_config_from_password.is_valid
         assert self.databricks_config_from_token.is_valid
 
     def test_is_valid_false(self):
         databricks_config = config.DatabricksConfig()
-        assert not databricks_config.is_valid
+        assert not databricks_config.is_valid(config.DEFAULT_SECTION)
```
loneknightpy: Do we need `DEFAULT_SECTION` everywhere? Can we simply call `assert not databricks_config.is_valid()`

databricks_cli/configure/cli.py
andrewmchen: Yeah it should. The profile here is just to distinguish between different types of configurations. Since the DEFAULT profile already works with consolidated I don't see why other profiles wouldn't.

databricks_cli/configure/config.py
andrewmchen: I think that would be useful as well. Regarding the command option, `aws configure` supports the `--profile` option so this PR follows this convention.

tests/configure/test_cli.py
andrewmchen: `test_configure_two_sections` should test that behavior.

tests/configure/test_config.py
andrewmchen: IMO, the explicit parameter adds some clarity here.

databricks_cli/configure/config.py
```diff
@@ -90,57 +96,52 @@ def overwrite(self):
...
-        return self._config.has_option(DEFAULT_SECTION, TOKEN) and \
-            self._config.has_option(DEFAULT_SECTION, HOST)
+    def is_valid_with_token(self, profile):
+        return self._config.has_option(profile, TOKEN) and \
+            self._config.has_option(profile, HOST)
 
-    @property
-    def host(self):
-        return self._config.get(DEFAULT_SECTION, HOST) if self.is_valid else None
+    def host(self, profile):
```
mengxr: Seems more extensible if we implement a `DatabricksConfigProvider` class (or `Factory`) that has `def from_profile(profile):` method. Then the code changes to

~~~python
conf = DatabricksConfProvider.from_profile(profile)
api_client = ApiClient(conf)
~~~

PR #67, opened 2018-01-05
Refactor tests to use click.testing.CliRunner

andrewmchen:
Instead of calling the callback function for the Click command directly we should instead use the built in `CliRunner` utility. Using this also allows us to verify that no PRs make breaking changes to the user interface (option names, etc).


loneknightpy:
lgtm

PR #74, opened 2018-02-15
Add support for Libraries API

andrewmchen:
https://docs.databricks.com/api/latest/libraries.html

### DIFF
--- a/databricks_cli/libraries/cli.py
+++ b/databricks_cli/libraries/cli.py
@@ -1,0 +1,239 @@
+# Databricks CLI
+# Copyright 2017 Databricks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"), except
+# that the use of services to which certain application programming
+# interfaces (each, an "API") connect requires that the user first obtain
+# a license for the use of the APIs from Databricks, Inc. ("Databricks"),
+# by creating an account at www.databricks.com and agreeing to either (a)
+# the Community Edition Terms of Service, (b) the Databricks Terms of
+# Service, or (c) another written agreement between Licensee and Databricks
+# for the use of the APIs.
+#
+# You may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import click
+
+from databricks_cli.click_types import ClusterIdClickType, OneOfOption
+from databricks_cli.configure.config import require_config
+from databricks_cli.libraries.api import all_cluster_statuses, cluster_status, install_libraries, \
+    uninstall_libraries
+from databricks_cli.utils import CONTEXT_SETTINGS, eat_exceptions, pretty_format
+from databricks_cli.version import print_version_callback, version
+
+
+def _all_cluster_statuses():
+    click.echo(pretty_format(all_cluster_statuses()))
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Get the status of all libraries.')
+@require_config
+@eat_exceptions # noqa
+def all_cluster_statuses_cli():
+    """
+    Get the status of all libraries on all clusters. A status will be available for all libraries
+    installed on this cluster via the API or the libraries UI as well as libraries set to be
+    installed on all clusters via the libraries UI. If a library has been set to be installed on
+    all clusters, is_library_for_all_clusters will be true.
+    """
+    _all_cluster_statuses()
+
+
+def _cluster_status(cluster_id):
+    click.echo(pretty_format(cluster_status(cluster_id)))
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Get the status of all libraries for a specified cluster.')
+@click.option('--cluster-id', required=True, type=ClusterIdClickType(),
+              help=ClusterIdClickType.help)
+@require_config
+@eat_exceptions # noqa
+def cluster_status_cli(cluster_id):
+    """
+    Get the status of all libraries for a specified cluster. A status will be available for all
+    libraries installed on this cluster via the API or the libraries UI as well as libraries set to
+    be installed on all clusters via the libraries UI. If a library has been set to be installed on
+    all clusters, is_library_for_all_clusters will be true.
+    """
+    _cluster_status(cluster_id)
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Shortcut to `all-cluster-statuses` or `cluster-status`.')
+@click.option('--cluster-id', type=ClusterIdClickType(), default=None,
+              help=ClusterIdClickType.help)
+@require_config
+@eat_exceptions # noqa
+def list_cli(cluster_id):
+    """
+    Get the statsus of all libraries for all clusters or for a specified cluster.
+    If the option --cluster-id is provided, then all libraries on that cluster will be listed,
+    (cluster-status). If the option --cluster-id is omitted, then all libraries on all clusters
+    will be listed (all-cluster-statuses).
+    """
+    if cluster_id is not None:
+        _cluster_status(cluster_id)
+    else:
+        _all_cluster_statuses()
+
+
+INSTALL_OPTIONS = ['jar', 'egg', 'maven-coordinates', 'pypi-package', 'cran-package']
+JAR_HELP = 'JAR on DBFS or S3 or WASB.'
+EGG_HELP = 'Egg on DBFS or S3 or WASB.'
+MAVEN_COORDINATES_HELP = """
+Maven coordinates in the form of GroupId:ArtifactId:Version (i.e. org.jsoup:jsoup:1.7.2).
+"""
+MAVEN_REPO_HELP = """
+Maven repository to install the Maven package from. If omitted, both Maven 
+Repository and Spark Packages are searched.
+"""
+MAVEN_EXCLUSION_HELP = """
+List of dependences to exclude. For example: --maven-exclusion "slf4j:slf4j" 
+--maven-exclusion "*:hadoop-client".
+"""
+PYPI_PACKAGE_HELP = """
+The name of the PyPI package to install. An optional exact version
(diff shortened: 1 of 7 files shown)


databricks_cli/libraries/cli.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+INSTALL_OPTIONS = ['jar', 'egg', 'maven-coordinates', 'pypi-package', 'cran-package']
+JAR_HELP = 'JAR uploaded to dbfs or s3.'
+EGG_HELP = 'Egg uploaded to dbfs or s3.'
+MAVEN_COORDINATES_HELP = """
+Maven coordinates in the form of GroupId:ArtifactId:Version (i.e. org.jsoup:jsoup:1.7.2).
+"""
+MAVEN_REPO_HELP = """
+Maven repo to install the Maven package from. If omitted, both Maven
```
mengxr: * `repo` -> `repository`
* If provided, do we still search Maven Central and Spark Packages?

databricks_cli/libraries/cli.py
```diff
@@ -0,0 +1,263 @@
...
+
+INSTALL_OPTIONS = ['jar', 'egg', 'maven-coordinates', 'pypi-package', 'cran-package']
+JAR_HELP = 'JAR uploaded to dbfs or s3.'
+EGG_HELP = 'Egg uploaded to dbfs or s3.'
+MAVEN_COORDINATES_HELP = """
+Maven coordinates in the form of GroupId:ArtifactId:Version (i.e. org.jsoup:jsoup:1.7.2).
+"""
+MAVEN_REPO_HELP = """
+Maven repo to install the Maven package from. If omitted, both Maven 
+Repository and Spark Packages are searched.
```
mengxr: Should say `Maven Central Repository`.

databricks_cli/libraries/cli.py
```diff
@@ -0,0 +1,263 @@
...
+"""
+MAVEN_EXCLUSION_HELP = """
+List of dependences to exclude. For example: ["slf4j:slf4j", "*:hadoop-client"].
+"""
+PYPI_PACKAGE_HELP = """
+The name of the pypi package to install. An optional exact version
```
mengxr: minor: `PyPI`

databricks_cli/libraries/cli.py
```diff
@@ -0,0 +1,263 @@
...
+CRAN_PACKAGE_HELP = """
+The name of the CRAN package to install.
+"""
+CRAN_REPO_HELP = """
+The repository where the package can be found. If not specified, the default CRAN repo is used.
+"""
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Install libraries on a cluster.')
```
mengxr: `Install libraries` -> `Install a library`?

databricks_cli/libraries/cli.py
```diff
@@ -0,0 +1,263 @@
...
+EGG_HELP = 'Egg uploaded to dbfs or s3.'
+MAVEN_COORDINATES_HELP = """
+Maven coordinates in the form of GroupId:ArtifactId:Version (i.e. org.jsoup:jsoup:1.7.2).
+"""
+MAVEN_REPO_HELP = """
+Maven repo to install the Maven package from. If omitted, both Maven 
+Repository and Spark Packages are searched.
+"""
+MAVEN_EXCLUSION_HELP = """
+List of dependences to exclude. For example: ["slf4j:slf4j", "*:hadoop-client"].
```
mengxr: The example should be `--maven-exclusion "slf4j:slf4j" --maven-exclusion "*:hadoop-client"`

databricks_cli/libraries/cli.py
```diff
@@ -0,0 +1,263 @@
...
+@require_config
+@eat_exceptions # noqa
+def install_cli(cluster_id, jar, egg, maven_coordinates, maven_repo, maven_exclusion, pypi_package, # noqa
+                pypi_repo, cran_package, cran_repo):
+    """
+    Install libraries on a cluster. Libraries must be first uploaded to dbfs or s3
+    (see `dbfs cp -h`). Unlike the API, only one library can be installed for each execution of
+    `databricks libraries install`.
+
+    Only provide one of [--jar, --egg, --maven-coordinates, --pypi-package, --cran-package].
```
mengxr: "Users should only provide ...`

databricks_cli/libraries/cli.py
```diff
@@ -0,0 +1,263 @@
...
+        if cran_repo is not None:
+            cran_library['cran']['repo'] = cran_repo
+        libraries.append(cran_library)
+    assert len(libraries) == 1, 'Should have one library to install.'
+    install_libraries(cluster_id, libraries)
+
+
+def _uninstall_cli_exit_help(cluster_id):
+    click.echo("WARNING: Uninstalling libraries requires a cluster restart.")
+    click.echo("databricks clusters restart --cluster-id {}".format(cluster_id))
```
mengxr: Make it clear that this is a command user can run manually.

databricks_cli/libraries/cli.py
```diff
@@ -0,0 +1,263 @@
...
+@click.option('--cluster-id', required=True, type=ClusterIdClickType(),
+              help=ClusterIdClickType.help)
+@click.option('--all', is_flag=True, default=False, help='If set, uninstall all libraries.')
+@click.option('--jar', cls=MutuallyExclusiveOption, mutually_exclusive=INSTALL_OPTIONS,
+              help=JAR_HELP)
+@click.option('--egg', cls=MutuallyExclusiveOption, mutually_exclusive=INSTALL_OPTIONS,
+              help=EGG_HELP)
+@click.option('--maven-coordinates', cls=MutuallyExclusiveOption,
+              mutually_exclusive=INSTALL_OPTIONS, help=MAVEN_COORDINATES_HELP)
+@click.option('--maven-repo', help=MAVEN_REPO_HELP)
```
mengxr: Is it needed? What happens if I installed one library with non-default repo and then tried uninstalling it only using the coordinates?

databricks_cli/libraries/cli.py
```diff
@@ -0,0 +1,263 @@
...
+@eat_exceptions # noqa
+def uninstall_cli(cluster_id, all, jar, egg, maven_coordinates, maven_repo, maven_exclusion, # noqa
+                  pypi_package, pypi_repo, cran_package, cran_repo):
+    """
+    Mark libraries on a cluster to be uninstalled. Libraries which are marked to be uninstalled
+    will stay attached until the cluster is restarted. (see `databricks clusters restart -h`).
+    """
+    if all:
+        library_statuses = _cluster_status(cluster_id).get('library_statuses', [])
+        libraries = [l_status['library'] for l_status in library_statuses]
```
mengxr: Do we want to filter by LibraryInstallStatus? Curious what users see if this tries to uninstall a library that is in "FAILED" state.

databricks_cli/libraries/cli.py
```diff
@@ -0,0 +1,263 @@
...
+        uninstall_libraries(cluster_id, libraries)
+        _uninstall_cli_exit_help(cluster_id)
+        return
+
+    libraries = []
```
mengxr: This is duplicate code. Should do some refactoring.

databricks_cli/libraries/cli.py
```diff
@@ -0,0 +1,263 @@
...
+        if cran_repo is not None:
+            cran_library['cran']['repo'] = cran_repo
+        libraries.append(cran_library)
+    assert len(libraries) == 1, 'Should have one library to uninstall.'
+    uninstall_libraries(cluster_id, libraries)
+    _uninstall_cli_exit_help(cluster_id)
+
+
+@click.group(context_settings=CONTEXT_SETTINGS,
+             short_help='Utility to interact with jobs.')
```
mengxr: `with libraries`

tests/libraries/test_cli.py
```diff
@@ -0,0 +1,344 @@
...
+ALL_CLUSTER_STATUSES_RETURN = {
+    'statuses': [{
+        'library_statuses': [{
+            'status': 'INSTALLED',
+            'is_library_for_all_clusters': False,
+            'library': {
+                'jar': 'dbfs:/test.jar'
+            }
+        }],
+        'cluster_id': '0213-212348-veeps379'
```
mengxr: `TEST_CLUSTER_ID`

tests/utils.py
```diff
@@ -31,3 +31,7 @@ def wrapper(test, *args, **kwargs):
         conf.overwrite()
         return test(*args, **kwargs)
     return decorator.decorator(wrapper, test)
+
+
+def output_equals(actual, expected):
```
mengxr: * The method doesn't do what it claims. We can call it `assert_cli_output` and document the expected behavior.
* It is better to combine assert and output_equals into a single method so you can output the actual and expected if they don't match. Currently, you only get a boolean output.

databricks_cli/libraries/cli.py
andrewmchen: Not sure, I copied these docs from the API documentation.

databricks_cli/libraries/cli.py
andrewmchen: ditto: I copied these docs from the API documentation.

databricks_cli/libraries/cli.py
andrewmchen: I've made the warning red to make the second echo pop out more.

databricks_cli/libraries/cli.py
andrewmchen: I just checked. Unfortunately it is needed.

databricks_cli/libraries/cli.py
andrewmchen: I don't think anything will happen, like the documentation says `Uninstalling libraries that are not installed on the cluster will have no impact but is not an error.`

andrewmchen:
I've checked the `uninstall requires maven-exclusion` concerns you had. Unfortunately, we need to include all options like `--maven-repo` and `--maven-exclusion` to uninstall.

I still have to check the gist license issue with Justin.

tests/libraries/test_cli.py
```diff
@@ -0,0 +1,344 @@
...
+
+CLUSTER_STATUS_RETURN = {
+    'library_statuses': [{
+        'status': 'INSTALLED',
+        'is_library_for_all_clusters': False,
+        'library': {
+            'jar': 'dbfs:/test.jar',
+        }
+    }],
+    'cluster_id': '0213-212348-veeps379'
```
mengxr: nit: TEST_CLUSTER_ID

PR #77, opened 2018-02-20
Adding json_file option to allow notebook_params to be in JSON file

ghost:
Adding an option to allow notebook parameters to be stored in external JSON file and be called via run-now. I'd expect this to be similar to cluster definition JSON file except notebook parameters JSON file would need notebook_params as a top level key.  Something like this:
{
  "notebook_params" : {
    "parameter1": 0,
    "parameter2": 1
  },
}


nfx:
Please upgrade to the latest version. It seems that the version you're referring is too old.

PR #79, opened 2018-02-22
Add RST check to README.rst

mengxr:
The release page on PyPI doesn't render `README.rst` as expected since 0.4.0. This PR adds check to ensure RST is in correct format.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,7 +49,7 @@ Workspace CLI Examples
 The implemented commands for the Workspace CLI can be listed by running ``databricks workspace -h``.
 Commands are run by appending them to ``databricks workspace``. To make it easier to use the workspace
 CLI, feel free to alias ``databricks workspace`` to something shorter. For more information
-reference `Aliasing Command Groups <#aliasing-command-groups>`_.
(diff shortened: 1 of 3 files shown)


mengxr:
the test failure is expected

README.rst
```diff
@@ -49,7 +49,7 @@ Workspace CLI Examples
 The implemented commands for the Workspace CLI can be listed by running ``databricks workspace -h``.
 Commands are run by appending them to ``databricks workspace``. To make it easier to use the workspace
 CLI, feel free to alias ``databricks workspace`` to something shorter. For more information
-reference `Aliasing Command Groups <#aliasing-command-groups>`_.
+reference `Aliasing Command Groups section <#aliasing-command-groups>`_.
```
mengxr: Need to change the text because the section heading is using the same link text. Tried `:ref:` but it seems GitHub doesn't render it.

mengxr:
Confirmed that this fixed the issue: https://testpypi.python.org/pypi?:action=display&name=databricks-cli&version=0.5.0.dev1

PR #81, opened 2018-02-23
Add support for profile option on all commands

andrewmchen:
This is part two of two for the configuration profile feature.

In this part, we actually read out the different confs given by the different profiles and use them in our API calls.

### DIFF
--- a/databricks_cli/clusters/api.py
+++ b/databricks_cli/clusters/api.py
@@ -20,41 +20,36 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-from databricks_cli.configure.config import get_clusters_client
-
-
-def create_cluster(json):
-    return get_clusters_client().client.perform_query('POST', '/clusters/create', data=json)
+from databricks_cli.sdk import ClusterService
 
 
-def start_cluster(cluster_id):
-    return get_clusters_client().start_cluster(cluster_id)
+class ClusterApi(object):
+    def __init__(self, api_client):
+        self.client = ClusterService(api_client)
 
+    def create_cluster(self, json):
+        return self.client.client.perform_query('POST', '/clusters/create', data=json)
 
-def restart_cluster(cluster_id):
-    return get_clusters_client().restart_cluster(cluster_id)
+    def start_cluster(self, cluster_id):
+        return self.client.start_cluster(cluster_id)
 
+    def restart_cluster(self, cluster_id):
+        return self.client.restart_cluster(cluster_id)
 
-def delete_cluster(cluster_id):
-    return get_clusters_client().delete_cluster(cluster_id)
+    def delete_cluster(self, cluster_id):
+        return self.client.delete_cluster(cluster_id)
 
+    def get_cluster(self, cluster_id):
+        return self.client.get_cluster(cluster_id)
 
-def get_cluster(cluster_id):
-    return get_clusters_client().get_cluster(cluster_id)
+    def list_clusters(self):
+        return self.client.list_clusters()
 
+    def list_zones(self):
+        return self.client.list_available_zones()
 
-def list_clusters():
-    return get_clusters_client().list_clusters()
+    def list_node_types(self):
+        return self.client.list_node_types()
 
-
-def list_zones():
-    return get_clusters_client().list_available_zones()
-
-
-def list_node_types():
-    return get_clusters_client().list_node_types()
-
-
-def spark_versions():
-    return get_clusters_client().list_spark_versions()
+    def spark_versions(self):
+        return self.client.list_spark_versions()
--- a/databricks_cli/clusters/cli.py
+++ b/databricks_cli/clusters/cli.py
@@ -25,11 +25,10 @@ import click
 from tabulate import tabulate
 
 from databricks_cli.click_types import OutputClickType, JsonClickType, ClusterIdClickType
-from databricks_cli.clusters.api import create_cluster, start_cluster, restart_cluster, \
-    delete_cluster, get_cluster, list_clusters, list_zones, list_node_types, spark_versions
+from databricks_cli.clusters.api import ClusterApi
 from databricks_cli.utils import eat_exceptions, CONTEXT_SETTINGS, pretty_format, json_cli_base, \
     truncate_string
-from databricks_cli.configure.config import require_config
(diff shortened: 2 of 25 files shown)


databricks_cli/clusters/cli.py
```diff
@@ -38,54 +37,58 @@
...
-def create_cli(json_file, json):
+def create_cli(api_client, json_file, json):
     """
     Creates a Databricks cluster.
 
     The specification for the request json can be found at
     https://docs.databricks.com/api/latest/clusters.html#create
     """
-    json_cli_base(json_file, json, create_cluster)
+    json_cli_base(json_file, json, lambda json: ClusterApi(api_client).create_cluster(json))
```
mengxr: Could `ClusterApi(api_client)` be saved by a `@provide_cluster_api` decorator?

databricks_cli/configure/config.py
```diff
@@ -28,54 +28,39 @@
 
 from databricks_cli.configure.provider import DEFAULT_SECTION, get_config_for_profile
 from databricks_cli.utils import error_and_quit
-from databricks_cli.sdk import ApiClient, DbfsService, WorkspaceService, JobsService, \
-    ClusterService, ManagedLibraryService
+from databricks_cli.sdk import ApiClient
 
 
 def require_config(function):
```
mengxr: maybe call it `@provide_api_client`. it is what the method does and the CLI functions shouldn't care if it is built from profile or some other ways. 

nit: Also what if `api_client` is already given? I think in this case it should simply pass it over. It might help testing.

databricks_cli/configure/config.py
```diff
@@ -28,54 +28,39 @@
...
+    """
     @six.wraps(function)
     def decorator(*args, **kwargs):
-        config = get_config_for_profile(DEFAULT_SECTION)
+        profile = kwargs.pop('profile')
+        config = get_config_for_profile(profile)
         if not config.is_valid:
-            error_and_quit(('You haven\'t configured the CLI yet! '
-                            'Please configure by entering `{} configure`').format(sys.argv[0]))
+            if profile == DEFAULT_SECTION:
```
mengxr: We should move this section out as separate methods since what we need here is simply get a config and assert valid.

databricks_cli/configure/config.py
```diff
@@ -28,54 +28,39 @@
...
+
         return function(*args, **kwargs)
     decorator.__doc__ = function.__doc__
     return decorator
 
 
 def profile_option(f):
-    return click.option('--profile', required=False, default=DEFAULT_SECTION)(f)
+    return click.option('--profile', required=False, default=DEFAULT_SECTION,
+                        help='Connection configuration to use.')(f)
```
mengxr: CLI connection profile to use. See `databricks configure -h`. Should also mention that if this is case sensitive or not.

databricks_cli/dbfs/api.py
```diff
@@ -67,71 +68,61 @@ class DbfsErrorCodes(object):
...
-
-
-def move(dbfs_src, dbfs_dst):
-    dbfs_api = get_dbfs_client()
-    dbfs_api.move(dbfs_src.absolute_path, dbfs_dst.absolute_path)
+class DbfsApi(object):
+    def __init__(self, api_client):
+        self.client = DbfsService(api_client)
+
+    def list_files(self, dbfs_path):
```
mengxr: @andrewmchen any implementation code change in the methods? or this is basically copy & paste?

databricks_cli/dbfs/cli.py
```diff
@@ -42,9 +41,10 @@
 @click.option('-l', is_flag=True, default=False,
               help='Displays full information including size and file type.')
 @click.argument('dbfs_path', nargs=-1, type=DbfsPathClickType())
+@profile_option
 @require_config
```
mengxr: Same comment here, shall we generate the `DbfsApi` directly?

databricks_cli/workspace/api.py
```diff
@@ -75,56 +74,49 @@ def from_json(cls, deserialized_json):
...
-
-
-def delete(workspace_path, is_recursive):
-    workspace_client = get_workspace_client()
-    workspace_client.delete(workspace_path, is_recursive)
+class WorkspaceApi(object):
+    def __init__(self, api_client):
+        self.client = WorkspaceService(api_client)
+
+    def get_status(self, workspace_path):
```
mengxr: Same question: Any real implementation code changes here?

databricks_cli/clusters/api.py
```diff
@@ -20,41 +20,36 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from databricks_cli.sdk import ClusterService
 
-from databricks_cli.configure.config import get_clusters_client
 
+class ClusterApi(object):
```
mengxr: Not sure if `ClusterClient` is a better name here, no strong preference.

databricks_cli/clusters/cli.py
andrewmchen: IMO, it's a small optimization that doesn't seem worth the complexity of adding another decorator.

databricks_cli/dbfs/api.py
andrewmchen: This is basically a copy paste.

databricks_cli/configure/config.py
```diff
@@ -27,55 +27,34 @@
...
+
         return function(*args, **kwargs)
     decorator.__doc__ = function.__doc__
     return decorator
 
 
 def profile_option(f):
-    return click.option('--profile', required=False, default=DEFAULT_SECTION)(f)
+    return click.option('--profile', required=False, default=DEFAULT_SECTION,
+                        help='CLI connection profile to use.')(f)
```
mengxr: Mention whether this is case sensitive or not. Btw, we should think more about case sensitivity. Convert all letters to uppercase might result different meanings in some languages. And in the current implementation, we expect users use all uppercase letters if they modify the `~/.databrickscfg` file directly and add sections there. Say if I add a section call `[test]` it won't be pick up by `--profile test`, which is definitely not the expected behavior.

mengxr: Btw, `databricks fs ls --profile XXX` works but not `databricks --profile XXX fs ls` nor `databricks fs --profile XXX ls`. This is different from AWS CLI experience, which users are more familiar with. We can address this in a follow-up PR.

PR #58, opened 2017-12-10
Python 3 compatibility

andreabedini:
Mostly boring stuff. The hardest part was figuring out the test suite was causing havoc in the request package by creating a `Request` object with a string as `_content` rather than bytes. Included Dockerfile for Python 3. Test suite passes both in Python 2/3 🍻 

Fixes #2.

PS: I left in a change of the Dockerfile that allows caching the requirements. Let me know if you want me to pull that out.

PR #89, opened 2018-02-28
Minimal set of changes to get python 3 working

andrewmchen:
Made some tweaks to the code to get Databricks CLI working on Python 3. Tests passed but two commands still failed because of insufficient test coverage.
  - `databricks workspace import_dir`
  - `databricks fs cp local remote`
I applied fixes for both of these in this PR. I didn't add regression tests for these two cases yet because the error is in the SDK layer which we currently mock out of all unit tests...

Lots of ideas stolen from https://github.com/databricks/databricks-cli/pull/58/files. Thanks @andreabedini!

### DIFF
--- a/databricks_cli/dbfs/api.py
+++ b/databricks_cli/dbfs/api.py
@@ -99,7 +99,8 @@ class DbfsApi(object):
                 contents = local_file.read(BUFFER_SIZE_BYTES)
                 if len(contents) == 0:
                     break
-                self.client.add_block(handle, b64encode(contents))
+                # add_block should not take a bytes object.
+                self.client.add_block(handle, b64encode(contents).decode())
             self.client.close(handle)
 
     def get_file(self, dbfs_path, dst_path, overwrite):
--- a/databricks_cli/sdk/api_client.py
+++ b/databricks_cli/sdk/api_client.py
@@ -34,7 +34,7 @@ import warnings
 import requests
 import ssl
 
-import version
+from . import version
 
 from requests.adapters import HTTPAdapter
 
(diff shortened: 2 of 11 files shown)


databricks_cli/dbfs/api.py
```diff
@@ -99,7 +99,8 @@ def put_file(self, src_path, dbfs_path, overwrite):
                 contents = local_file.read(BUFFER_SIZE_BYTES)
                 if len(contents) == 0:
                     break
-                self.client.add_block(handle, b64encode(contents))
+                # add_block should not take a bytes object.
+                self.client.add_block(handle, b64encode(contents).decode())
```
mengxr: Is it just for py3 or a bug fix? If this is a correctness issue independent of py3 compatibility, we should fix it separately and make a bug fix release.

databricks_cli/sdk/api_client.py
```diff
@@ -62,16 +62,8 @@ class ApiClient(object):
...
-        if configUrl:
-            self.url = configUrl
-            params = self.performQuery("/", headers = {})[1]
-            params = credential.json
-            user = str(params["user"])
-            password = str(params["password"])
-            host = str(params["apiUrl"].split("/api")[0])
-
+    def __init__(self, user = None, password = None, host = None, token = None,
+                 apiVersion = version.API_VERSION, default_headers = {}, verify = True):
```
mengxr: Not part of this PR but `{}` shouldn't be the default value because it is immutable.

databricks_cli/sdk/api_client.py
```diff
@@ -62,16 +62,8 @@ class ApiClient(object):
     A partial Python implementation of dbc rest api
     to be used by different versions of the client.
     """
-    def __init__(self, user = None, password = None, host = None, token = None, configUrl = None,
-            apiVersion = version.API_VERSION, default_headers = {}, verify = True):
-        if configUrl:
```
mengxr: Why we no longer need this block?

tests/workspace/test_cli.py
```diff
@@ -47,7 +46,7 @@ def test_export_workspace_cli(workspace_api_mock, tmpdir):
     workspace_api_mock.get_status.return_value = \
         WorkspaceFileInfo('/notebook-name', NOTEBOOK, WorkspaceLanguage.SCALA)
     runner = CliRunner()
-    runner.invoke(cli.export_workspace_cli, ['--format', 'SOURCE', '/notebook-name', path])
+    print(runner.invoke(cli.export_workspace_cli, ['--format', 'SOURCE', '/notebook-name', path]).output)
```
mengxr: What does `print` do here?

databricks_cli/dbfs/api.py
andrewmchen: It's just for py3.

databricks_cli/sdk/api_client.py
andrewmchen: Turns out we just never used this piece of code. It's even a bit buggy. (credentials isn't defined).

tests/workspace/test_cli.py
andrewmchen: ah this was for debugging. i need to revert it now.

PR #95, opened 2018-03-08
Fix bug with `databricks libraries uninstall --all`.

andrewmchen:
`databricks libraries uninstall --all` just doesn't work. I've fixed the problems and also added a unit test to catch this error next time.


mengxr:
Thanks for the fix!

Issue #84, opened 2018-02-26
Should support `--profile` in different positions


mengxr:
Users expect all of the following work:

* `databricks --profile XXX fs ls`
* `databricks fs --profile XXX ls`
* `databricks fs ls --profile XXX`

However, only the last one works with the current implementation.

PR #99, opened 2018-03-19
Use --profile option anywhere.

andrewmchen:
Previously only commands like `databricks jobs list --profile <profile>` would work. Let's enable people to use the `profile` option in more places i.e.

- `databricks --profile <profile> jobs list`
- `databricks jobs --profile <profile> list`

This closes https://github.com/databricks/databricks-cli/issues/84

### DIFF
--- a/databricks_cli/click_types.py
+++ b/databricks_cli/click_types.py
@@ -23,6 +23,8 @@
 
 from click import ParamType, Option, MissingParameter, UsageError
 
+from databricks_cli.configure.provider import DEFAULT_SECTION
+
 
 class OutputClickType(ParamType):
     name = 'FORMAT'
@@ -79,3 +81,12 @@ class OneOfOption(Option):
         if len(cleaned_opts.intersection(set(self.one_of))) > 1:
             raise UsageError('Only one of {} should be provided.'.format(self.one_of))
         return super(OneOfOption, self).handle_parse_result(ctx, opts, args)
+
+
+class ContextObject(object):
+    def __init__(self):
+        self.profile = DEFAULT_SECTION
+
+    def set_profile(self, profile):
+        if self.profile == DEFAULT_SECTION:
+            self.profile = profile
--- a/databricks_cli/click_types.py
+++ b/databricks_cli/click_types.py
@@ -85,8 +85,14 @@ class OneOfOption(Option):
 
 class ContextObject(object):
     def __init__(self):
-        self.profile = DEFAULT_SECTION
+        self._profile = None
 
     def set_profile(self, profile):
-        if self.profile == DEFAULT_SECTION:
-            self.profile = profile
+        if self._profile is not None:
+            raise UsageError('--profile can only be provided once.')
+        self._profile = profile
+
+    def get_profile(self):
+        if self._profile is None:
+            return DEFAULT_SECTION
+        return self._profile
--- a/databricks_cli/click_types.py
+++ b/databricks_cli/click_types.py
@@ -89,7 +89,8 @@ class ContextObject(object):
 
     def set_profile(self, profile):
         if self._profile is not None:
-            raise UsageError('--profile can only be provided once.')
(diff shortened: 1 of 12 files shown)


databricks_cli/click_types.py
```diff
@@ -79,3 +81,12 @@ def handle_parse_result(self, ctx, opts, args):
         if len(cleaned_opts.intersection(set(self.one_of))) > 1:
             raise UsageError('Only one of {} should be provided.'.format(self.one_of))
         return super(OneOfOption, self).handle_parse_result(ctx, opts, args)
+
+
+class ContextObject(object):
```
mengxr: Maybe we could keep this one simple by not understanding the default section name. Just store the value provided by users, or `None` otherwise.

databricks_cli/configure/config.py
```diff
@@ -47,11 +49,19 @@ def decorator(*args, **kwargs):
...
+def profile_option_injected(f):
     return click.option('--profile', required=False, default=DEFAULT_SECTION,
                         help='CLI connection profile to use.')(f)
 
 
+def profile_option(f):
+    def callback(ctx, param, value): #  NOQA
+        context_object = ctx.ensure_object(ContextObject)
+        context_object.set_profile(value)
+    return click.option('--profile', required=False, default=DEFAULT_SECTION, callback=callback,
```
mengxr: Should mention which one takes over if there are multiple `--profile` specified.

databricks_cli/configure/cli.py
```diff
@@ -64,7 +63,7 @@ def _configure_cli_password(profile):
 @click.command(context_settings=CONTEXT_SETTINGS,
                short_help='Configures host and authentication info for the CLI.')
 @click.option('--token', show_default=True, is_flag=True, default=False)
-@profile_option
+@profile_option_injected
```
mengxr: This would make `databricks --profile XXX configure` update the default section. I would rename `@profile_option` to `@profile_option_to_ctx` and then provide a util method to extract the profile stored in it or using default.

andrewmchen: Yes that's a great point. I think we can get it out of ctx inside the implementation of this function.

mengxr: Btw, it would be fine to throw an error if a user tries to set the profile multiple times. Always safe to start with the most restrictive behavior.

databricks_cli/click_types.py
andrewmchen: I've addressed your comments by doing the following. The default value for the `--profile` option is None. If it is not NONE we set the global `ContextObject`'s `._profile`. If we set it twice, then throw an exception.

mengxr: It seems `click` doesn't provide a way to tell if the value was explicitly specified by users or the default. Then the solution looks good to me.

databricks_cli/click_types.py
```diff
@@ -79,3 +81,18 @@ def handle_parse_result(self, ctx, opts, args):
...
+    def __init__(self):
+        self._profile = None
+
+    def set_profile(self, profile):
+        if self._profile is not None:
+            raise UsageError('--profile can only be provided once.')
```
mengxr: useful to mention both profiles in the error message.

databricks_cli/configure/config.py
```diff
@@ -47,9 +48,19 @@ def decorator(*args, **kwargs):
...
+
 def profile_option(f):
-    return click.option('--profile', required=False, default=DEFAULT_SECTION,
-                        help='CLI connection profile to use.')(f)
+    def callback(ctx, param, value): #  NOQA
+        if value is not None:
+            context_object = ctx.ensure_object(ContextObject)
+            context_object.set_profile(value)
+    return click.option('--profile', required=False, default=None, callback=callback,
+                        expose_value=False, help='CLI connection profile to use.')(f)
```
mengxr: Should mention the default value in the help message. Seems not displayed previously either.

PR #104, opened 2018-04-10
Add SDK for secret manager

kunmin-db:
This PR extends the SDK with `SecretService` to interact with Databricks secret manager by the [API](https://docs.azuredatabricks.net/api/latest/secrets.html).

### DIFF
--- a/databricks_cli/sdk/api_client.py
+++ b/databricks_cli/sdk/api_client.py
@@ -79,10 +79,7 @@ class ApiClient(object):
         else:
             auth = {}
         user_agent = {'user-agent': 'databricks-cli-{v}'.format(v=databricks_cli_version)}
-        self.default_headers = {}
-        self.default_headers.update(auth)
-        self.default_headers.update(default_headers)
-        self.default_headers.update(user_agent)
+        self.default_headers = dict(auth.items() + default_headers.items() + user_agent.items())
         self.verify = verify
 
     def close(self):
--- a/databricks_cli/sdk/api_client.py
+++ b/databricks_cli/sdk/api_client.py
@@ -79,7 +79,10 @@ class ApiClient(object):
         else:
             auth = {}
         user_agent = {'user-agent': 'databricks-cli-{v}'.format(v=databricks_cli_version)}
-        self.default_headers = dict(auth.items() + default_headers.items() + user_agent.items())
+        self.default_headers = {}
+        self.default_headers.update(auth)
+        self.default_headers.update(default_headers)
+        self.default_headers.update(user_agent)
         self.verify = verify
 
     def close(self):
--- a/databricks_cli/sdk/service.py
+++ b/databricks_cli/sdk/service.py
@@ -352,6 +352,25 @@ class ClusterService(object):
         _data = {}
     
         return self.client.perform_query('GET', '/clusters/list-zones', data=_data)
+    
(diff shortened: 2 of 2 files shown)


databricks_cli/sdk/api_client.py
```diff
@@ -79,10 +79,7 @@ def __init__(self, user = None, password = None, host = None, token = None,
         else:
             auth = {}
         user_agent = {'user-agent': 'databricks-cli-{v}'.format(v=databricks_cli_version)}
-        self.default_headers = {}
-        self.default_headers.update(auth)
-        self.default_headers.update(default_headers)
-        self.default_headers.update(user_agent)
+        self.default_headers = dict(auth.items() + default_headers.items() + user_agent.items())
```
andrewmchen: Why did we change this line?

mengxr:
@kunmin-db The secrets API is still in preview. Do we want to hold until its GA?

databricks_cli/sdk/service.py
```diff
@@ -352,6 +352,25 @@ def list_available_zones(self):
         _data = {}
     
         return self.client.perform_query('GET', '/clusters/list-zones', data=_data)
+    
+    def get_events(self, cluster_id, start_time=None, end_time=None, order=None, event_types=None,
```
ericwang-db: We shouldn't include this since it's not part of our change to add the Secrets API. My guess is clusters team added a new API endpoint, and this was autogenerated.

databricks_cli/sdk/service.py
```diff
@@ -522,3 +541,83 @@ def get_status(self, path):
...
+    
+    def delete_scope(self, scope):
+        _data = {}
+        if scope is not None:
+            _data['scope'] = scope
+        return self.client.perform_query('POST', '/secret/scopes/delete', data=_data)
+    
+    def list_scopes(self):
+        _data = {}
+
```
ericwang-db: nit: extra newline

databricks_cli/sdk/service.py
yunzhangit: +1 we should probably remove this

ericwang-db:
@mengxr This should be OK, since the underlying non-preview prefix APIs are available (just undocumented). DBUtils already leverages them since we want to be able to deprecate /preview support without requiring that customers to upgrade their clients.

databricks_cli/sdk/service.py
kunmin-db: Maybe we should keep it in order to be consistent with the other parts, as it's autogenerated

mengxr:
@ericwang-db @kunmin-db if you are comfortable with GAing the feature, shall we move it out from `preview`, GA it first in this release, and then add it to CLI?

ericwang-db:
@mengxr The endpoints are already available without the preview prefix (i.e. you can hit /api/2.0/preview/secrets/scope/list and /api/2.0/secrets/scope/list). We are still working on other aspects before we can GA this feature with Microsoft, namely defining SLA/KPI.

mengxr:
That shouldn't be called "available" as there exists no public documentation for it. In CLI, we try our best to mirror our REST API, so we do not have maintenance burden here. We might consider a "preview" subcommand in CLI and put "secrets" under it. Once we GAed it, we move it out as in REST API.

PR #105, opened 2018-04-10
Added "secrets" subcommand to interact with secret manager

kunmin-db:
Added the CLI subcommand "secrets", including commands to read/write/delete on scopes/secret/ACL.

This PR is based on SDK changes in #104.

### DIFF
--- a/databricks_cli/secrets/api.py
+++ b/databricks_cli/secrets/api.py
@@ -20,39 +20,69 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+import six
+from requests.exceptions import HTTPError
+
 from databricks_cli.sdk import SecretService
+from databricks_cli.utils import error_and_quit
+
+
+def eat_404_exception(function):
+    @six.wraps(function)
+    def decorator(*args, **kwargs):
+        try:
+            return function(*args, **kwargs)
+        except HTTPError as exception:
+            if exception.response.status_code == 404:
+                error_and_quit('Secrets API is not supported for this workspace.'
+                               ' Please contact Databricks support.')
+            else:
+                raise exception
+    decorator.__doc__ = function.__doc__
+    return decorator
 
 
 class SecretApi(object):
     def __init__(self, api_client):
         self.client = SecretService(api_client)
 
+    @eat_404_exception
     def create_scope(self, scope, initial_manage_acl):
         return self.client.create_scope(scope, initial_manage_acl)
 
+    @eat_404_exception
     def delete_scope(self, scope):
         return self.client.delete_scope(scope)
 
+    @eat_404_exception
     def list_scopes(self):
         return self.client.list_scopes()
 
+    @eat_404_exception
     def write_secret(self, scope, key, string_value, bytes_value):
         return self.client.write_secret(scope, key, string_value, bytes_value)
 
+    @eat_404_exception
     def delete_secret(self, scope, key):
         return self.client.delete_secret(scope, key)
 
+    @eat_404_exception
     def list_secrets(self, scope):
         return self.client.list_secrets(scope)
 
+    @eat_404_exception
     def write_acl(self, scope, principal, permission):
         return self.client.write_acl(scope, principal, permission)
 
+    @eat_404_exception
     def delete_acl(self, scope, principal):
         return self.client.delete_acl(scope, principal)
 
+    @eat_404_exception
     def list_acls(self, scope):
         return self.client.list_acls(scope)
 
+    @eat_404_exception
     def get_acl(self, scope, principal):
         return self.client.get_acl(scope, principal)
--- a/databricks_cli/secrets/api.py
+++ b/databricks_cli/secrets/api.py
@@ -21,68 +21,39 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import six
-from requests.exceptions import HTTPError
-
 from databricks_cli.sdk import SecretService
-from databricks_cli.utils import error_and_quit
-
-
-def eat_404_exception(function):
-    @six.wraps(function)
-    def decorator(*args, **kwargs):
-        try:
-            return function(*args, **kwargs)
-        except HTTPError as exception:
-            if exception.response.status_code == 404:
-                error_and_quit('Secret manager API is not supported in this workspace.'
-                               ' Please try with different profile or contact Databricks support.')
-            else:
-                raise exception
-    decorator.__doc__ = function.__doc__
-    return decorator
 
 
 class SecretApi(object):
     def __init__(self, api_client):
         self.client = SecretService(api_client)
 
-    @eat_404_exception
     def create_scope(self, scope, initial_manage_acl):
         return self.client.create_scope(scope, initial_manage_acl)
 
-    @eat_404_exception
     def delete_scope(self, scope):
         return self.client.delete_scope(scope)
 
-    @eat_404_exception
     def list_scopes(self):
         return self.client.list_scopes()
 
-    @eat_404_exception
     def write_secret(self, scope, key, string_value, bytes_value):
         return self.client.write_secret(scope, key, string_value, bytes_value)
 
-    @eat_404_exception
     def delete_secret(self, scope, key):
         return self.client.delete_secret(scope, key)
 
-    @eat_404_exception
     def list_secrets(self, scope):
         return self.client.list_secrets(scope)
 
-    @eat_404_exception
     def write_acl(self, scope, principal, permission):
         return self.client.write_acl(scope, principal, permission)
 
-    @eat_404_exception
     def delete_acl(self, scope, principal):
         return self.client.delete_acl(scope, principal)
 
-    @eat_404_exception
     def list_acls(self, scope):
         return self.client.list_acls(scope)
 
-    @eat_404_exception
     def get_acl(self, scope, principal):
         return self.client.get_acl(scope, principal)
--- a/databricks_cli/secrets/api.py
+++ b/databricks_cli/secrets/api.py
@@ -28,8 +28,8 @@ class SecretApi(object):
     def __init__(self, api_client):
         self.client = SecretService(api_client)
 
-    def create_scope(self, scope, initial_manage_acl):
-        return self.client.create_scope(scope, initial_manage_acl)
+    def create_scope(self, scope, initial_manage_principal):
+        return self.client.create_scope(scope, initial_manage_principal)
 
     def delete_scope(self, scope):
         return self.client.delete_scope(scope)
--- a/databricks_cli/secrets/api.py
+++ b/databricks_cli/secrets/api.py
@@ -1,0 +1,58 @@
+# Databricks CLI
+# Copyright 2017 Databricks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"), except
+# that the use of services to which certain application programming
+# interfaces (each, an "API") connect requires that the user first obtain
+# a license for the use of the APIs from Databricks, Inc. ("Databricks"),
+# by creating an account at www.databricks.com and agreeing to either (a)
+# the Community Edition Terms of Service, (b) the Databricks Terms of
+# Service, or (c) another written agreement between Licensee and Databricks
+# for the use of the APIs.
+#
+# You may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from databricks_cli.sdk import SecretService
+
+
+class SecretApi(object):
+    def __init__(self, api_client):
+        self.client = SecretService(api_client)
+
+    def create_scope(self, scope, initial_manage_acl):
+        return self.client.create_scope(scope, initial_manage_acl)
+
+    def delete_scope(self, scope):
+        return self.client.delete_scope(scope)
+
+    def list_scopes(self):
+        return self.client.list_scopes()
+
+    def write_secret(self, scope, key, string_value, bytes_value):
+        return self.client.write_secret(scope, key, string_value, bytes_value)
+
+    def delete_secret(self, scope, key):
+        return self.client.delete_secret(scope, key)
+
+    def list_secrets(self, scope):
+        return self.client.list_secrets(scope)
+
+    def write_acl(self, scope, principal, permission):
+        return self.client.write_acl(scope, principal, permission)
+
+    def delete_acl(self, scope, principal):
+        return self.client.delete_acl(scope, principal)
+
+    def list_acls(self, scope):
+        return self.client.list_acls(scope)
+
+    def get_acl(self, scope, principal):
+        return self.client.get_acl(scope, principal)
--- a/databricks_cli/secrets/api.py
+++ b/databricks_cli/secrets/api.py
@@ -35,8 +35,8 @@ def eat_404_exception(function):
             return function(*args, **kwargs)
         except HTTPError as exception:
             if exception.response.status_code == 404:
-                error_and_quit('Secrets API is not supported for this workspace.'
-                               ' Please contact Databricks support.')
+                error_and_quit('Secret manager API is not supported in this workspace.'
+                               ' Please try with different profile or contact Databricks support.')
             else:
                 raise exception
     decorator.__doc__ = function.__doc__
--- a/databricks_cli/secrets/api.py
+++ b/databricks_cli/secrets/api.py
@@ -57,3 +57,4 @@ class SecretApi(object):
 
     def get_acl(self, scope, principal):
         return self.client.get_acl(scope, principal)
+
--- a/databricks_cli/secrets/api.py
+++ b/databricks_cli/secrets/api.py
@@ -57,4 +57,3 @@ class SecretApi(object):
 
     def get_acl(self, scope, principal):
         return self.client.get_acl(scope, principal)
-
--- a/databricks_cli/secrets/cli.py
+++ b/databricks_cli/secrets/cli.py
@@ -21,14 +21,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import base64
 import click
 from tabulate import tabulate
 
-from databricks_cli.click_types import OutputClickType, OneOfOption
+from databricks_cli.click_types import OutputClickType
 from databricks_cli.secrets.api import SecretApi
 from databricks_cli.utils import eat_exceptions, CONTEXT_SETTINGS, pretty_format, truncate_string, \
-    translate_value
+    error_and_quit, is_base64_str
 from databricks_cli.configure.config import provide_api_client, profile_option
 from databricks_cli.version import print_version_callback, version
 
@@ -42,24 +41,25 @@ VALUE_OPTIONS = ['string-value', 'bytes-value']
 @click.command(context_settings=CONTEXT_SETTINGS,
                short_help="Creates a secret scope.")
 @click.option('--scope', required=True)
-@click.option('--initial-manage-acl',
-              type=click.Choice(['CREATOR_ONLY', 'ALL_USERS']), default='CREATOR_ONLY',
-              help='The initial ACL applied to the secret scope. Must be either "CREATOR_ONLY"'
-                   ' or "ALL_USERS". Defaults to "CREATOR_ONLY".')
+@click.option('--initial-manage-principal',
+              help='The initial principal that can manage the created secret scope.')
 @profile_option
 @eat_exceptions
 @provide_api_client
-def create_scope(api_client, scope, initial_manage_acl):
+def create_scope(api_client, scope, initial_manage_principal):
     """
     Creates a new secret scope with given name.
 
-    "initial_manage_acl" controls the initial ACL applied to the secret scope.
-    If "CREATOR_ONLY", the initial ACL applied to scope is MANAGE permission, assigned to
-    the request issuer's user id.
-    If "ALL_USERS", the initial ACL applied to scope is MANAGE permission, assigned to
-    the group "all-users".
+    If "initial_manage_principal" is specified, the initial ACL applied to the scope is
+    applied to the supplied principal (user or group) with MANAGE permissions.
+    The only supported principal for this option is the group "users", which
+    contains all users in the workspace. If "initial_manage_principal" is not specified,
+    the initial ACL with MANAGE permission applied to the scope is assigned to the
+    API request issuer's user identity.
     """
-    SecretApi(api_client).create_scope(scope, initial_manage_acl)
+    if initial_manage_principal is not None and initial_manage_principal != "users":
+        error_and_quit("The only supported principal for --initial-manage-principal is users")
+    SecretApi(api_client).create_scope(scope, initial_manage_principal)
 
 
 def _scopes_to_table(scopes_json):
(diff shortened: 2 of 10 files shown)


yunzhangit:
lgtm

andrewmchen:
Can you rebase this on master?

mengxr:
@kunmin-db @yunzhangit Let's hold and discuss if we want to merge this before GAing the REST API.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+SECRET_HEADER = ('Key name', 'Last updated')
+ACL_HEADER = ('Principal', 'Permission')
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help="Creates a scope")
+@click.option('--scope', required=True)
+@click.option('--initial-manage-acl',
+              type=click.Choice(['creator-only', 'all-users']), default='creator-only',
+              help='The initial ACL applied to the scope. Must be either "creator-only"'
```
ericwang-db: secret scope

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+
+def _scopes_to_table(scopes_json):
+    ret = []
+    for s in scopes_json.get('scopes', []):
+        ret.append((truncate_string(s['name']), s['backend_type']))
+    return ret
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Lists all the scopes')
```
ericwang-db: secret scopes

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+    """
+    scopes_json = SecretApi(api_client).list_scopes()
+    if OutputClickType.is_json(output):
+        click.echo(pretty_format(scopes_json))
+    else:
+        click.echo(tabulate(_scopes_to_table(scopes_json), headers=SCOPE_HEADER))
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Deletes a scope')
```
ericwang-db: secret scope

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,290 @@
...
+from databricks_cli.version import print_version_callback, version
+
+
+SCOPE_HEADER = ('Scope', 'Backend')
+SECRET_HEADER = ('Key name', 'Last updated')
+ACL_HEADER = ('Principal', 'Permission')
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help="Creates a scope")
```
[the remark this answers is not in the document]
ericwang-db: For every short_help, can we add punctuation (just a '.') at the end? To be consistent with `databricks workspace`'s help info:
```
% databricks workspace                                                                     [04/11/18 4:31:07 PM]  ✭
Usage: databricks workspace [OPTIONS] COMMAND [ARGS]...

  Utility to interact with the Databricks workspace. Workspace paths must be
  absolute and be prefixed with `/`.

Options:
  -v, --version   0.6.0
  --profile TEXT  CLI connection profile to use. The default profile is
                  "DEFAULT".
  -h, --help      Show this message and exit.

Commands:
  delete      Deletes objects from the Databricks workspace. rm and delete are
              synonyms.
  export      Exports a file from the Databricks workspace.
  export_dir  Recursively exports a directory from the Databricks workspace.
  import      Imports a file from local to the Databricks workspace.
  import_dir  Recursively imports a directory to the Databricks workspace.
  list        List objects in the Databricks Workspace. ls and list are
              synonyms.
  ls          List objects in the Databricks Workspace. ls and list are
              synonyms.
  mkdirs      Make directories in the Databricks Workspace.
  rm          Deletes objects from the Databricks workspace. rm and delete are
              synonyms.
```

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+                       " [--string-value, --bytes-value]")
+
+    if string_value:
+        return translate_value(value, no_strip), None
+    if bytes_value:
+        return None, base64.encodestring(translate_value(value, no_strip))
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Writes a secret')
```
ericwang-db: Writes a secret to a scope.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+    """
+    secrets_json = SecretApi(api_client).list_secrets(scope)
+    if OutputClickType.is_json(output):
+        click.echo(pretty_format(secrets_json))
+    else:
+        click.echo(tabulate(_secrets_to_table(secrets_json), headers=SECRET_HEADER))
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Writes a ACL for a principal')
```
ericwang-db: Writes an access control rule for a principal applied to a given scope.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+def write_acl(api_client, scope, principal, permission):
+    """
+    Creates or overwrites the ACL associated with the given principal (user or group) on the
+    specified scope.
+    """
+    SecretApi(api_client).write_acl(scope, principal, permission)
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Deletes a ACL for a principal')
```
ericwang-db: Deletes an access control rule for a principal.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+    """
+    acls_json = SecretApi(api_client).list_acls(scope)
+    if OutputClickType.is_json(output):
+        click.echo(pretty_format(acls_json))
+    else:
+        click.echo(tabulate(_acls_to_table(acls_json), headers=ACL_HEADER))
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Gets details of ACL')
```
ericwang-db: Gets the details for an access control rule.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+
+def _acls_to_table(acls_json):
+    ret = []
+    for s in acls_json.get('items', []):
+        ret.append((s['principal'], s['permission'].upper()))
+    return ret
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Lists ACLs in a scope')
```
ericwang-db: Lists all access control rules for a given scope.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+
+@click.group(context_settings=CONTEXT_SETTINGS,
+             short_help='Utility to interact with Databricks secret manager.')
+@click.option('--version', '-v', is_flag=True, callback=print_version_callback,
+              expose_value=False, is_eager=True, help=version)
+@profile_option
+@eat_exceptions
+def secrets_group():
+    """
+    Utility to interact with Databricks secret manager. For now, secret manager is only
```
ericwang-db: For now -> Currently

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+               short_help='Gets details of ACL')
+@click.option('--scope', required=True)
+@click.option('--principal', required=True)
+@click.option('--output', default=None, help=OutputClickType.help, type=OutputClickType())
+@profile_option
+@eat_exceptions
+@provide_api_client
+def get_acl(api_client, scope, principal, output):
+    """
+    Describes the details about the given ACL.
```
ericwang-db: Describe the details about the given ACL for the principal and secret scope.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+
+    You should specify exactly one flag to indicate the value is "string-value" or "bytes-value".
+
+    If VALUE is an empty string or not provided, an editor will be opened for you to enter your
+    secret value.
+    If VALUE starts with '@', the rest of the string will be seen as a path to a file, the content
+    of file will be read as content.
+    Otherwise, the VALUE itself will be seen as secret value.
+
+    The specification for input can be found at:
```
ericwang-db: I think we can remove this line, because you've already explained the input requirements above. Also, we don't necessarily want to point them to Azure only documentation, since we'll have to change this when we support AWS :).

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+@click.option('--initial-manage-acl',
+              type=click.Choice(['creator-only', 'all-users']), default='creator-only',
+              help='The initial ACL applied to the scope. Must be either "creator-only"'
+                   ' or "all-users". Defaults to "creator-only".')
+@profile_option
+@eat_exceptions
+@provide_api_client
+def create_scope(api_client, scope, initial_manage_acl):
+    """
+    Creates a new scope.
```
ericwang-db: Create a new secret scope.

^ Add some more detail related to what an initial manage ACL means here.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+
+    The specification for the scope name and initial_manage_acl can be found at:
```
ericwang-db: same comment as below, don't link to documentation. Just add a bit more above.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Deletes a secret')
+@click.option('--scope', required=True)
+@click.option('--key', required=True)
+@profile_option
+@eat_exceptions
+@provide_api_client
+def delete_secret(api_client, scope, key):
+    """
+    Deletes the secret stored in this secret scope. You must have WRITE or MANAGE permission on
```
ericwang-db: A note of consistency, you don't mention required permission levels for other commands. We should be consistent, but including or not including this is OK.

tests/secrets/test_cli.py
```diff
@@ -0,0 +1,162 @@
+# Databricks CLI
```
ericwang-db: A couple endpoints aren't tested in this file, namely:
deleteAcl
writeAcl
deleteSecret
deleteScope

Can we add some simple tests for these?

databricks_cli/utils.py
```diff
@@ -75,6 +78,43 @@ def json_cli_base(json_file, json, api):
...
+    Generate content based on format of value. If value is None or empty string, open a temporary
+    file for user to input the content. If value starts with a '@', treat the rest of string as
+    a path and read content from file. Otherwise, use the value itself as content.
+    Strip the trailing '\n' unless a no_strip flag is set to True.
+    """
+    prompt = "# Delete this line and paste your secret value." \
+             " Any trailing new lines will be stripped (unless --no-strip specified)."
+    if value is None or len(value) == 0:
+        editor = os.environ.get('EDITOR', 'vim')
+        with tempfile.NamedTemporaryFile(suffix=".tmp") as tf:
```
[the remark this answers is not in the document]
kunmin-db: I believe the `tempfile` module will destroy the file as soon as the file is closed. https://docs.python.org/2/library/tempfile.html

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+        return None, base64.encodestring(translate_value(value, no_strip))
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Writes a secret to a scope.')
+@click.option('--scope', required=True)
+@click.option('--key', required=True)
+@click.option('--string-value', is_flag=True, default=False,
+              help='A flag indicates the value will be stored in UTF-8 (MB4) form')
+@click.option('--bytes-value', is_flag=True, default=False,
```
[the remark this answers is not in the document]
kunmin-db: Yes, it's base64 encoded.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,291 @@
...
+@profile_option
+@eat_exceptions
+@provide_api_client
+def write_secret(api_client, scope, key, string_value, bytes_value, no_strip, value):
+    """
+    Writes a secret to the provided scope with the given name. Overwrites if the name exists.
+
+    You should specify exactly one flag to indicate if the value is "string-value" or "bytes-value".
+
+    If VALUE is an empty string or not provided, an editor will be opened for you to enter your
```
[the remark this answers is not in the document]
kunmin-db: As mentioned below, it's not persisted. But I am not sure if the module like `tempfile` and `subprocess` will work on Windows.

databricks_cli/secrets/api.py
```diff
@@ -0,0 +1,59 @@
...
+# limitations under the License.
+
+from databricks_cli.sdk import SecretService
+
+
+class SecretApi(object):
+    def __init__(self, api_client):
+        self.client = SecretService(api_client)
+
+    def create_scope(self, scope, initial_manage_acl):
```
ericwang-db: this should be updated

databricks_cli/secrets/api.py
```diff
@@ -0,0 +1,59 @@
...
+        return self.client.write_acl(scope, principal, permission)
+
+    def delete_acl(self, scope, principal):
+        return self.client.delete_acl(scope, principal)
+
+    def list_acls(self, scope):
+        return self.client.list_acls(scope)
+
+    def get_acl(self, scope, principal):
+        return self.client.get_acl(scope, principal)
```
ericwang-db: missing newline? was this autogenerated?

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,308 @@
...
+
+def _scopes_to_table(scopes_json):
+    ret = []
+    for s in scopes_json.get('scopes', []):
+        ret.append((truncate_string(s['name']), s['backend_type']))
+    return ret
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Lists all the secret scopes.')
```
ericwang-db: remove "the"

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,308 @@
...
+@click.option('--output', default=None, help=OutputClickType.help, type=OutputClickType())
+@profile_option
+@eat_exceptions
+@provide_api_client
+def list_scopes(api_client, output):
+    """
+    Lists all secret scopes available in the workspace.
```
ericwang-db: remove "available in the workspace".

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,308 @@
...
+
+SCOPE_HEADER = ('Scope', 'Backend')
+SECRET_HEADER = ('Key name', 'Last updated')
+ACL_HEADER = ('Principal', 'Permission')
+VALUE_OPTIONS = ['string-value', 'bytes-value']
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help="Creates a secret scope.")
+@click.option('--scope', required=True)
```
ericwang-db: This applies to other options, but `@click.option` that are required should still have a helpful tip describing what the argument is supposed to be.

In this case, "The name of the secret scope".

Can you fix for all options with `required=True`?

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,308 @@
...
+def _verify_and_translate_options(string_value, binary_file):
+    """
+    Translates options into actual parameters for API call.
+    Return tuple with two values representing (string_value, bytes_value).
+    """
+    if string_value and binary_file:
+        error_and_quit("At most one of {} should be provided."
+                       .format(['string-value', 'binary-file']))
+
+    if string_value is None and binary_file is None:
```
ericwang-db: maybe make these `if` statements `elif`'s, so we know that exactly one block gets executed.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,308 @@
...
+                           ' Please follow the instruction to input secret value.')
+
+        return content.rstrip('\n'), None
+
+    if string_value is not None:
+        return string_value, None
+
+    if binary_file is not None:
+        with open(binary_file, 'rb') as f:
+            binary_content = f.read().rstrip('\n')
```
ericwang-db: why are we striping the last newline of a binary file? Shouldn't we pass the binary entirely as it is supplied?

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,308 @@
...
+    If none of "string-value" and "binary-file" specified, an editor will be opened for
+    inputting secret value. The value will be stored in UTF-8 (MB4) form.
+    """
+    string_param, bytes_param = _verify_and_translate_options(string_value, binary_file)
+    SecretApi(api_client).write_secret(scope, key, string_param, bytes_param)
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Deletes a secret.')
+@click.option('--scope', required=True)
```
ericwang-db: add description

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,308 @@
...
+    ret = []
+    for s in secrets_json.get('secrets', []):
+        ret.append((s['key'], s.get('last_updated_timestamp', 'Not Available')))
+    return ret
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Lists all the secrets in a scope.')
+@click.option('--scope', required=True)
+@click.option('--output', default=None, help=OutputClickType.help, type=OutputClickType())
```
ericwang-db: set default here to whatever we agreed to in the doc

ericwang-db: same for other `--output` options

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,308 @@
...
+    if OutputClickType.is_json(output):
+        click.echo(pretty_format(acl_json))
+    else:
+        acl_list = _acls_to_table({'items': [acl_json]})
+        click.echo(tabulate(acl_list, headers=ACL_HEADER))
+
+
+@click.group(context_settings=CONTEXT_SETTINGS,
+             short_help='Utility to interact with Databricks secret API.'
+                        ' (Available only in Azure Databricks)')
```
ericwang-db: Remove the only available in Azure Databricks, so we don't need to make any updates to the CLI when we release AWS. The cloud-specific error messaging is handled by the ApiProxy now.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,308 @@
...
+@click.option('--version', '-v', is_flag=True, callback=print_version_callback,
+              expose_value=False, is_eager=True, help=version)
+@profile_option
+@eat_exceptions
+def secrets_group():
+    """
+    Utility to interact with secret API. As of Apr 13, 2018, secrets API is only
```
ericwang-db: same here

databricks_cli/secrets/cli.py
kunmin-db: The default value is handled in `OutputClickType` and click's default is already `None` if not specified. I am removing `default=None` for now.

databricks_cli/secrets/cli.py
kunmin-db: I believe we assume the file contains base64 encoded data? Then it doesn't make sense the data contains trailing new lines.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,312 @@
...
+        return None, binary_content
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Writes a secret to a scope.')
+@click.option('--scope', required=True, type=SecretScopeClickType(), help=SecretScopeClickType.help)
+@click.option('--key', required=True, type=SecretKeyClickType(), help=SecretKeyClickType.help)
+@click.option('--string-value', default=None,
+              help='Read value from string and stored in UTF-8 (MB4) form')
+@click.option('--binary-file', default=None, type=click.Path(exists=True, readable=True),
```
yunzhangit: I think this API will be confusing to the users because `binary file` is not the same as `file with base64 encoded strings`. We should just call it `--value-from-file` or something similar

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,312 @@
...
+            error_and_quit('No changes made, write secret aborted.'
+                           ' Please follow the instruction to input secret value.')
+
+        return content.rstrip('\n'), None
+
+    elif string_value is not None:
+        return string_value, None
+
+    elif binary_file is not None:
+        with open(binary_file, 'rb') as f:
```
yunzhangit: We may want to explicitly define the expected encoding of the file, e.g. `UTF-8, unicode` etc. as part of the interface

ericwang-db: binary doesn't have an expected encoding, it is stored exactly as bytes.

databricks_cli/secrets/cli.py
ericwang-db: @yunzhangit The user doesn't need to know anything about base64, nor encode the string using base64 beforehand. "binary" should make it obvious that the data is stored as is, which is different from strings which get decoded as UTF-8.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,312 @@
...
+            binary_content = f.read().rstrip('\n')
```
mengxr: shouldn't strip newline for binary content

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,312 @@
...
+
+        if not is_base64_str(binary_content):
```
mengxr: We shouldn't expect base64 encoded files. Instead, we should base64-encode the content and send it to the endpoint.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,312 @@
...
+
+        return None, binary_content
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Writes a secret to a scope.')
+@click.option('--scope', required=True, type=SecretScopeClickType(), help=SecretScopeClickType.help)
+@click.option('--key', required=True, type=SecretKeyClickType(), help=SecretKeyClickType.help)
+@click.option('--string-value', default=None,
+              help='Read value from string and stored in UTF-8 (MB4) form')
```
mengxr: Just to confirm, do we do any encoding/transformation?

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,312 @@
...
+def write_secret(api_client, scope, key, string_value, binary_file):
+    """
+    Writes a secret to the provided scope with the given name. Overwrites if the name exists.
+
+    You should specify at most one option in "string-value" and "binary-file".
+
+    If "string-value", the argument will be stored in UTF-8 (MB4) form.
+
+    If "binary-file", the argument should be a path to file. The file should store
+    base64 encoded data. File content will be read as secret value and stored as bytes.
```
mengxr: Base64 encoding is not necessary.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,312 @@
...
+@profile_option
+@eat_exceptions
+@provide_api_client
+def create_scope(api_client, scope, initial_manage_principal):
+    """
+    Creates a new secret scope with given name.
+
+    If "initial_manage_principal" is specified, the initial ACL applied to the scope is
+    applied to the supplied principal (user or group) with MANAGE permissions.
+    The only supported principal for this option is the group "users", which
```
mengxr: This should go to the help message of `initial-manage-principal`.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,312 @@
...
+@click.option('--initial-manage-principal',
+              help='The initial principal that can manage the created secret scope.')
+@profile_option
+@eat_exceptions
+@provide_api_client
+def create_scope(api_client, scope, initial_manage_principal):
+    """
+    Creates a new secret scope with given name.
+
+    If "initial_manage_principal" is specified, the initial ACL applied to the scope is
```
mengxr: `initial-manage-principal` (users cannot see the underscore version)

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,312 @@
...
+from databricks_cli.utils import eat_exceptions, CONTEXT_SETTINGS, pretty_format, truncate_string, \
+    error_and_quit, is_base64_str
+from databricks_cli.configure.config import provide_api_client, profile_option
+from databricks_cli.version import print_version_callback, version
+
+
+SCOPE_HEADER = ('Scope', 'Backend')
+SECRET_HEADER = ('Key name', 'Last updated')
+ACL_HEADER = ('Principal', 'Permission')
+VALUE_OPTIONS = ['string-value', 'bytes-value']
```
mengxr: Is it used?

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,312 @@
...
+    Creates a new secret scope with given name.
+
+    If "initial_manage_principal" is specified, the initial ACL applied to the scope is
+    applied to the supplied principal (user or group) with MANAGE permissions.
+    The only supported principal for this option is the group "users", which
+    contains all users in the workspace. If "initial_manage_principal" is not specified,
+    the initial ACL with MANAGE permission applied to the scope is assigned to the
+    API request issuer's user identity.
+    """
+    if initial_manage_principal is not None and initial_manage_principal != "users":
```
mengxr: CLI shouldn't maintain this check. We should expect server return the right error message and we just forward it to users. This helps when we support other principals on the server side while users' CLI is not or cannot be updated.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,312 @@
...
+            ' UTF-8 (MB4) form. Exit without saving to abort writing secret.'
+
+        # underlying edit function made sure using a temporary file for editing
+        content = click.edit(prompt)
+        # return None means editor is closed without changes
+        if content is None:
+            error_and_quit('No changes made, write secret aborted.'
+                           ' Please follow the instruction to input secret value.')
+
+        return content.rstrip('\n'), None
```
mengxr: click.edit already strip the newlines

kunmin-db: I think `click.edit` does not strip the trailing newlines. See from the [doc](http://click.pocoo.org/5/api/) and [implementation](https://github.com/pallets/click/blob/master/click/_termui_impl.py#L412), looks like it just translating `\r\n` to `\n`.

mengxr: Yes, you are write. I misread the one that translates `\n`

databricks_cli/secrets/cli.py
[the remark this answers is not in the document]
kunmin-db: Yes, we change from field `initial_manage_acl` to `initial_manage_principal`.

databricks_cli/secrets/cli.py
kunmin-db: Follow @andrewmchen 's suggestion to use a `git commit` style message.

databricks_cli/secrets/cli.py
kunmin-db: I think this is an extra warning as a comparison to the `--binary-file` option. If user stores bytes value, `dbutils.preview.secret.get` may not return a string.

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,310 @@
...
+            '# Exit without saving will abort writing secret.'
+
+        # underlying edit function made sure using a temporary file for editing
+        content = click.edit('\n\n' + marker + prompt)
+        # return None means editor is closed without changes
+        if content is None:
+            error_and_quit('No changes made, write secret aborted.'
+                           ' Please follow the instruction to input secret value.')
+
+        return content.split(marker, 1)[0].rstrip('\n'), None
```
andrewmchen: Can we validate that the marker was not changed? Also, can we modify unit tests to test this splitting logic?

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,310 @@
...
+              ' workspace. If not specified, the initial ACL with MANAGE permission applied to the'
+              ' scope is assigned to the request issuer\'s user identity.')
+@profile_option
+@eat_exceptions
+@provide_api_client
+def create_scope(api_client, scope, initial_manage_principal):
+    """
+    Creates a new secret scope with given name.
+    """
+    SecretApi(api_client).create_scope(scope, initial_manage_principal)
```
andrewmchen: Why does
```
databricks secrets create-scope --scope andrew --initial-manage-principal apple
```
not get rejected by the backend?

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,310 @@
...
+        if content is None:
+            error_and_quit('No changes made, write secret aborted.'
+                           ' Please follow the instruction to input secret value.')
+
+        return content.split(marker, 1)[0].rstrip('\n'), None
+
+    elif string_value is not None:
+        return string_value, None
+
+    elif binary_file is not None:
```
andrewmchen: Can we add unit test coverage for this case?

databricks_cli/secrets/cli.py
kunmin-db: I think the backend change is not fully rolled out.

ericwang-db: Proto silently ignores invalid JSON fields too :).

databricks_cli/secrets/cli.py
ericwang-db: @mengxr It's important for users to know that if they store their secrets as strings in the backend, we assume an encoding for when it's subsequently read via `dbutils.preview.secret.get`. Depending on the user-defined encoding, we may not be able to provide them a reliable encoding, so if they want to handle encoding themselves, they must use --binary-file.

databricks_cli/secrets/cli.py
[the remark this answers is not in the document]
ericwang-db: @andrewmchen The group was never called "all_users". It was called "users", and we initially named the ACL "all_users" to make it more obvious. Now it's 1-1 with the workspace group name.

ericwang-db:
lgtm

databricks_cli/secrets/cli.py
```diff
@@ -0,0 +1,313 @@
...
+            '# Exit without saving will abort writing secret.'
+
+        # underlying edit function made sure using a temporary file for editing
+        content = click.edit('\n\n' + DASH_MARKER + prompt)
+        # return None means editor is closed without changes
+        if content is None:
+            error_and_quit('No changes made, write secret aborted.'
+                           ' Please follow the instruction to input secret value.')
+
+        elif DASH_MARKER not in content:
```
ericwang-db: I find it a bit weird that a user can't immediately delete everything in the buffer to add their secret. However, the example in http://click.pocoo.org/5/utils/#launching-editors is also weird in that accidentally editing the commented text will include it in the saved data.

I'm ok as is, since it's harder for a user to make a mistake and unexpected include unwanted data in the secret. :)

PR #109, opened 2018-04-11
Fix python 3 basic auth.

andrewmchen:
There was a bug with python 3 basic auth encoding where we put a string into the function `base64.standard_b64encode`.

### DIFF
--- a/databricks_cli/sdk/api_client.py
+++ b/databricks_cli/sdk/api_client.py
@@ -72,8 +72,9 @@ class ApiClient(object):
 
         self.url = "%s/api/%s" % (host, apiVersion)
         if user is not None and password is not None:
-            userHeaderData = "Basic " + base64.standard_b64encode(user + ":" + password)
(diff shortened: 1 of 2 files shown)


databricks_cli/sdk/api_client.py
```diff
@@ -72,8 +72,9 @@ def __init__(self, user = None, password = None, host = None, token = None,
 
         self.url = "%s/api/%s" % (host, apiVersion)
         if user is not None and password is not None:
-            userHeaderData = "Basic " + base64.standard_b64encode(user + ":" + password)
-            auth = {'Authorization': userHeaderData, 'Content-Type': 'text/json'}
+            encoded_auth = (user + ":" + password).encode()
+            user_header_data = "Basic " + base64.standard_b64encode(encoded_auth).decode()
+            auth = {'Authorization': user_header_data, 'Content-Type': 'text/json'}
```
mengxr: (not part of this PR) "Content-Type" shouldn't be part of the auth header.

PR #112, opened 2018-04-13
Fix test runner.

andrewmchen:
We actually didn't run any PR tests since this https://github.com/databricks/databricks-cli/pull/94/files went in...


mengxr:
@andrewmchen Seems there is a linter issue. Could you fix?

PR #118, opened 2018-04-30
Add insecure option to configure

andrewmchen:
We would like to use the CLI against shards without proper SSL certificates. (For development purposes.)

### DIFF
--- a/databricks_cli/configure/cli.py
+++ b/databricks_cli/configure/cli.py
@@ -36,15 +36,15 @@ PROMPT_PASSWORD = 'Password' #  NOQA
 PROMPT_TOKEN = 'Token' #  NOQA
 
 
-def _configure_cli_token(profile):
+def _configure_cli_token(profile, insecure):
     config = get_config_for_profile(profile)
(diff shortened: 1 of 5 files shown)


ericwang-db:
@andrewmchen Nice! Why did you prefer to make this a config option vs a command line option that's per command invocation (like `--profile`). That seems to be more in line with CURL behavior, and makes the lack of trust more explicit.

andrewmchen:
IMO, that doesn't make a lot of sense. Whether or not the host has SSL set up correctly is super coupled to the profile so it should be configured with the profile I think.

ericwang-db:
@andrewmchen looks good! However, could we make `insecure` not show up in ~/.databrickscfg when it is set to None? That happens for username/password when switching the default config to token-based auth.

ericwang-db:
@andrewmchen made a change to make ^ happen. https://github.com/ericwang-db/databricks-cli/commit/81abbeee520851be6f3dcef5092a5ae9063cbd8a

Could you cherry-pick and let's close this PR?

ericwang-db:
lgtm

PR #119, opened 2018-04-30
Update secrets CLI to match REST API endpoint changes

ericwang-db:


ericwang-db:
Yeah we keep the previous endpoints, users using v0.7.0 will still work. The DB CLI upgrade is breaking for users who use `secrets write` and `secrets write-acl`. I'm investigating if I can configure an alias for `write` that is not displayed in documentation.

PR #123, opened 2018-05-28
Hide token from console during configuration.

andrewmchen:
We should hide the token from the console during configuration for security purposes.


smurching (approved):
LGTM

chrisst:

PR #131, opened 2018-06-29
Move dbfs cp functions from dbfs/cli to dbfs/api

alinxie:
I want to be able to use the dbfs cp functionality when I'm implementing stack deployment for dbfs resources.

It looks like there isn't tests for the functions yet, so that may be something to implement in the future.

### DIFF
--- a/databricks_cli/dbfs/api.py
+++ b/databricks_cli/dbfs/api.py
@@ -128,7 +128,7 @@ class DbfsApi(object):
     def move(self, dbfs_src, dbfs_dst):
         self.client.move(dbfs_src.absolute_path, dbfs_dst.absolute_path)
 
-    def copy_to_dbfs_non_recursive(self, src, dbfs_path_dst, overwrite):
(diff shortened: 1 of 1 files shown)


databricks_cli/dbfs/api.py
```diff
@@ -127,3 +127,100 @@ def mkdirs(self, dbfs_path):
...
+                raise e
+        self.put_file(src, dbfs_path_dst, overwrite)
+
+    def copy_from_dbfs_non_recursive(self, dbfs_path_src, dst, overwrite):
+        # Munge dst path in case dst is a dir
+        if os.path.isdir(dst):
+            dst = os.path.join(dst, dbfs_path_src.basename)
+        self.get_file(dbfs_path_src, dst, overwrite)
+
+    def copy_to_dbfs_recursive(self, src, dbfs_path_dst, overwrite):
```
andrewmchen: nit: can we make these private? `copy_to_dbfs_recursive` -> `_copy_to_dbfs_recursive`

alinxie: yes

PR #132, opened 2018-06-29
Command Information in user-agent header API requests

alinxie:
Appended the command and subcommand along with a randomly generated UUID into the user-agent header. The command and subcommand are extracted from the click context command path while the uuid is generated from uuid.uuid1, which ensures uniqueness between users and commands through temporal and machine-specific features.

Ex. For a the command `databricks jobs get`:
Headers: `{"user-agent": "databricks-cli-v0.xx-jobs-get-1234-5678"...}`

### DIFF
--- a/databricks_cli/configure/config.py
+++ b/databricks_cli/configure/config.py
@@ -21,6 +21,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import uuid
 import click
 import six
 
@@ -35,16 +36,18 @@ def provide_api_client(function):
     Injects the api_client keyword argument to the wrapped function.
     All callbacks wrapped by provide_api_client expect the argument ``profile`` to be passed in.
     """
-    @click.pass_context
+    # @click.pass_context
     @six.wraps(function)
-    def decorator(ctx, *args, **kwargs):
+    def decorator(*args, **kwargs):
+        ctx = click.get_current_context()
         command_name = "-".join(ctx.command_path.split(" ")[1:])
-        command_name += "_randomId"
+        command_uuid = str(uuid.uuid1())
+        default_headers = {"cli-command-name": command_name, "cli-command-uuid": command_uuid}
         profile = get_profile_from_context()
         config = get_config_for_profile(profile)
         if not config.is_valid:
             raise InvalidConfigurationError(profile)
-        kwargs['api_client'] = _get_api_client(config, command_name)
+        kwargs['api_client'] = _get_api_client(config, default_headers)
 
         return function(*args, **kwargs)
     decorator.__doc__ = function.__doc__
@@ -67,8 +70,8 @@ def profile_option(f):
                         help='CLI connection profile to use. The default profile is "DEFAULT".')(f)
 
 
-def _get_api_client(config, command_name=""):
-    default_headers = {"command_name": command_name}
+def _get_api_client(config, default_headers=None):
+    default_headers = {} if default_headers is None else default_headers
     verify = config.insecure is None
     if config.is_valid_with_token:
         return ApiClient(host=config.host, token=config.token, verify=verify,
--- a/databricks_cli/configure/config.py
+++ b/databricks_cli/configure/config.py
@@ -40,13 +40,12 @@ def provide_api_client(function):
     def decorator(*args, **kwargs):
         ctx = click.get_current_context()
         command_name = "-".join(ctx.command_path.split(" ")[1:])
-        command_uuid = str(uuid.uuid1())
-        default_headers = {'cli-command-name': command_name, 'cli-command-uuid': command_uuid}
+        command_name += "-" + str(uuid.uuid1())
         profile = get_profile_from_context()
         config = get_config_for_profile(profile)
         if not config.is_valid:
             raise InvalidConfigurationError(profile)
-        kwargs['api_client'] = _get_api_client(config, default_headers)
+        kwargs['api_client'] = _get_api_client(config, command_name)
 
         return function(*args, **kwargs)
     decorator.__doc__ = function.__doc__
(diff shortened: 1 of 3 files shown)


tests/configure/test_config.py
```diff
@@ -78,3 +81,38 @@ def test_command(): # noqa
...
+
+    @click.command()
+    @config.profile_option
+    @config.provide_api_client
+    def test_command(api_client): # noqa
+        click.echo(json.dumps(api_client.default_headers))
+
+    with mock.patch("databricks_cli.configure.provider.DatabricksConfig") as ConfigMock:
+        with mock.patch("uuid.uuid1") as uuid_mock:
+            ConfigMock.return_value = DatabricksConfig.from_token(TEST_HOST, TEST_TOKEN)
```
xuanwang14: Why is this CamelCase instead of under_score?

databricks_cli/configure/config.py
```diff
@@ -37,11 +38,15 @@ def provide_api_client(function):
     """
     @six.wraps(function)
     def decorator(*args, **kwargs):
+        ctx = click.get_current_context()
+        command_name = "-".join(ctx.command_path.split(" ")[1:])
```
xuanwang14: some commands may have "-" in it. What would be the standard way of doing it? How about use `"/".join` instead?

databricks_cli/configure/config.py
```diff
@@ -37,11 +38,15 @@ def provide_api_client(function):
...
+        command_uuid = str(uuid.uuid1())
+        default_headers = {'cli-command-name': command_name, 'cli-command-uuid': command_uuid}
```
xuanwang14: Could you explain why this head is needed here? Do you know where in the logs we can extract such information?

tests/configure/test_config.py
```diff
@@ -78,3 +81,38 @@ def test_command(): # noqa
...
+
+def test_command_headers():
+    @click.group()
+    @config.profile_option
+    def test_group():
+        pass
+
+    @click.group()
+    @config.profile_option
+    def test_group_2():
```
xuanwang14: Instead of `test_group` and `test_group_2` could you use some better names?

tests/configure/test_config.py
xuanwang14: Why do you need to mock this class here while other tests don't? Could you add some comments here?

tests/configure/test_config.py
```diff
@@ -78,3 +81,38 @@ def test_command(): # noqa
...
+    @config.profile_option
+    @config.provide_api_client
+    def test_command(api_client): # noqa
+        click.echo(json.dumps(api_client.default_headers))
+
+    with mock.patch("databricks_cli.configure.provider.DatabricksConfig") as ConfigMock:
+        with mock.patch("uuid.uuid1") as uuid_mock:
+            ConfigMock.return_value = DatabricksConfig.from_token(TEST_HOST, TEST_TOKEN)
+            uuid_mock.return_value = '1234'
+            test_group_2.add_command(test_command, 'test')
```
xuanwang14: I don't understand this. Should "test_command" be consistent with "test"?

databricks_cli/configure/config.py
xuanwang14: Actually, would it be cleaner if you just add the command and sub command separately? Since we already have a json, it would be good to keep the structured data.

databricks_cli/configure/config.py
```diff
@@ -37,11 +38,15 @@ def provide_api_client(function):
     """
     @six.wraps(function)
     def decorator(*args, **kwargs):
+        ctx = click.get_current_context()
+        command_name = "-".join(ctx.command_path.split(" ")[1:])
+        command_uuid = str(uuid.uuid1())
```
xuanwang14: Should UUID be generated per CLI commands? Here it seems to be generated for every instance of API client. Also what if the stack CLI command calls other CLI methods (will there be any duplicated records)?

tests/configure/test_config.py
alinxie: sure

tests/configure/test_config.py
alinxie: I think I was following the syntax of the imported module but I'll change it

alinxie: When I mock it, any instantiation of the DatabricksConfig will be valid in order to properly return the API client.

tests/configure/test_config.py
alinxie: yeah I suppose it should, I'll change it.

databricks_cli/configure/config.py
alinxie: I'm not sure, I was going to ask you about that.

alinxie: I was thinking that a new metric would need to be defined

databricks_cli/configure/config.py
alinxie: Hm command and subcommand stored separately may work, but I'm worried that there may be even more nested subcommands in the future (ie. a subcommand within a subcommand) and that will be inconvenient to handle

alinxie: Actually I think it may be appropriate just to include the spaces

databricks_cli/configure/config.py
```diff
@@ -37,11 +38,14 @@ def provide_api_client(function):
     """
     @six.wraps(function)
     def decorator(*args, **kwargs):
+        ctx = click.get_current_context()
+        command_name = "-".join(ctx.command_path.split(" ")[1:])
+        command_name += "-" + str(uuid.uuid1())
```
xuanwang14: Is this generated per CLI method call?

databricks_cli/configure/config.py
xuanwang14: This function is a little confusing. Could you add some examples in comments?

tests/configure/test_config.py
```diff
@@ -78,3 +81,38 @@ def test_command(): # noqa
...
+    def test_command(api_client, x): # noqa
+        click.echo(json.dumps(api_client.default_headers))
+
+    with mock.patch("databricks_cli.configure.provider.DatabricksConfig") as config_mock:
+        with mock.patch("uuid.uuid1") as uuid_mock:
+            config_mock.return_value = DatabricksConfig.from_token(TEST_HOST, TEST_TOKEN)
+            uuid_mock.return_value = '1234'
+            inner_test_group.add_command(test_command, 'test-command')
+            outer_test_group.add_command(inner_test_group, 'my')
+            result = CliRunner().invoke(outer_test_group, ['my', 'test-command', '--x', '12'])
```
xuanwang14: "my test-command" => "command subcommand"? for better readability?

xuanwang14:
Please also update the title and descriptions.

databricks_cli/configure/config.py
alinxie: yes

databricks_cli/configure/config.py
nfx: @xuanwang14 @alinxie currently it's hard to analyse the usage of subcommands.

PR #135, opened 2018-07-04
Added get_status_json to dbfs and workspace api's

alinxie:
Added api functions to get the status in a json/dict format instead of the Workspace FileInfo and Dbfs FileInfo classes. 

With this change, new CLI API's will be able to directly get a JSON response from the Jobs CLI API instead of dealing with serializing FileInfo classes to JSON or having to call the Jobs Service in the SDK.

### DIFF
--- a/databricks_cli/dbfs/api.py
+++ b/databricks_cli/dbfs/api.py
@@ -89,8 +89,7 @@ class DbfsApi(object):
         return True
 
     def get_status(self, dbfs_path):
-        json = self.client.get_status(dbfs_path.absolute_path)
-        return FileInfo.from_json(json)
+        return FileInfo.from_json(self.client.get_status(dbfs_path.absolute_path))
 
     def get_status_json(self, dbfs_path):
         return self.client.get_status(dbfs_path.absolute_path)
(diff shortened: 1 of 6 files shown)


databricks_cli/dbfs/api.py
```diff
@@ -92,6 +92,9 @@ def get_status(self, dbfs_path):
         json = self.client.get_status(dbfs_path.absolute_path)
         return FileInfo.from_json(json)
 
+    def get_status_json(self, dbfs_path):
```
xuanwang14: It is not clean to expose both `get_status_json` and `get_status` in the api. Maybe `get_status` is sufficient?

alinxie: Maybe I'll use `_get_status_json` so it's not exposed in the API

alinxie: I want to get the actual JSON instead of the FileInfo object

alinxie: Actually, I'd like to keep it as `get_status_json` since I want it to be exposed in the API for purposes of stack deployment. I'll update the description of this for clarity.

alinxie: Hm yeah that seems cleaner. I added changes to the WorkspaceFileInfo class with a `json` property but haven't deleted the `get_status_json` function yet to keep the thread up. let me know how it looks!

alinxie: ok actually it's probably better to call get_status from the WorkspaceService `client` of `WorkspaceApi`.

PR #138, opened 2018-07-06
Add get_jobs_by_name function in Jobs API

alinxie:
The get_jobs_by_name function takes in a job name and returns a list of the jobs that have the same name. If no jobs have the same name as the inputted job, an empty list will be returned.

### DIFF
--- a/databricks_cli/jobs/api.py
+++ b/databricks_cli/jobs/api.py
@@ -40,7 +40,7 @@ class JobsApi(object):
         return self.client.get_job(job_id)
 
     def get_jobs_by_name(self, full_job_name):
-        jobs = self.client.list_jobs()['jobs']
+        jobs = self.list_jobs()['jobs']
         result = list(filter(lambda job: job['settings']['name'] == full_job_name, jobs))
         return result
 
--- a/databricks_cli/jobs/api.py
+++ b/databricks_cli/jobs/api.py
@@ -39,6 +39,11 @@ class JobsApi(object):
     def get_job(self, job_id):
         return self.client.get_job(job_id)
 
+    def get_jobs_by_name(self, full_job_name):
+        jobs = self.client.list_jobs()['jobs']
+        result = list(filter(lambda job: job['settings']['name'] == full_job_name, jobs))
+        return result
+
     def reset_job(self, json):
         return self.client.client.perform_query('POST', '/jobs/reset', data=json)
 
--- a/databricks_cli/jobs/api.py
+++ b/databricks_cli/jobs/api.py
@@ -39,14 +39,14 @@ class JobsApi(object):
     def get_job(self, job_id):
         return self.client.get_job(job_id)
 
-    def get_jobs_by_name(self, name):
-        jobs = self.list_jobs()['jobs']
-        result = list(filter(lambda job: job['settings']['name'] == name, jobs))
-        return result
-
     def reset_job(self, json):
         return self.client.client.perform_query('POST', '/jobs/reset', data=json)
 
     def run_now(self, job_id, jar_params, notebook_params, python_params, spark_submit_params):
         return self.client.run_now(job_id, jar_params, notebook_params, python_params,
                                    spark_submit_params)
+
+    def _list_jobs_by_name(self, name):
+        jobs = self.list_jobs()['jobs']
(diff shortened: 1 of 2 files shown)


databricks_cli/jobs/api.py
```diff
@@ -39,6 +39,11 @@ def delete_job(self, job_id):
     def get_job(self, job_id):
         return self.client.get_job(job_id)
 
+    def get_jobs_by_name(self, full_job_name):
+        jobs = self.client.list_jobs()['jobs']
+        result = list(filter(lambda job: job['settings']['name'] == full_job_name, jobs))
```
xuanwang14: What would happen if the jobName does not exist? or if there are more than one jobs with the same names?
Do we need unit tests for those behaviors?

alinxie: If jobName doesn't exist, it should be an empty list. I'll unit test it!

databricks_cli/jobs/api.py
```diff
@@ -39,6 +39,11 @@ def delete_job(self, job_id):
     def get_job(self, job_id):
         return self.client.get_job(job_id)
 
+    def get_jobs_by_name(self, full_job_name):
```
xuanwang14: `job_name` or just `name` so that it is consistent with REST API?

tests/jobs/test_api.py
```diff
@@ -0,0 +1,66 @@
...
+        jobs_api_mock = JobsApi(None)
+        yield jobs_api_mock
+
+
+@provide_conf
+def test_get_jobs_by_name(jobs_api):
+    job_name = 'test job'
+    job_other_name = 'test job alt'
+    test_job = {'settings': {'name': job_name}}
+    test_job_alt = {'settings': {'name': job_other_name}}
```
xuanwang14: It would be good to keep the naming consistent, e.g. `test_other_job` or ` job_name_alt`.

xuanwang14: Or just `job_name_1` `job_name_2`, `test_job_1`, `test_job_2`. The goal is that reviewer can understand the unit tests without going back and forth checking the references.

tests/jobs/test_api.py
```diff
@@ -0,0 +1,66 @@
...
+    res = jobs_api.get_jobs_by_name(test_job_name)
+    assert len(res) == 0
+
+    jobs_api.list_jobs.return_value = {'jobs': [test_job_alt]}
+    res = jobs_api.get_jobs_by_name(test_job_name)
+    assert len(res) == 0
+
+    jobs_api.list_jobs.return_value = {'jobs': [test_job, test_job_alt]}
+    res = jobs_api.get_jobs_by_name(test_job_name)
+    assert len(res) == 1
```
mengxr: You should assert on the actual job names result in this case. Same below.

tests/jobs/test_api.py
```diff
@@ -0,0 +1,66 @@
...
+
+# pylint:disable=redefined-outer-name
+
+import mock
+import pytest
+
+from databricks_cli.jobs.api import JobsApi
+from tests.utils import provide_conf
+
+CREATE_RETURN = {'job_id': 5}
```
mengxr: not used

tests/jobs/test_api.py
```diff
@@ -0,0 +1,66 @@
...
+CREATE_JSON = '{"name": "test_job"}'
```
mengxr: not used

databricks_cli/jobs/api.py
```diff
@@ -39,6 +39,11 @@ def delete_job(self, job_id):
     def get_job(self, job_id):
         return self.client.get_job(job_id)
 
+    def get_jobs_by_name(self, name):
```
mengxr: Could you rename it to "_list_jobs_by_name" and move it to the bottom? This doesn't match any API endpoint.

tests/jobs/test_api.py
alinxie: will do.

databricks_cli/jobs/api.py
alinxie: Since I'm planning to call this function from the stack api, isn't it bad practice to call functions from other classes that have a `_` in front? please advise.

PR #143, opened 2018-07-18
idempotency_token [ES-6905]

dougbateman:
Adds an option to specify an idempotency_token in the ClusterService.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,6 @@
 *.swo
 .cache
 .*project
+.idea/*
+*.iml
 env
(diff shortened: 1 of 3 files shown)


ileshko:
Biggest ask - if the `idempotency_token` is not passed in, please do not generate one

dougbateman:
Oh.  How does this fix the bug then that Jacob is experiencing where clusters were created twice (the CLI only made the one rest call).

dougbateman:
In this case, I think we should cancel this PR and just go back to passing in the idempotency_token in the JSON passed to the CLI.  But I don't think it will fix the bug we're encountering with the REST API creating a cluster twice.

Issue #110, opened 2018-04-11
Add global --debug flag


andrewmchen:
It'd be useful to add a global flag called `--debug` which prints out the stack trace when an error is raised.

dc-coder:
would be really useful :)

PR #145, opened 2018-07-18
Add global --debug flag

alinxie:
Added `debug_mode` property in `ContextObject` to be accessed by the `error_and_quit` function in `utils.py`

For Issue: https://github.com/databricks/databricks-cli/issues/110

### DIFF
--- a/databricks_cli/configure/config.py
+++ b/databricks_cli/configure/config.py
@@ -58,6 +58,14 @@ def get_profile_from_context():
     return context_object.get_profile()
 
 
+def debug_option(f):
+    def callback(ctx, param, value):
+        context_object = ctx.ensure_object(ContextObject)
+        context_object.set_debug(value)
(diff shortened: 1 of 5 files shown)


andrewmchen:
Can we add a quick test for this?

databricks_cli/configure/config.py
```diff
@@ -58,6 +58,14 @@ def get_profile_from_context():
     return context_object.get_profile()
 
 
+def debug_option(f):
+    def callback(ctx, param, value): #  NOQA
+        context_object = ctx.ensure_object(ContextObject)
+        context_object.set_debug(value)
+    return click.option('--debug', '-d', is_flag=True, callback=callback,
```
xuanwang14: Probably we do not need "-d" here? It is okey to require the full option name for debug

alinxie:
Added test and removed shorthand debug flag (-d)

andrewmchen:
Thanks for the feature!

PR #147, opened 2018-07-25
Stack download workspace

alinxie:
PR containing stack download features that download Workspace Directories and Notebooks based on a JSON configuration. Similar structure to `stack deploy`, except there isn't any error thrown when a resource's service isn't compatible with the stack download feature. This is because not all resources (ex. jobs) are defined to be compatible with download. Also the stack status is not used in this for simplicity, and there is not a clear use case for it to be used.


databricks_cli/stack/cli.py
```diff
@@ -50,6 +50,25 @@ def deploy(api_client, config_path, **kwargs):
     click.echo('#' * 80)
 
 
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Download stack given a JSON configuration of the stack')
+@click.argument('config_path', type=click.Path(exists=True), required=True)
+@click.option('--overwrite', '-o', is_flag=True, help='Include to overwrite existing'
+                                                      ' notebooks in the workspace.')
```
andrewmchen: notebooks in the local filesystem

databricks_cli/stack/api.py
```diff
@@ -321,6 +382,39 @@ def _deploy_workspace(self, resource_properties, physical_id, overwrite):
...
+                raise StackError("Workspace Notebook language and format cannot be inferred."
+                                 "Please check file extension of notebook 'source_path'.")
+            (_, fmt) = language_fmt
+            local_dir = os.path.dirname(os.path.abspath(local_path))
+            if not os.path.exists(local_dir):
+                os.makedirs(local_dir)
+            self.workspace_client.export_workspace(workspace_path, local_path, fmt, overwrite)
+        elif object_type == 'DIRECTORY':
+            if not os.path.exists(local_path):
+                os.makedirs(local_path)
```
andrewmchen: Is there test coverage for this?

databricks_cli/stack/api.py
```diff
@@ -321,6 +382,39 @@ def _deploy_workspace(self, resource_properties, physical_id, overwrite):
...
+        :param overwrite: Whether or not to overwrite the contents of workspace notebooks.
+        """
+        # Required fields. TODO(alinxie) put in _validate_config
+        local_path = resource_properties.get('source_path')
+        workspace_path = resource_properties.get('path')
+        object_type = resource_properties.get('object_type')
+        click.echo('Downloading {} from Databricks path {} to {}'.format(object_type,
+                                                                         workspace_path,
+                                                                         local_path))
+        if object_type == 'NOTEBOOK':
```
andrewmchen: Put this in a constant please!

databricks_cli/stack/cli.py
alinxie: 👍

databricks_cli/stack/api.py
alinxie: just found out i don't need this. But I'll add test coverage for the `'NOTEBOOK'` case.

databricks_cli/stack/api.py
alinxie: will use the constant from workspace API

PR #149, opened 2018-07-26
Stack Deploy DBFS

alinxie:
Decided that this will just deploy dbfs assets, instead of download dbfs assets.

The main logic uses the `DbfsApi.cp` function to import dbfs files into the workspace. Tests are identical to that of deploy workspace assets.

This conflicts with the deploy workspace PR #147 because instead of having separate --overwrite-dbfs and  --overwrite-notebooks flags, we're planning to have just one --overwrite flag for both dbfs and notebooks. Therefore, please notify me when the the aformentioned PR is merged.


databricks_cli/stack/api.py
```diff
@@ -308,6 +321,46 @@ def _deploy_workspace(self, resource_properties, physical_id, overwrite):
...
+        :param physical_id: dict containing physical identifier of workspace asset on Databricks.
+        Should contain the field 'path'.
+        :param overwrite: Whether or not to overwrite the contents of workspace notebooks.
+        :return: (dict, dict) of (physical_id, deploy_output). physical_id is the physical ID for
+        the stack status that contains the dbfs path of the file on Databricks.
+        deploy_output is the initial information about the asset on Databricks at deploy time
+        returned by the REST API.
+        """
+        # Required fields. TODO(alinxie) put in _validate_config
+        local_path = resource_properties.get('source_path')
```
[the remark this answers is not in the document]
alinxie: I named it `local_path` since I don't want confusion as to whether source_path is local or not.

databricks_cli/stack/cli.py
```diff
@@ -34,8 +34,10 @@
 @click.command(context_settings=CONTEXT_SETTINGS,
                short_help='Deploy stack given a JSON configuration of the stack')
 @click.argument('config_path', type=click.Path(exists=True), required=True)
-@click.option('--overwrite-notebooks', '-o', is_flag=True, default=False, show_default=True,
+@click.option('--overwrite-notebooks', is_flag=True, default=False, show_default=True,
```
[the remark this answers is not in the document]
alinxie: yeah that's what I was thinking as well. is there a use case to separate them?

databricks_cli/stack/api.py
```diff
@@ -308,6 +321,45 @@ def _deploy_workspace(self, resource_properties, physical_id, overwrite):
...
+        deploy_output is the initial information about the dbfs asset at deploy time
+        returned by the REST API.
+        """
+        # Required fields. TODO(alinxie) validate fields in _validate_config
+        local_path = resource_properties.get('source_path')
+        dbfs_path = resource_properties.get('path')
+        is_dir = resource_properties.get('is_dir')
+
+        if is_dir != os.path.isdir(local_path):
+            raise StackError("local source_path '{}' is inconsistent"
```
andrewmchen: This error message isn't super helpful. Would prefer something more instructive to the user on the lines of "this path you gave me is a directory but you said it isn't"....

databricks_cli/stack/api.py
```diff
@@ -308,6 +321,45 @@ def _deploy_workspace(self, resource_properties, physical_id, overwrite):
...
+                             " with is_dir: {}".format(local_path, is_dir))
+        if is_dir:
+            click.echo('Uploading Directory from {} to Dbfs at {}'.format(local_path, dbfs_path))
```
andrewmchen: micronit: `Uploading directory from {} to DBFS at {}`.

databricks_cli/stack/api.py
```diff
@@ -308,6 +321,45 @@ def _deploy_workspace(self, resource_properties, physical_id, overwrite):
...
+            self.dbfs_client.cp(recursive=True, overwrite=overwrite, src=local_path, dst=dbfs_path)
+        else:
+            click.echo('Uploading File from {} to Dbfs at {}'.format(local_path, dbfs_path))
```
andrewmchen: micronit: Uploading directory from {} to DBFS at {}.

databricks_cli/stack/api.py
```diff
@@ -308,6 +321,45 @@ def _deploy_workspace(self, resource_properties, physical_id, overwrite):
...
+            self.dbfs_client.cp(recursive=False, overwrite=overwrite, src=local_path, dst=dbfs_path)
+
+        if physical_id and physical_id['path'] != dbfs_path:
```
andrewmchen: we should check this in the validation logic right?

alinxie: This is here to notify the user if from the past deployment of the dbfs asset (if one exists), the dbfs path has changed.

databricks_cli/stack/api.py
alinxie: 👍

databricks_cli/stack/api.py
andrewmchen: What if the `physical_id` you get from the status doesn't actually have the `path` key. Then you'll get a `KeyError` here right?

PR #154, opened 2018-08-02
Stack CLI: Service-specific validation for resources in stack status and stack config. 

alinxie:
Added service-specific validation for resources in both the stack status and stack config. Also added  an end-to-end test that makes sure that during deployment, the functions responsible for creating the correct fields in the stack status do their job correctly for resources of all the different services.

In addition using global variables for field names of specific service's resource properties fields so that schema can by just changing one line of code.


tests/stack/test_api.py
```diff
@@ -529,3 +548,69 @@ def test_download_resource(self, stack_api):
...
+        stack_api._validate_status(new_stack_status)
+        new_stack_status = stack_api.deploy_config(test_stack, stack_status=TEST_STATUS)
+        stack_api._validate_status(new_stack_status)
+        new_stack_status = stack_api.deploy_config(test_stack, stack_status=TEST_STATUS,
+                                                   overwrite=True)
+        stack_api._validate_status(new_stack_status)
+
+    def test_end_to_end_download(self, stack_api):
+        """
+            The stack API should not go through any validation exceptions for any resource types.
```
xuanwang14: Same here. It makes more sense to test "it should throw if the input is not expected" rather than testing "no exception if the input is expected". This test is also not end-to-end because there is no guarantee that the cli will perform the correct actions.
We should verify the correct REST API endpoints are called (with good parameters) after deploy/download are invoked, if we want to do an end to end tests.

databricks_cli/stack/api.py
```diff
@@ -459,56 +472,115 @@ def _validate_config(self, stack_config):
...
+                                                          DBFS_SERVICE, resource_id)
+                self._assert_field_in_resource_properties(DBFS_RESOURCE_SOURCE_PATH,
+                                                          resource_properties, DBFS_SERVICE,
+                                                          resource_id)
+                self._assert_field_in_resource_properties(DBFS_RESOURCE_IS_DIR, resource_properties,
+                                                          DBFS_SERVICE, resource_id)
+            else:
+                raise StackError("Resource service '{}' not supported".format(resource_service))
+
+    def _assert_field_in_resource_properties(self, field, properties, service, resource_id):
```
xuanwang14: Most of the arguments are only used in the message, without which the function name and fuctionality could be really generic. How about
```
if resouce_service == JOB_SERVICE:
click.echo("Validating resource: $resource_id)
_assert_fields_in_dict([field1, field2], resource_properties)
elif ...
```

databricks_cli/stack/api.py
```diff
@@ -459,56 +472,115 @@ def _validate_config(self, stack_config):
...
+                self._assert_field_in_resource_physical_id(WORKSPACE_RESOURCE_PATH,
+                                                           resource_physical_id,
+                                                           resource_service, resource_id)
+            elif resource_service == DBFS_SERVICE:
+                self._assert_field_in_resource_physical_id(DBFS_RESOURCE_PATH, resource_physical_id,
+                                                           resource_service, resource_id)
+            else:
+                raise StackError("{} not a valid resource status service".format(resource_service))
+
+    def _assert_field_in_resource_physical_id(self, field, physical_id, service, resource_id):
```
xuanwang14: There are too many small helpers, which may be an overhead to maintain in the long run. Let's use some generic methods and make simple messages.

tests/stack/test_api.py
```diff
@@ -529,3 +572,83 @@ def test_download_resource(self, stack_api):
             api.RESOURCE_PROPERTIES: {'test': 'test'}
         }
         stack_api._download_resource(resource_badservice)
+
+    def test_deploy_valid_stack_status(self, stack_api, tmpdir):
+        """
+            The stack API should not go through any validation exceptions for any resource types
```
xuanwang14: This test is really just validate the status file under different inputs

tests/stack/test_api.py
```diff
@@ -529,3 +572,83 @@ def test_download_resource(self, stack_api):
...
+                        workspace_api.DIRECTORY:
+                    os.makedirs(resource_properties[api.WORKSPACE_RESOURCE_SOURCE_PATH])
+            elif resource_service == api.DBFS_SERVICE:
+                if resource_properties[api.DBFS_RESOURCE_IS_DIR]:
+                    os.makedirs(resource_properties[api.DBFS_RESOURCE_SOURCE_PATH])
+                else:
+                    with open(resource_properties[api.DBFS_RESOURCE_SOURCE_PATH], 'w') as f:
+                        f.write("print('test')\n")
+
+        # Run deploy command on stack and validate stack status. If there are no exceptions
```
xuanwang14: It seems now you also check the content of the status right? So it is not just validate schemas

tests/stack/test_api.py
```diff
@@ -529,3 +572,83 @@ def test_download_resource(self, stack_api):
...
+
+    def test_download(self, stack_api):
+        """
+            stack_api.download_from_config should call the correct workspace_client endpoints
+            when the a valid input is given.
+        """
+        stack_api.workspace_client.export_workspace = mock.MagicMock()
+        stack_api.workspace_client.export_workspace_dir = mock.MagicMock()
+
+        stack_api.download_from_config(TEST_STACK)
```
xuanwang14: It would be better to separate those into two tests, because it may be confusing and hard to debug if you count the total number of invocations of the REST API methods.

databricks_cli/stack/api.py
```diff
@@ -59,6 +59,20 @@
 RESOURCE_DEPLOY_TIMESTAMP = 'timestamp'
 CLI_VERSION_KEY = 'cli_version'
 
+# Job Service Properties
+JOBS_RESOURCE_NAME = 'name'
+JOBS_RESOURCE_ID = 'job_id'
```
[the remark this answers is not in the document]
alinxie: sounds good.

databricks_cli/stack/api.py
```diff
@@ -59,6 +59,20 @@
...
+
+# Workspace Service Properties
+WORKSPACE_RESOURCE_SOURCE_PATH = 'source_path'
+WORKSPACE_RESOURCE_PATH = 'path'
+WORKSPACE_RESOURCE_OBJECT_TYPE = 'object_type'
+
+# Workspace Service Properties
```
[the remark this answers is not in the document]
alinxie: 👍

tests/stack/test_api.py
alinxie: I'm going to actually just make this test count the number of times that `_download_workspace` is called. I think that will be easier to debug.

tests/stack/test_api.py
alinxie: Yeah I'll clarify that in the comments.

tests/stack/test_api.py
```diff
@@ -529,3 +578,84 @@ def test_download_resource(self, stack_api):
             api.RESOURCE_PROPERTIES: {'test': 'test'}
         }
         stack_api._download_resource(resource_badservice)
+
+    def test_deploy_valid_stack_status(self, stack_api, tmpdir):
+        """
+            Test valid creation of the stack status on a correctly setup stack
```
xuanwang14: Improve the descriptions?
"Stack status should be valid when ...."
from the current description is is unclear what to expect

tests/stack/test_api.py
```diff
@@ -529,3 +578,84 @@ def test_download_resource(self, stack_api):
...
+                    os.makedirs(resource_properties[api.WORKSPACE_RESOURCE_SOURCE_PATH])
+            elif resource_service == api.DBFS_SERVICE:
+                if resource_properties[api.DBFS_RESOURCE_IS_DIR]:
+                    os.makedirs(resource_properties[api.DBFS_RESOURCE_SOURCE_PATH])
+                else:
+                    with open(resource_properties[api.DBFS_RESOURCE_SOURCE_PATH], 'w') as f:
+                        f.write("print('test')\n")
+
+        # Run deploy command on stack and then verify stack status. Validation for the stack
+        # status is done within the deploy command, so if the function doesn't error, then
```
xuanwang14: This is not a test.

tests/stack/test_api.py
```diff
@@ -529,3 +578,84 @@ def test_download_resource(self, stack_api):
...
+        # the generated stack is valid.
+        new_stack_status_1 = stack_api.deploy_config(test_stack)
+        # Asserting some correctness criteria for the generated stack status.
+        assert new_stack_status_1.get(api.STACK_RESOURCES) == test_stack.get(api.STACK_RESOURCES)
+        for deployed_resource in new_stack_status_1.get(api.STACK_DEPLOYED):
+            # All functions to pull the deployed output were mocked to return deploy_output
+            assert deployed_resource.get(api.RESOURCE_DEPLOY_OUTPUT) == test_deploy_output
+
+        # Test that when passing in a  status that the new status generated will still be valid.
```
xuanwang14: Again, this is not testing any behaviors for the stack_api.

tests/stack/test_api.py
```diff
@@ -529,3 +578,84 @@ def test_download_resource(self, stack_api):
...
+    def test_download(self, stack_api):
+        """
+            stack_api.download_from_config should should call _download_resource on all of its
+            resources but only call _download_workspace on the workspace resources.
+        """
+        stack_api._download_workspace = mock.MagicMock()
+        stack_api._download_resource = mock.Mock(wraps=stack_api._download_resource)
+
+        stack_api.download_from_config(TEST_STACK)
+        # because there is only two workspace resources, _download
```
xuanwang14: finish this comments?

tests/stack/test_api.py
```diff
@@ -529,3 +578,84 @@ def test_download_resource(self, stack_api):
...
+        assert stack_api._download_resource.call_count == 5
```
xuanwang14: Could you explain why 5? and why 2? You can create some simpler config just for testing this method.

tests/stack/test_api.py
```diff
@@ -529,3 +578,69 @@ def test_download_resource(self, stack_api):
             api.RESOURCE_PROPERTIES: {'test': 'test'}
         }
         stack_api._download_resource(resource_badservice)
+
+    def test_deploy_valid_stack_status(self, stack_api, tmpdir):
+        """
+            The stack status generated from a correctly set up stack passed through deployment
```
xuanwang14: The test name "test_deploy_valid_stack_status" is not clear, how about just "test_deploy_config", because this is the main method tested here.
Also could you improve the docstring? It seems that "deploy_config method should generate correct status" is suffice.

PR #155, opened 2018-08-03
Stack CLI: Stack API interface change and little changes to comments

alinxie:
The meat of this PR is a major change in the `StackApi` (`databricks_cli/stack/api`) interface. What used to be `StackApi.deploy_config` and `StackApi.download_from_config` now are `StackApi.deploy` and `StackApi.download` respectively. Now all the reading of configuration files and stack statuses is done in the CLI File (`databricks_cli/stack/cli`). 

Also fixed up comments:
- Switched `'` comments to `"` comments so that there can be more consistency with JSON standards.
- Added some needed spaces in comments.
- Completed docstring comment for `stack_cli.deploy`


databricks_cli/stack/api.py
```diff
@@ -124,14 +94,17 @@ def deploy_config(self, stack_config, stack_status=None, **kwargs):
         :param stack_status: Must have the fields of
```
andrewmchen: Fix this comment please!

databricks_cli/stack/cli.py
```diff
@@ -46,7 +92,20 @@ def deploy(api_client, config_path, **kwargs):
...
+    stack_status = _load_json(status_path)
+    config_dir = os.path.dirname(os.path.abspath(config_path))
+    cli_dir = os.getcwd()
+    os.chdir(config_dir)  # Switch current working directory to where json config is stored
+    new_stack_status = StackApi(api_client).deploy(stack_config, stack_status, **kwargs)
+    click.echo('#' * 80)
+    os.chdir(cli_dir)
+    click.echo("Saving stack status to {}".format(status_path))
+    _save_json(status_path, new_stack_status)
+    click.echo('Warning- The stack status file is an automatically generated file that is'
```
xuanwang14: "WARNIN" usually mean there are some bad style, which does not break the program immediately. Here it would make more sense if you put those comments in the generated  "status" file. (Or just remove this warning in this PR).

tests/stack/test_cli.py
```diff
@@ -22,13 +22,31 @@
...
 from click.testing import CliRunner
 
 import databricks_cli.stack.cli as cli
 from tests.utils import provide_conf
+from tests.stack.test_api import TEST_STATUS, TEST_STACK
+
+TEST_STACK_PATH = 'stack/stack.json'
+
+
+def _write_stack(tmpdir):
```
xuanwang14: To be more specific, `_save_test_stack_config`?

databricks_cli/stack/cli.py
alinxie: I would put it as a comment in the generated status file, but JSON doesn't allow for comments. Maybe I can put "Note:"?

tests/stack/test_cli.py
alinxie: maybe `_write_test_stack_config`? Since `save` implies that I have made changes to the stack.

PR #156, opened 2018-08-03
Do some minor refactoring on groups CLI

andrewmchen:
Main changes:
  - Enforce OneOf at the click level.
  - CLI documentation updates.
  - Simplifying internal API.


databricks_cli/groups/api.py
```diff
@@ -31,22 +31,11 @@ class GroupsApi(object):
...
-                                            group_name=member_name)
-        elif member_type == "user":
-            return self.client.add_to_group(parent_name=parent_name,
-                                            user_name=member_name)
-        else:
-            raise GroupsException("Invalid 'member_type' {}".format(
-                member_type
-            ))
+    def add_member(self, parent_name, user_name, group_name):
+        assert user_name is not None or group_name is not None
```
aarondav: Is the oneof assertion message OK? It would be great to get a good error like "Error: only one of `--user-name` or `--group-name` may be provided". And here, it may be good to state in a comment or assertion that it's a xor.

andrewmchen: This should really never happen since the caller will make sure only one of them is passed in. I can add a comment about the xor too. and make the check actually a xor.

aarondav: Checked it out, it looks like

`Error: Only one of ['user-name', 'group-name'] should be provided.`

which seems reasonable. Could we state the assumption at this point in the code as well? My main concern is if someone uses this in a Python library.

PR #157, opened 2018-08-03
Enable Python clients to set an overriding DatabricksConfigProvider

aarondav:
This PR cleans up and provides a clearer Python API for how we discover Databricks configuration (API frontend hostname and authentication information). The goal of this kind of refactoring is to enable other Python libraries (e.g., https://github.com/mlflow/mlflow) that want to rely on Databricks authentication to be able to use the Databricks CLI config resolution logic in a relatively futureproof and understandable fashion.

Two major additions:
- Adds a `DatabricksConfigProvider` abstract class, and three implementations (ProfileConfigProvider, EnvironmentConfigProvider, and DefaultConfigProvider) which users can leverage for canonical implementations of identifying Databricks config.
- Adds a `set_config_provider` method which alters the behavior of the global `get_config`.

We also strive to maintain significant backwards-compatibility. MLflow (and potentially other libraries) currently rely on `get_config_for_profile`, so we maintain this API with its exact current behavior, to be deprecated in a later release.

We do change the behavior of `databricks --profile <foo>`. This invocation, where the profile is explicitly provided, will no longer look for environment variables. The invocation where profile is not provided altogether will continue to look for environment variables prior to looking at the DEFAULT profile.

## How is this patch tested?

- [x] Unit tests added
- [x] Manual tested against a Databricks API. Tried using a profile, no profile, in both the existence and non-existence cases.

### DIFF
--- a/tests/configure/test_provider.py
+++ b/tests/configure/test_provider.py
@@ -21,9 +21,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from mock import patch
 import os
 import pytest
+
+from mock import patch
 from databricks_cli.configure.provider import DatabricksConfig, DEFAULT_SECTION, \
     update_and_persist_config, get_config_for_profile, get_config, \
     set_config_provider, ProfileConfigProvider, _get_path, DatabrickConfigProvider
@@ -185,4 +186,3 @@ def test_get_config_override_custom():
 def test_get_config_bad_override():
     with pytest.raises(Exception):
         set_config_provider("NotAConfigProvider")
-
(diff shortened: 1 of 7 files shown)


tests/configure/test_provider.py
```diff
@@ -117,30 +123,66 @@ def test_get_config_if_password_environment_set():
...
+            return DatabricksConfig.from_token("Override", "Token!")
+
+    try:
+        provider = TestConfigProvider()
+        set_config_provider(provider)
+        config = get_config()
+        assert config.host == "Override"
+        assert config.token == "Token!"
+    finally:
+        set_config_provider(None)
```
andrewmchen: Weird that pytest doesn't reset the state here.

aarondav: Btw, confirmed that this state is not reset between test runs (got rid of this statement and observed that _config_provider was still set). We do create a new homedir (~) each test, though.

Issue #166, opened 2018-08-14
docker image won't build:./lint.sh: line 8: prospector: command not found


jeffreybreen:
Dockerfile in master fails to build image with error `./lint.sh: line 8: prospector: command not found`

```
$ docker build -t databricks-cli .
Sending build context to Docker daemon    834kB
Step 1/5 : FROM python:2.7
 ---> 17c0fe4e76a5
Step 2/5 : WORKDIR /usr/src/databricks-cli
 ---> Using cache
 ---> b5504b00245b
Step 3/5 : COPY . .
 ---> Using cache
 ---> a11d261149fd
Step 4/5 : RUN pip install --upgrade pip &&     pip install -r dev-requirements.txt &&     pip list &&     ./lint.sh &&     pip install . &&     pytest tests
 ---> Running in b6cca86af096
Requirement already up-to-date: pip in /usr/local/lib/python2.7/site-packages (18.0)
Collecting tox==2.9.1 (from -r dev-requirements.txt (line 1))
  Downloading https://files.pythonhosted.org/packages/1d/4e/20c679f8c5948f7c48591fde33d442e716af66a31a88f5791850a75041eb/tox-2.9.1-py2.py3-none-any.whl (73kB)
Requirement already satisfied: virtualenv>=1.11.2; python_version != "3.2" in /usr/local/lib/python2.7/site-packages (from tox==2.9.1->-r dev-requirements.txt (line 1)) (16.0.0)
Collecting py>=1.4.17 (from tox==2.9.1->-r dev-requirements.txt (line 1))
  Downloading https://files.pythonhosted.org/packages/f3/bd/83369ff2dee18f22f27d16b78dd651e8939825af5f8b0b83c38729069962/py-1.5.4-py2.py3-none-any.whl (83kB)
Collecting six (from tox==2.9.1->-r dev-requirements.txt (line 1))
  Downloading https://files.pythonhosted.org/packages/67/4b/141a581104b1f6397bfa78ac9d43d8ad29a7ca43ea90a2d863fe3056e86a/six-1.11.0-py2.py3-none-any.whl
Collecting pluggy<1.0,>=0.3.0 (from tox==2.9.1->-r dev-requirements.txt (line 1))
  Downloading https://files.pythonhosted.org/packages/f5/f1/5a93c118663896d83f7bcbfb7f657ce1d0c0d617e6b4a443a53abcc658ca/pluggy-0.7.1-py2.py3-none-any.whl
[...]

PR #167, opened 2018-08-15
updated Dockerfile so tests would pass and image would build

jeffreybreen:
To follow up on [Issue 166](https://github.com/databricks/databricks-cli/issues/166), the current Dockerfile fails to build as the tests do not pass. 

I have made the following changes and it now builds (and runs the CLI) successfully:

* The packages listed in `tox-requirements.txt` are required for the tests to run (and pass), so the Dockerfile now installs them
  * I bumped the versions of several of these packages to resolve errors regarding `pytest-django` and `pyflakes`
  * there is still a warning that prospector requires a newer version of pylint, but all tests are now passing
* `click` and `tabulate` packages seem to be required by the main codebase, so I added them to `dev-requirements.txt` so they are installed (and stick around)
* In order to satisfy pyroma, the Apache license is now listed in the classifiers in `setup.py`

To save space in the image:

* All `pip install` commands now have the `--no-cache-dir` option
* All packages listed in `tox-requirements.txt` are now removed after the tests have been run

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,11 +4,15 @@ WORKDIR /usr/src/databricks-cli
 
 COPY . .
 
-RUN pip install --upgrade pip && \
-    pip install -r dev-requirements.txt && \
+RUN pip install --upgrade --no-cache-dir pip &&   \
+    pip install --no-cache-dir     \
+        -r dev-requirements.txt    \
+        -r tox-requirements.txt && \
     pip list && \
     ./lint.sh && \
-    pip install . && \
-    pytest tests
+    pip install --no-cache-dir . && \
+    pytest tests  && \
+    pip uninstall -y \
+        -r tox-requirements.txt
 
 ENTRYPOINT [ "databricks" ]
--- a/databricks_cli/workspace/api.py
+++ b/databricks_cli/workspace/api.py
@@ -152,7 +152,7 @@ class WorkspaceApi(object):
                     click.echo(('{} does not have a valid extension of {}. Skip this file and ' +
                                 'continue.').format(cur_src, extensions))
 
-    def export_workspace_dir(self, source_path, target_path, overwrite):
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -4,11 +4,15 @@ WORKDIR /usr/src/databricks-cli
 
 COPY . .
 
-RUN pip install --upgrade pip && \
-    pip install -r dev-requirements.txt && \
```
andrewmchen: Is a minimal fix here just to also install tox-requirements?

jeffreybreen: That is certainly the heart of the change -- and almost enough to fix the issue,

You also need to bump the versions on some of the packages in `tox-requirements.txt` or it will fail due to pytest-django and pyflakes dependencies. Also, you need to add the `tabulate` and `click` packages to `dev-requirements.txt`.

My other edits to Dockerfile were an attempt to minimize the resulting image size, but I didn't get a dramatic savings.

andrewmchen: Gotcha. I'm a bit concerned about changing the `dev-requirements.txt` and `tox-requirements.txt` to make this docker container build work.

I think a minimal fix is to intall `tox-requirements.txt` and to move up `pip install .` above the `./lint.sh`

Also it may be useful to add a `.dockerignore` file containing
```
**/__pycache__
**/*.pyc
```
so we don't run into https://stackoverflow.com/questions/44067609/getting-error-importmismatcherror-while-running-py-test

joanteixi: with @andrewmchen proposals (add tox-requirements to pip install and moving ./lint.sh) works without problem. 

thanks

hagridaaron:
Any idea when this will be merged?

andrewmchen:
@hagridaaron, as it is, this PR isn't ready to merge (merge conflicts and changes to `workspace/api.py`. Is your goal just to make the docker image build?

chrisst:

PR #169, opened 2018-08-20
Include error cause in exception raising

hithwen:
This is specially useful for bad requests calls

### DIFF
--- a/databricks_cli/sdk/api_client.py
+++ b/databricks_cli/sdk/api_client.py
@@ -106,5 +106,5 @@ class ApiClient(object):
         try:
             resp.raise_for_status()
         except requests.exceptions.HTTPError as e:
-            raise e
+            raise requests.exceptions.HTTPError(resp.content, e)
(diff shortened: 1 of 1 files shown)


databricks_cli/sdk/api_client.py
```diff
@@ -106,5 +106,5 @@ def perform_query(self, method, path, data = {}, headers = None):
         try:
             resp.raise_for_status()
         except requests.exceptions.HTTPError as e:
-            raise e
+            raise requests.exceptions.HTTPError(resp.content, e)
```
andrewmchen: How does this work? Does HTTPError handle the arguments or does IOError do it?

Also, could we just use `e.response` where ever this exception is caught again? Not sure what the purpose of throwing this exception again is anyways actually.

andrewmchen:
Looks like https://github.com/databricks/databricks-cli/pull/192 implements similar behavior.

PR #170, opened 2018-08-22
Add retry to API call to endpoint "2.0/dbfs/add-block"

k4jiang:
Context:
The API endpoint `2.0/dbfs/add-block` is called repeatedly to upload large files. In the event of a HTTP error, the entire upload would fail.

Instead of failing the entire upload, we will attempt to retry the failed block 3 times before raising an error.

### DIFF
--- a/databricks_cli/dbfs/api.py
+++ b/databricks_cli/dbfs/api.py
@@ -35,8 +35,8 @@ from databricks_cli.dbfs.dbfs_path import DbfsPath
 from databricks_cli.dbfs.exceptions import LocalFileExistsException
 
 BUFFER_SIZE_BYTES = 2**20
-DEFAULT_TRIES_INTEGER = 3
-DEFAULT_DELAY_INTEGER = 2
+DEFAULT_TRIES = 3
+DEFAULT_DELAY_SEC = 2
 
 
 class FileInfo(object):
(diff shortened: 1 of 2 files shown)


databricks_cli/dbfs/api.py
```diff
@@ -34,6 +35,8 @@
 from databricks_cli.dbfs.exceptions import LocalFileExistsException
 
 BUFFER_SIZE_BYTES = 2**20
+DEFAULT_TRIES = 3
+DEFAULT_DELAY = 2
```
andrewmchen: let's put some units on this variable name! :)

databricks_cli/dbfs/api.py
```diff
@@ -92,15 +95,37 @@ def get_status(self, dbfs_path):
         json = self.client.get_status(dbfs_path.absolute_path)
         return FileInfo.from_json(json)
 
-    def put_file(self, src_path, dbfs_path, overwrite):
+    def put_file(self, src_path, dbfs_path,
+                 overwrite, tries=DEFAULT_TRIES, delay=DEFAULT_DELAY_SEC):
```
david-weiluo-ren: isn't `delay` actually the initial_delay here?

david-weiluo-ren:
FYI, Josh mentioned in https://github.com/databricks/universe/pull/28496 that
```
 don't we have to worry about idempotency here? Is it possible that transient issues could somehow cause the same block to be added twice? Don't we need some sort of chunk ID / itempotency token in order to be able to safely retry like this, similar to how AWS has multipart piece IDs? /cc @aarondav
```
which is probably a valid concern

k4jiang:
I'm investigating this right now. I'll create a separate PR if changes are necessary.

PR #190, opened 2018-11-07
Add cluster resize menu option

justinmills:
The underlying sdk support was already there, it was only missing wiring up to the command layer. I
did not implement the autoscale option.


andrewmchen:
Thanks for the contribution!

PR #192, opened 2019-01-09
Improve error message

timvancann:
Currently the databricks CLI (and SDK) do not return any useful messages when the server returns a 400. Most often this is the result of an invalid json configuration when creating jobs, runs, clusters etc.

The only way to currently figure out what is wrong in the json is to use the REST API directly (e.g. using `curl`).

This PR aims to add the meaningful message that the server provides to the output of the CLI and SDK.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,6 +6,8 @@ To isolate dependencies from your global python installation, it is important to
 Another option would be to use docker directly, i.e. 
 ```bash
 docker run -it -v `echo $PWD`:/root python:3.6.8 bash
+docker run -it -v `echo $PWD`:/root python:2.7.12 bash
+
 ```
 
 - `pip install -e .`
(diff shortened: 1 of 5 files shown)


andrewmchen:
Thanks for the interest and the PR. Could you provide some information about how to reproduce this?
> Currently the databricks CLI (and SDK) do not return any useful messages when the server returns a 400. Most often this is the result of an invalid json configuration when creating jobs, runs, clusters etc.

For example, something like this achieves the following.

```
$ databricks clusters create --json '{"name": "apple"}'                                                                                                                                                                                                                                                                                       [7:57:45]
Error: {"error_code":"INVALID_PARAMETER_VALUE","message":"Missing required field: size"}
```

timvancann:
@andrewmchen I suppose it's related to using the SDK
```
from databricks_cli.sdk import ApiClient
from databricks_cli.clusters.api import ClusterApi

client = ApiClient(host=host, token=token)
api = ClusterApi(client)
api.create_cluster({"name": "apple"})
```
gives
```
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: <host>/api/2.0/clusters/create
```

Using the SDK often happens when scripting deployment and CI/CD pipelines. Good error messages are key here 😃 .

andrewmchen:
Got it, the CLI was never really intended to be used as a SDK but given that many people use it this way your request makes sense.

Issue #174, opened 2018-09-16
Bad error message when credentials are invalid


george-unbotify:
Used databricks cli to import workspace with bad credentials (invalid Token).
Received the following error message:

```
$ databricks workspace import_dir -o dummy/ /Shared/dummy
Error: ValueError: No JSON object could be decoded
```

tiagoh:
I have exactly the same issue with Databricks CLI version 0.8.2

easonlau02:
Also hit the issue under v0.8.2, but my credentials are absolutely valid. No idea about it.

andrewmchen:
Found a repro for this. It seems that if you include the `o` header in your host this may be a problem. For example, when my profile looks like 
```
[azure]
host = https://westus.azuredatabricks.net/?o=2291051382977284
username = admin
password = admin
```
I will get the `No JSON object could be decoded` error. 

This happens because of how we build the URL, https://github.com/databricks/databricks-cli/blob/master/databricks_cli/sdk/api_client.py#L74. When building it in this way we get a redirect to the login page.

easonlau02:
@andrewmchen Yes. When I remove the URI and keep domain only, it was fixed.

PR #193, opened 2019-01-15
Fix URL construction in the ApiClient class

andrewmchen:
Should close https://github.com/databricks/databricks-cli/issues/174.

In that issue, we construct incorrect URLs when the `host` parameter to `ApiClient` looks like `https://westus.azuredatabricks.net/?o=2291051382977284`.

### DIFF
--- a/databricks_cli/sdk/api_client.py
+++ b/databricks_cli/sdk/api_client.py
@@ -74,8 +74,8 @@ class ApiClient(object):
         self.session.mount('https://', TlsV1HttpAdapter())
 
         parsed_url = urlparse(host)
-        scheme = parsed_url[0]
-        hostname = parsed_url[1]
+        scheme = parsed_url.scheme
+        hostname = parsed_url.hostname
         self.url = "%s://%s/api/%s" % (scheme, hostname, apiVersion)
         if user is not None and password is not None:
             encoded_auth = (user + ":" + password).encode()
--- a/databricks_cli/sdk/api_client.py
+++ b/databricks_cli/sdk/api_client.py
@@ -39,6 +39,7 @@ import pprint
 from . import version
 
 from requests.adapters import HTTPAdapter
+from six.moves.urllib.parse import urlparse
 
 try:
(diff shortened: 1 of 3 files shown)


tests/configure/test_config.py
```diff
@@ -136,7 +136,7 @@ def inner_test_group():
     def test_command(api_client, x): # noqa
         click.echo(json.dumps(api_client.default_headers))
 
-    with mock.patch("databricks_cli.configure.provider.DatabricksConfig") as config_mock:
+    with mock.patch("databricks_cli.configure.config.get_config") as config_mock:
```
andrewmchen: This is necessary to fix this test.. https://travis-ci.org/databricks/databricks-cli/jobs/479983808

databricks_cli/sdk/api_client.py
```diff
@@ -72,7 +73,10 @@ def __init__(self, user=None, password=None, host=None, token=None,
         self.session = requests.Session()
         self.session.mount('https://', TlsV1HttpAdapter())
 
-        self.url = "%s/api/%s" % (host, apiVersion)
+        parsed_url = urlparse(host)
+        scheme = parsed_url[0]
```
aarondav: Could we use the urlparse named return values instead of indices? e.g., parsed_url.scheme, parsed_url.hostname?

tests/sdk/test_api_client.py
```diff
@@ -60,3 +60,7 @@ def test_content_from_server_on_error(m):
     with pytest.raises(requests.exceptions.HTTPError) as e:
         client.perform_query('GET', '/endpoint')
         assert error_message_contains in e.value.message
+
+def test_api_client_url_with_queryparams():
+    client = ApiClient(host='https://databricks.com/?o=123')
```
aarondav: Just to make it obvious, would you mind including a few other host names and their expected result (e.g., with no slash, or with a slash and no orgId)? Or with a different scheme?

PR #194, opened 2019-01-16
Initialize config provider when running inside DB notebooks

aarondav:


sueann:
More context (from slack):

PR to initialize our custom config provider w/in databricks-cli to retroactively fix the credentials issue within Databricks notebooks on first attach for all clusters.

PR was tested in this notebook:
https://dogfood.staging.cloud.databricks.com/#notebook/14262889/command/14262929

On Python 3, the following commands now succeed where they didn’t before:
```
dbutils.library.installPyPI("mlflow")
dbutils.library.install("dbfs:/tmp/aaron/databricks_cli-0.8.2-py3-none-any.whl")

import mlflow
mlflow.tracking.MlflowClient().list_experiments()
```

aarondav:
@andrewmchen and @sueann approved

PR #200, opened 2019-02-01
Add feature to permanent delete cluster.

### LINKED ISSUE #176: Missing option to permanently delete a cluster
There seems to be no option to use the cli to call this REST endpoint  

2.0/clusters/permanent-delete

https://docs.azuredatabricks.net/api/latest/clusters.html#permanent-delete

andrewmchen:
Closes #176.

### DIFF
--- a/databricks_cli/clusters/cli.py
+++ b/databricks_cli/clusters/cli.py
@@ -248,6 +248,23 @@ def spark_versions_cli(api_client):
     click.echo(pretty_format(ClusterApi(api_client).spark_versions()))
 
 
+@click.command(context_settings=CONTEXT_SETTINGS)
+@click.option('--cluster-id', required=True, type=ClusterIdClickType(),
+              help=ClusterIdClickType.help)
+@debug_option
+@profile_option
+@eat_exceptions
+@provide_api_client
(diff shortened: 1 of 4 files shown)


databricks_cli/clusters/cli.py
```diff
@@ -248,6 +248,23 @@ def spark_versions_cli(api_client):
...
+@click.command(context_settings=CONTEXT_SETTINGS)
+@click.option('--cluster-id', required=True, type=ClusterIdClickType(),
+              help=ClusterIdClickType.help)
+@debug_option
+@profile_option
+@eat_exceptions
+@provide_api_client
+def permanent_delete_cli(api_client, cluster_id):
+    """
+    Permanently deletes a Spark cluster.
```
smurching: Nanonit: I know we copied this from the [REST API docs](https://docs.databricks.com/api/latest/clusters.html#permanent-delete) - we could consider explicitly mentioning that this is irreversible (although I guess "permanent" already implies that). Also, maybe we should say "Databricks cluster" instead of "Spark cluster" to maintain parity with the other clusters commands, e.g:

"Permanently and irreversibly deletes a Databricks cluster"

Issue #209, opened 2019-02-13
databricks workspace list broken due to new options


andrewmchen:
```$ databricks workspace ls --debug
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/databricks_cli/utils.py", line 42, in decorator
    return function(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/databricks_cli/configure/config.py", line 55, in decorator
    return function(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/databricks_cli/workspace/cli.py", line 55, in ls_cli
    objects = WorkspaceApi(api_client).list_objects(workspace_path)
  File "/usr/local/lib/python3.6/site-packages/databricks_cli/workspace/api.py", line 94, in list_objects
    return [WorkspaceFileInfo.from_json(f) for f in objects]
  File "/usr/local/lib/python3.6/site-packages/databricks_cli/workspace/api.py", line 94, in <listcomp>
    return [WorkspaceFileInfo.from_json(f) for f in objects]
  File "/usr/local/lib/python3.6/site-packages/databricks_cli/workspace/api.py", line 76, in from_json
    return cls(**deserialized_json)
TypeError: __init__() got an unexpected keyword argument 'created_at'
```
It's broken now because the workspace API now returns more fields like created_at.

PR #210, opened 2019-02-13
Fix workspace list so it can handle additional fields in the response

andrewmchen:
Closes the issue #209


jkbradley:
LGTM though I haven't tested it

PR #219, opened 2019-03-20
Fix boolean query parameters to GET requests

andrewmchen:
It looks like we broke get requests with boolean parameters in https://github.com/databricks/databricks-cli/pull/213. The reason why is because `request('GET', 'https://databricks.com/endpoint', params={'active_only': True})` gets serialized as

`https://databricks.com/endpoint?active_only=True` instead of `https://databricks.com/endpoint?active_only=true`.

I've also added an assertion that we're not using lists as parameters for get requests. I've audited all our API calls and checked that we're not doing this.

### DIFF
--- a/databricks_cli/sdk/api_client.py
+++ b/databricks_cli/sdk/api_client.py
@@ -111,7 +111,8 @@ class ApiClient(object):
         with warnings.catch_warnings():
             warnings.simplefilter("ignore", exceptions.InsecureRequestWarning)
             if method == 'GET':
-                resp = self.session.request(method, self.url + path, params = data,
+                translated_data = {k: _translate_boolean_to_query_param(data[k]) for k in data}
+                resp = self.session.request(method, self.url + path, params = translated_data,
                     verify = self.verify, headers = headers)
             else:
                 resp = self.session.request(method, self.url + path, data = json.dumps(data),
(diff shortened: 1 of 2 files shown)


databricks_cli/sdk/api_client.py
```diff
@@ -111,7 +111,8 @@ def perform_query(self, method, path, data = {}, headers = None):
         with warnings.catch_warnings():
             warnings.simplefilter("ignore", exceptions.InsecureRequestWarning)
             if method == 'GET':
-                resp = self.session.request(method, self.url + path, params = data,
+                translated_data = {k: _translate_boolean_to_query_param(data[k]) for k in data}
```
thunterdb: I personally prefer to do `{k: f(v) for k,v in data.items()}` but the intent is clear and the perf difference is negligible here.

databricks_cli/sdk/api_client.py
```diff
@@ -127,3 +128,13 @@ def perform_query(self, method, path, data = {}, headers = None):
                 pass
             raise requests.exceptions.HTTPError(message, response=e.response)
         return resp.json()
+
+
+def _translate_boolean_to_query_param(value):
+    assert not isinstance(value, list), 'The value of the data dict cannot be a list'
```
aarondav: This is a weird error, referring to `data dict`. Maybe "GET parameters cannot pass lists of objects" or something?

PR #224, opened 2019-04-30
Address dependencies needed to allow Docker image to build

jwalkerwelltok:
This is the smallest patch I could come up with to allow for this to build in Docker. This addresses issue #166. Both _click_ and _tabulate_ are imported in the application directly, so those libraries were added to _dev-requirements.txt_ instead of _tox-requirements.txt_


jwalkerwelltok:
@andrewmchen, do you have a chance to review this? I ask because you were involved in discussion on #167

chrisst:

PR #225, opened 2019-05-31
Add support for dbfs to dbfs copies

danielvdende:
Hi all,

This PR enables copying files from dbfs locations to other dbfs locations by using a temp file/dir on the local filesystem as a stepping stone.

Cheers,
Daniel


andrewmchen:
Let's use #226 instead.

PR #232, opened 2019-06-11
Add a CredentialProvider reading from TaskContext in Spark Executors.

tomasatdatabricks:
This PR adds ability to pass db config via TaskContext when running in Spark.

### DIFF
--- a/databricks_cli/configure/provider.py
+++ b/databricks_cli/configure/provider.py
@@ -215,11 +215,12 @@ class SparkTaskContextConfigProvider(DatabricksConfigProvider):
         context = self._get_spark_task_context_or_none()
         if context is not None:
             host = context.getLocalProperty("spark.databricks.api.url")
-       	    token = context.getLocalProperty("spark.databricks.token")
+            token = context.getLocalProperty("spark.databricks.token")
             config = DatabricksConfig.from_token(host=host, token=token)
             if config.is_valid:
                 return config
         return None
+
 
 class EnvironmentVariableConfigProvider(DatabricksConfigProvider):
     """Loads from system environment variables."""
--- a/databricks_cli/configure/provider.py
+++ b/databricks_cli/configure/provider.py
@@ -211,20 +211,20 @@ class SparkTaskContextConfigProvider(DatabricksConfigProvider):
             return TaskContext.get()
         except ImportError:
             return None
-    
+  
     @classmethod
     def set_ignore_insecure(cls, x):
-        from pyspark import SparkContext        
+        from pyspark import SparkContext  # pylint: disable=import-error      
         new_val = "True" if x else None
         SparkContext._active_spark_context.setLocalProperty("spark.databricks.ignoreTls", new_val)
-    
+ 
     def get_config(self):
         context = self._get_spark_task_context_or_none()
         if context is not None:
             host = context.getLocalProperty("spark.databricks.api.url")
             token = context.getLocalProperty("spark.databricks.token")
-            ignoreTls = context.getLocalProperty("spark.databricks.ignoreTls")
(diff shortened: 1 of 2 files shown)


databricks_cli/configure/provider.py
```diff
@@ -185,20 +185,41 @@ def get_config(self):
...
+            ProfileConfigProvider()
+        )
 
     def get_config(self):
-        env_config = self.env_provider.get_config()
-        if env_config:
-            return env_config
+        for provider in self._providers:
+            config = provider.get_config()
+            if config is not None and config.is_valid:
```
andrewmchen: Why is the `config.is_valid` piece necessary here?

databricks_cli/configure/provider.py
```diff
@@ -185,20 +185,41 @@ def get_config(self):
...
+            SparkTaskContextConfigProvider(),
+            EnvironmentVariableConfigProvider(),
+            ProfileConfigProvider()
+        )
 
     def get_config(self):
-        env_config = self.env_provider.get_config()
-        if env_config:
-            return env_config
+        for provider in self._providers:
```
andrewmchen: Do we have a test for this logic?

databricks_cli/configure/provider.py
```diff
@@ -185,20 +185,41 @@ def get_config(self):
...
+            from pyspark import TaskContext  # pylint: disable=import-error
+            return TaskContext.get()
+        except:
+            return None
 
+    def get_config(self):
+        context = self._get_spark_task_context_or_none()
+        if context is not None:
+            host = context.getLocalProperty("spark.databricks.api.url")
+       	    token = context.getLocalProperty("spark.databricks.token")
```
andrewmchen: nit: let's fix this indent ;)

tests/configure/test_provider.py
```diff
@@ -133,6 +133,27 @@ def test_get_config_uses_default_profile():
...
+                return "url"
+            elif x == "spark.databricks.token":
+                return "token"
+            else:
+                raise Exception("should not get here.")
+    ctx_class = "databricks_cli.configure.provider.SparkTaskContextConfigProvider._get_spark_task_context_or_none"
+    with patch(ctx_class) as get_context_mock:
+        def get_context(*args, **kwargs):
+            return TaskContextMock()
+        get_context_mock.side_effect = get_context
```
andrewmchen: Does this work?

```
get_context_mock.return_value = TaskContextMock()
```

databricks_cli/configure/provider.py
tomasatdatabricks: It's not technically necessary because the providers have this test inside and return None if not valid. 
But I think it's better to have this test here so a)  we make sure this is consistent across all (possible future) providers and b) it makes it clear that the intended behavior is to go through the chain until valid credentials are found.. 

I can also remove the test from the individual providers but I did not want to make more changes than necessary.

databricks_cli/configure/provider.py
tomasatdatabricks: Oh, good catch, I did this in vim cause it was so small

tests/configure/test_provider.py
tomasatdatabricks: Yeah this looks better, thanks!

databricks_cli/configure/provider.py
tomasatdatabricks: The logic is the same as before except there is an extra provider at the start of the chain . 

I think the existing tests + the the new test for the new provider added in this PR should suffice. 
We do not test the order of the providers. I don't think that requires a dedicated test, I can add one though if you think we should have it.

databricks_cli/configure/provider.py
tomasatdatabricks: wait, what's wrong with the indent? I don't see it

PR #237, opened 2019-06-21
Add DeltaPipelines Deploy/Delete commands to the CLI

arulajmani:
- Add databricks pipelines group of commands
- Add deploy and delete 
- Add tests for deploy and delete

### DIFF
--- a/databricks_cli/click_types.py
+++ b/databricks_cli/click_types.py
@@ -84,7 +84,7 @@ class SecretPrincipalClickType(ParamType):
 
 class PipelineSpecClickType(ParamType):
     name = 'SPEC'
-    help = 'The path to the pipelines spec file'
+    help = 'The path to the pipelines deployment spec file'
 
 
 class OneOfOption(Option):
--- a/databricks_cli/click_types.py
+++ b/databricks_cli/click_types.py
@@ -82,6 +82,11 @@ class SecretPrincipalClickType(ParamType):
     help = 'The name of the principal.'
 
 
+class PipelinesSpecPathClickType(ParamType):
+    name = 'SPEC_PATH'
+    help = 'The path to the pipelines spec file'
+
+
 class OneOfOption(Option):
     def __init__(self, *args, **kwargs):
         self.one_of = kwargs.pop('one_of')
--- a/databricks_cli/click_types.py
+++ b/databricks_cli/click_types.py
@@ -82,8 +82,8 @@ class SecretPrincipalClickType(ParamType):
     help = 'The name of the principal.'
 
 
-class PipelinesSpecPathClickType(ParamType):
-    name = 'SPEC_PATH'
+class PipelineSpecClickType(ParamType):
+    name = 'SPEC'
     help = 'The path to the pipelines spec file'
 
 
--- a/databricks_cli/pipelines/api.py
+++ b/databricks_cli/pipelines/api.py
@@ -79,7 +79,7 @@ class PipelinesApi(object):
                 # exactly 1 or 3
                 if parsed_uri.path.startswith('//') or parsed_uri.netloc != '':
                     raise RuntimeError('invalid file uri scheme, '
-                                       'did you mean to use file:/ or file://')
+                                       'did you mean to use file:/ or file:///')
                 local_lib_objects.append(LibraryObject(lib_object.lib_type, parsed_uri.path))
             else:
                 external_lib_objects.append(lib_object)
--- a/databricks_cli/pipelines/api.py
+++ b/databricks_cli/pipelines/api.py
@@ -25,7 +25,7 @@
 from hashlib import sha1
 import os
 
-from six.moves.urllib.parse import urlsplit
+from six.moves import urllib
 
 from databricks_cli.sdk import DeltaPipelinesService
 from databricks_cli.dbfs.api import DbfsApi
@@ -37,6 +37,8 @@ from databricks_cli.configure.config import get_profile_from_context
 from databricks_cli.configure.provider import get_config, ProfileConfigProvider
 from databricks_cli.utils import InvalidConfigurationError
 
+BUFFER_SIZE = 1024 * 64
+
 
 class PipelinesApi(object):
     def __init__(self, api_client):
@@ -47,7 +49,7 @@ class PipelinesApi(object):
     def _partition_libraries_and_extract_local_paths(lib_objects):
         local_lib_objects, rest_lib_objects = [], []
         for lib_object in lib_objects:
-            uri_scheme = urlsplit(lib_object.path).scheme
+            uri_scheme = urllib.parse.urlsplit(lib_object.path).scheme
             if lib_object.lib_type == 'jar' and uri_scheme == '':
                 local_lib_objects.append(lib_object)
             elif lib_object.lib_type == 'jar' and uri_scheme.lower() == 'file':
@@ -59,7 +61,6 @@ class PipelinesApi(object):
     @staticmethod
     def _get_hashed_path(path):
         hash_buffer = sha1()
-        BUFFER_SIZE = 1024 * 64
         try:
             with open(path, 'rb') as f:
                 while True:
@@ -79,14 +80,15 @@ class PipelinesApi(object):
                         , local_lib_objects))
 
     def _get_files_to_upload(self, local_lib_objects, remote_lib_objects):
-        transformed_remote_lib_objects = map(lambda rlo: LibraryObject(rlo.lib_type, DbfsPath(rlo.path))
-                                             , remote_lib_objects)
+        transformed_remote_lib_objects = map(
+            lambda rlo: LibraryObject(rlo.lib_type, DbfsPath(rlo.path))
+            , remote_lib_objects)
         return list(filter(lambda lo_tuple: not self.dbfs_client.file_exists(lo_tuple[1].path)
                            , zip(local_lib_objects, transformed_remote_lib_objects)))
 
     """
-    Only required until the deploy/delete APIs requires the credentials in the body as well as the header.
-     Once the API requirement is relaxed, this function can be stripped out and 
+    Only required until the deploy/delete APIs requires the credentials in the body as well 
+    as the header.Once the API requirement is relaxed, this function can be stripped out and 
      includes for this function removed.
     """
     @staticmethod
--- a/databricks_cli/pipelines/api.py
+++ b/databricks_cli/pipelines/api.py
@@ -150,3 +150,8 @@ class LibraryObject(object):
         for lib_object in lib_objects:
             libraries.append({lib_object.lib_type: lib_object.path})
         return libraries
+
+    def __eq__(self, other):
+        if not isinstance(other, LibraryObject):
+            return NotImplemented
+        return self.path == other
--- a/databricks_cli/pipelines/api.py
+++ b/databricks_cli/pipelines/api.py
@@ -1,0 +1,147 @@
+# Databricks CLI
+# Copyright 2017 Databricks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"), except
+# that the use of services to which certain application programming
+# interfaces (each, an "API") connect requires that the user first obtain
+# a license for the use of the APIs from Databricks, Inc. ("Databricks"),
+# by creating an account at www.databricks.com and agreeing to either (a)
+# the Community Edition Terms of Service, (b) the Databricks Terms of
+# Service, or (c) another written agreement between Licensee and Databricks
+# for the use of the APIs.
+#
+# You may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from hashlib import sha1
+import os
+
+from six.moves.urllib.parse import urlsplit
+
+from databricks_cli.sdk import DeltaPipelinesService
+from databricks_cli.dbfs.api import DbfsApi
+from databricks_cli.dbfs.dbfs_path import DbfsPath
+
+
+# These imports are specific to the credentials part
+from databricks_cli.configure.config import get_profile_from_context
+from databricks_cli.configure.provider import get_config, ProfileConfigProvider
+from databricks_cli.utils import InvalidConfigurationError
+
+
+class PipelinesApi(object):
+    def __init__(self, api_client):
+        self.client = DeltaPipelinesService(api_client)
+        self.dbfs_client = DbfsApi(api_client)
+
+    @staticmethod
+    def _partition_libraries_and_extract_local_paths(lib_objects):
+        local_lib_objects, rest_lib_objects = [], []
+        for lib_object in lib_objects:
+            uri_scheme = urlsplit(lib_object.path).scheme
+            if lib_object.lib_type == 'jar' and uri_scheme == '':
+                local_lib_objects.append(lib_object)
+            elif lib_object.lib_type == 'jar' and uri_scheme.lower() == 'file':
+                local_lib_objects.append(LibraryObject(lib_object.lib_type, lib_object.path[5:]))
+            else:
+                rest_lib_objects.append(lib_object)
+        return local_lib_objects, rest_lib_objects
+
+    @staticmethod
+    def _get_hashed_path(path):
+        hash_buffer = sha1()
+        BUFFER_SIZE = 1024 * 64
+        try:
+            with open(path, 'rb') as f:
+                while True:
+                    data = f.read(BUFFER_SIZE)
+                    if not data:
+                        break
+                    hash_buffer.update(data)
+        except Exception as e:
+            raise RuntimeError('Error \'{}\' while processing {}'.format(e, path))
+
+        file_hash = hash_buffer.hexdigest()
+        path = 'dbfs:/pipelines/code/{}{}'.format(file_hash, os.path.splitext(path)[1])
+        return path
+
+    def _get_remote_mappings(self, local_lib_objects):
+        return list(map(lambda llo: LibraryObject(llo.lib_type, self._get_hashed_path(llo.path))
+                        , local_lib_objects))
+
+    def _get_files_to_upload(self, local_lib_objects, remote_lib_objects):
+        transformed_remote_lib_objects = map(lambda rlo: LibraryObject(rlo.lib_type, DbfsPath(rlo.path))
+                                             , remote_lib_objects)
+        return list(filter(lambda lo_tuple: not self.dbfs_client.file_exists(lo_tuple[1].path)
+                           , zip(local_lib_objects, transformed_remote_lib_objects)))
+
+
+    """
+    Only required until the deploy/delete APIs requires the credentials in the body as well as the header.
+     Once the API requirement is relaxed, this function can be stripped out and 
+     includes for this function removed.
+    """
+    @staticmethod
+    def _get_credentials_for_request():
+        profile = get_profile_from_context()
+        if profile:
+            config = ProfileConfigProvider.get_config(profile)
+        else:
+            config = get_config()
+        if not config or not config.is_valid:
+            raise InvalidConfigurationError.for_profile(profile)
+
+        if config.is_valid_with_token:
+            return {'token': config.token}
+        else:
+            return {'user': config.username, 'password': config.password}
+
+    def deploy(self, spec, headers=None):
+        lib_objects = LibraryObject.convert_from_libraries(spec.get('libraries', []))
+        local_lib_objects, rest_lib_objects = self._partition_libraries_and_extract_local_paths(lib_objects)
+        remote_lib_objects = self._get_remote_mappings(local_lib_objects)
+        upload_files = self._get_files_to_upload(local_lib_objects, remote_lib_objects)
+
+        for llo, rlo in upload_files:
+            try:
+                self.dbfs_client.put_file(llo.path, rlo.path, False)
+            except Exception as e:
+                raise RuntimeError('Error \'{}\' while uploading {}'.format(e, llo.path))
+
+        rest_lib_objects.extend(remote_lib_objects)
+        spec['libraries'] = LibraryObject.convert_to_libraries(rest_lib_objects)
+        spec['credentials'] = self._get_credentials_for_request()
+        self.client.client.perform_query('PUT', '/pipelines/{}'.format(spec['id']), data=spec, headers=headers)
+
+    def delete(self, pipeline_id, headers=None):
+        self.client.delete(pipeline_id, self._get_credentials_for_request(), headers)
+
+
+class LibraryObject(object):
+    def __init__(self, lib_type, lib_path):
+        self.path = lib_path
+        self.lib_type = lib_type
+
+    @classmethod
+    def convert_from_libraries(cls, libraries):
+        lib_objects = []
+        for library in libraries:
+            for lib_type, path in library.items():
+                lib_objects.append(LibraryObject(lib_type, path))
+        return lib_objects
+
+    @classmethod
+    def convert_to_libraries(cls, lib_objects):
+        libraries = []
+        for lib_object in lib_objects:
+            libraries.append({lib_object.lib_type: lib_object.path})
+        return libraries
--- a/databricks_cli/pipelines/api.py
+++ b/databricks_cli/pipelines/api.py
@@ -37,7 +37,7 @@ from databricks_cli.configure.provider import get_config, ProfileConfigProvider
 from databricks_cli.utils import InvalidConfigurationError
 
 BUFFER_SIZE = 1024 * 64
-base_pipelines_dir = 'dbfs:/pipelines/code/'
+base_pipelines_dir = 'dbfs:/pipelines/code'
 
 
 class PipelinesApi(object):
@@ -107,7 +107,7 @@ class PipelinesApi(object):
     def _get_hashed_path(path):
         """
         Finds the corresponding dbfs file path for the file located at the supplied path by
-        calculating its hash.
+        calculating its hash using SHA1.
         :param path: Local File Path
         :return: Remote Path (pipeline_base_dir + file_hash (dot) file_extension)
         """
@@ -123,7 +123,8 @@ class PipelinesApi(object):
             raise RuntimeError('Error \'{}\' while processing {}'.format(e, path))
 
         file_hash = hash_buffer.hexdigest()
-        path = '{}{}{}'.format(base_pipelines_dir, file_hash, os.path.splitext(path)[1])
+        # splittext includes the period in the extension
+        path = '{}/{}{}'.format(base_pipelines_dir, file_hash, os.path.splitext(path)[1])
         return path
 
     @staticmethod
--- a/databricks_cli/pipelines/api.py
+++ b/databricks_cli/pipelines/api.py
@@ -47,12 +47,6 @@ class PipelinesApi(object):
         self.dbfs_client = DbfsApi(api_client)
 
     def deploy(self, spec, headers=None):
-        """
-        Pipelines Deploy API
-        :param spec:
-        :param headers:
-        :return:
-        """
         lib_objects = LibraryObject.convert_from_libraries(spec.get('libraries', []))
         local_lib_objects, rest_lib_objects = \
             self._partition_libraries_and_extract_local_paths(lib_objects)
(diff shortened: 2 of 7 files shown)


databricks_cli/pipelines/cli.py
```diff
@@ -0,0 +1,121 @@
...
+             short_help='Utility to interact with the Databricks Delta Pipelines.')
+@click.option('--version', '-v', is_flag=True, callback=print_version_callback,
+              expose_value=False, is_eager=True, help=version)
+@debug_option
+@profile_option
+def pipelines_group():
+    """
+    Utility to interact with the Databricks pipelines.
+    """
+    pass
```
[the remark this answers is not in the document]
arulajmani: This is used to add pipelines so that you can say databricks pipelines ... 

pipelines group is added to the databricks command, and deploy/delete are added to pipelines group. By itself, the function doesn't do much. The short help string is printed if someone does databricks pipelines -h

databricks_cli/click_types.py
```diff
@@ -82,6 +82,11 @@ class SecretPrincipalClickType(ParamType):
     help = 'The name of the principal.'
 
 
+class PipelinesSpecPathClickType(ParamType):
```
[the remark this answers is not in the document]
arulajmani: Changing to PipelineSpecClickType and the name to 'SPEC'

databricks_cli/pipelines/cli.py
```diff
@@ -0,0 +1,121 @@
...
+from databricks_cli.version import print_version_callback, version
+from databricks_cli.pipelines.api import PipelinesApi
+from databricks_cli.configure.config import provide_api_client, profile_option, debug_option
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Deploys a delta pipeline')
+@click.argument('spec_arg', default=None, required=False)
+@click.option('--spec', default=None, help=PipelinesSpecPathClickType.help)
+@debug_option
```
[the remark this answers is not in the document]
arulajmani: Options need to be applied on every individual command, even the ones that are common across the whole group (eg. you can't pull out debug option and apply it to the pipelines group and have that work with both deploy and delete)

databricks_cli/sdk/service.py
```diff
@@ -706,3 +706,45 @@ def get_groups_for_principal(self, user_name=None, group_name=None, headers=None
...
+        if id is not None:
+            _data['id'] = id
+        if name is not None:
+            _data['name'] = name
+        if storage is not None:
+            _data['storage'] = storage
+        if filters is not None:
+            _data['filters'] = filters
+            if not isinstance(filters, dict):
+                raise TypeError('Expected databricks.Filters() or dict for field filters')
```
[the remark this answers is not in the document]
arulajmani: This is from the autogeneration, I'm not completely sure why it only checks Filters and Credentials. 
Also this is being bypassed because of the self.client.client

tests/pipelines/test_api.py
```diff
@@ -0,0 +1,197 @@
...
+        DeltaPipelinesServiceMock.return_value = mock.MagicMock()
+        _pipelines_api = api.PipelinesApi(None)
+        yield _pipelines_api
+
+
+def assert_library_object_lists_are_same(actual_llo, expected_llo):
+    assert len(actual_llo) == len(expected_llo)
+    for actual, expected in zip(actual_llo, expected_llo):
+        assert actual.lib_type == expected.lib_type
+        assert actual.path == expected.path
```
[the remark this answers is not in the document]
arulajmani: There's a better way by overloading the __eq__() method in the class.  I'll switch to that.

tests/pipelines/test_api.py
```diff
@@ -0,0 +1,194 @@
...
+def test_partition_correctly_identifies_local_paths(pipelines_api):
+    libraries = [
+        LibraryObject('jar', '//absolute/path/abc.ext'),
+        LibraryObject('jar', '/relative/path.ext'),
+        LibraryObject('jar', 'file://file/scheme/abs/path.ext'),
+        LibraryObject('jar', 'file:/file/scheme/relative/path.ext'),
+        LibraryObject('jar', 'FILE:/all/caps.ext'),
+        LibraryObject('jar', 'FiLe:/weird/case.ext'),
+        LibraryObject('jar', 'file.ext')
+    ]
```
mukulmurthy: I thought we said if it starts with "/", it's an absolute path, not relative?

tests/pipelines/test_api.py
```diff
@@ -0,0 +1,194 @@
...
+        LibraryObject('jar', 's3:file:/file/scheme/relative/path.ext'),
+        LibraryObject('jar', 'scheme:file.ext'),
+        LibraryObject('jar', 'scheme://abs/path.ext'),
+        LibraryObject('jar', 'scheme:/rel/path.ext'),
+        LibraryObject('egg', 'file:/rel/path.ext'),
+        LibraryObject('whl', '/rel/path.ext')
+    ]
+    llo, rest = pipelines_api._partition_libraries_and_extract_local_paths(libraries)
+    assert llo == expected_llo
+    assert rest == expected_rest
```
mukulmurthy: What are this test and the one above testing that you're not testing in test_deploy? the only thing is different schemes and relative paths, right? But here you're never actually testing the ability to use relative paths. I'd make test_deploy take in the list of input paths (like below you have [{'jar': jar1}, {'jar': jar2}, {'jar': 'dbfs:/pipelines/code/file.jar'}]) and the expected output paths and verify these that way.

databricks_cli/pipelines/api.py
```diff
@@ -150,3 +150,8 @@ def convert_to_libraries(cls, lib_objects):
         for lib_object in lib_objects:
             libraries.append({lib_object.lib_type: lib_object.path})
         return libraries
+
+    def __eq__(self, other):
+        if not isinstance(other, LibraryObject):
+            return NotImplemented
```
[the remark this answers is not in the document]
arulajmani: Python convention

databricks_cli/pipelines/api.py
```diff
@@ -0,0 +1,157 @@
...
+            return {'token': config.token}
+        else:
+            return {'user': config.username, 'password': config.password}
+
+    def deploy(self, spec, headers=None):
+        lib_objects = LibraryObject.convert_from_libraries(spec.get('libraries', []))
+        local_lib_objects, rest_lib_objects = \
+            self._partition_libraries_and_extract_local_paths(lib_objects)
+        remote_lib_objects = self._get_remote_mappings(local_lib_objects)
+        upload_files = self._get_files_to_upload(local_lib_objects, remote_lib_objects)
```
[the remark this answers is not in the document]
arulajmani: Might be hard to fold together because the return values are needed further down in the computation. 
I've removed _get_remote_mappings because it was a simple map on a list and documented all other functions.

databricks_cli/pipelines/api.py
```diff
@@ -0,0 +1,157 @@
...
+
+    @staticmethod
+    def _partition_libraries_and_extract_local_paths(lib_objects):
+        local_lib_objects, rest_lib_objects = [], []
+        for lib_object in lib_objects:
+            uri_scheme = urllib.parse.urlsplit(lib_object.path).scheme
+            if lib_object.lib_type == 'jar' and uri_scheme == '':
+                local_lib_objects.append(lib_object)
+            elif lib_object.lib_type == 'jar' and uri_scheme.lower() == 'file':
+                local_lib_objects.append(LibraryObject(lib_object.lib_type, lib_object.path[5:]))
```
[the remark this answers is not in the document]
arulajmani: Yes, everything in file: scheme has to be an absolute path. I'm updating the api doc to reflect this change.

databricks_cli/pipelines/api.py
```diff
@@ -38,28 +38,80 @@
...
     def __init__(self, api_client):
         self.client = DeltaPipelinesService(api_client)
         self.dbfs_client = DbfsApi(api_client)
 
+    def deploy(self, spec, headers=None):
+        """
+        Pipelines Deploy API
+        :param spec:
+        :param headers:
+        :return:
```
mukulmurthy: docs

databricks_cli/pipelines/api.py
```diff
@@ -38,28 +38,80 @@
...
+        upload_files = self._get_files_to_upload(local_lib_objects, remote_lib_objects)
+
+        for llo, rlo in upload_files:
+            try:
+                self.dbfs_client.put_file(llo.path, rlo.path, False)
+            except Exception as e:
+                raise RuntimeError('Error \'{}\' while uploading {}'.format(e, llo.path))
+
+        rest_lib_objects.extend(remote_lib_objects)
+        spec['libraries'] = LibraryObject.convert_to_libraries(rest_lib_objects)
```
mukulmurthy: Nit: We're not overloading rest_lib_objects to first be the non-upload files and later all the files? Can we write this functionally as LibraryObject.convert_to_libraries(rest_lib_objects + remote_lib_objects) ?

databricks_cli/pipelines/api.py
```diff
@@ -38,28 +38,80 @@
...
+                                         data=spec,
+                                         headers=headers)
+
+    def delete(self, pipeline_id, headers=None):
+        """
+        Pipelines Delete API
+        :param pipeline_id:
+        :param headers:
+        :return:
+        """
```
mukulmurthy: docs

databricks_cli/pipelines/api.py
```diff
@@ -38,28 +38,80 @@
...
         local_lib_objects, rest_lib_objects = [], []
         for lib_object in lib_objects:
             uri_scheme = urllib.parse.urlsplit(lib_object.path).scheme
             if lib_object.lib_type == 'jar' and uri_scheme == '':
                 local_lib_objects.append(lib_object)
             elif lib_object.lib_type == 'jar' and uri_scheme.lower() == 'file':
+                if lib_object.path.startswith('file:////'):
+                    raise RuntimeError('Invalid file uri scheme')
                 local_lib_objects.append(LibraryObject(lib_object.lib_type, lib_object.path[5:]))
```
mukulmurthy: 1. https://en.wikipedia.org/wiki/File_URI_scheme#How_many_slashes? says that file:// is technically invalid. I'm still down to support it, but if you want to disallow it that's reasonable too.

2. That startswith check needs to be case insensitive.

3. If we strip off the first 5 characters, and the file starts with file:///path, we'll end up with ///path. Does this work when you try to read the file later on?

databricks_cli/pipelines/api.py
```diff
@@ -72,27 +124,30 @@ def _get_hashed_path(path):
...
+        path = '{}{}{}'.format(base_pipelines_dir, file_hash, os.path.splitext(path)[1])
         return path
 
-    def _get_remote_mappings(self, local_lib_objects):
-        return list(map(lambda llo: LibraryObject(llo.lib_type, self._get_hashed_path(llo.path)),
-                        local_lib_objects))
-
     def _get_files_to_upload(self, local_lib_objects, remote_lib_objects):
+        """
+        Returns a Local/Remote pair for every file that needs to be uploaded to dbfs
```
mukulmurthy: I'd mention here that files are stored under base_pipeline_dir and named after the hashes of their contents, and we only upload files that don't already exist there.

databricks_cli/pipelines/api.py
```diff
@@ -72,27 +124,30 @@ def _get_hashed_path(path):
...
-    # Only required until the deploy/delete APIs requires the credentials in the body as well
-    # as the header.Once the API requirement is relaxed, this function can be stripped out and
-    # includes for this function removed.
-    #
     @staticmethod
     def _get_credentials_for_request():
+        """
+        Only required until the deploy/delete APIs requires the credentials in the body as well
+        as the header.Once the API requirement is relaxed, this function can be stripped out and
+        includes for this function removed.
```
mukulmurthy: Only required while* and require*

databricks_cli/pipelines/api.py
```diff
@@ -146,12 +182,22 @@ def convert_from_libraries(cls, libraries):
...
         for lib_object in lib_objects:
             libraries.append({lib_object.lib_type: lib_object.path})
         return libraries
 
     def __eq__(self, other):
+        """
+        Equality Comparison
+        :param other:
+        :return:
+        """
```
mukulmurthy: It's probably good to document all the stuff if we're going to include this. For this one, I think it's pretty obvious what it does and so the doc can be removed.

tests/pipelines/test_api.py
```diff
@@ -192,3 +143,74 @@ def test_delete(get_credentials_mock, pipelines_api):
...
+        LibraryObject('jar', 's3:file:/file/scheme/abs/path.ext'),
+        LibraryObject('jar', 'scheme:file.ext'),
+        LibraryObject('jar', 'scheme:/abs/path.ext'),
+        LibraryObject('jar', 'scheme://abs/path.ext'),
+        LibraryObject('egg', 'file:/abs/path.ext'),
+        LibraryObject('whl', 'rel/path.ext')
+    ]
+    llo, rest = pipelines_api._partition_libraries_and_extract_local_paths(libraries)
+    assert llo == expected_llo
+    assert rest == expected_rest
```
mukulmurthy: just for completeness, let's make sure that expected_llo + expected_rest == libraries

tests/pipelines/test_api.py
```diff
@@ -192,3 +143,74 @@ def test_delete(get_credentials_mock, pipelines_api):
...
+def test_partition_local_remote(pipelines_api):
+    libraries = [
+        LibraryObject('jar', '/absolute/path/abc.ext'),
+        LibraryObject('jar', 'relative/path.ext'),
+        LibraryObject('jar', 'file:///file/scheme/abs/path.ext'),
+        LibraryObject('jar', 'file://file/scheme/two/slashes/path.ext'),
+        LibraryObject('jar', 'FILE:/all/caps.ext'),
+        LibraryObject('jar', 'FiLe:/weird/case.ext'),
+        LibraryObject('jar', 'file.ext'),
+        # remote
```
mukulmurthy: instead of remote, I'd say shouldn't be uploaded, and the top can say should be uploaded.

databricks_cli/pipelines/api.py
arulajmani: 1. It is technically invalid, but seems like it is still common. Another option is to be like dbfs, which only allows a single slash and nothing else. 

3. yes, ///path and //path and /path are equivalent when trying to open the file.

tests/pipelines/test_api.py
arulajmani: Changed to remote, I'm commenting the top as "local files" because should be uploading is a bit misleading, as we only upload the files that don't already exist.

databricks_cli/pipelines/api.py
arulajmani: This allows to save a copy, as extend modifies rest_lib_objects in place whereas + returns a new list. That being said, using + is more readable so I'm changing to that.

tests/pipelines/test_api.py
arulajmani: That's not true because the function also strips out the file: prefix.

databricks_cli/pipelines/api.py
```diff
@@ -67,37 +61,31 @@ def deploy(self, spec, headers=None):
...
+        :return: List[List[LibraryObject], List[LibraryObject]] [Local, Remote]
         """
         local_lib_objects, rest_lib_objects = [], []
         for lib_object in lib_objects:
             uri_scheme = urllib.parse.urlsplit(lib_object.path).scheme
             if lib_object.lib_type == 'jar' and uri_scheme == '':
                 local_lib_objects.append(lib_object)
             elif lib_object.lib_type == 'jar' and uri_scheme.lower() == 'file':
-                if lib_object.path.startswith('file:////'):
+                if lib_object.path[4:].startswith(':////'):
```
mukulmurthy: does something like stripprefix(scheme) work?

databricks_cli/pipelines/cli.py
```diff
@@ -0,0 +1,121 @@
...
+    Reads the spec at src as a JSON if no file extension is provided, or if in the extension format
+    if the format is supported.
+    :param src:
+    :return: json of the spec
+    """
+    extension = os.path.splitext(src)[1]
+    if extension == '' or extension.lower() == '.json':
+        return _read_json_spec(src)
+    else:
+        raise RuntimeError('The provided file extension for the spec is not supported')
```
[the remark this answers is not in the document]
andrewmchen: Is this consistent with how we handle JSON files everywhere else? I don't recall checking the extensions anywhere...

databricks_cli/pipelines/api.py
```diff
@@ -0,0 +1,187 @@
...
+BUFFER_SIZE = 1024 * 64
+base_pipelines_dir = 'dbfs:/pipelines/code/'
+
+
+class PipelinesApi(object):
+    def __init__(self, api_client):
+        self.client = DeltaPipelinesService(api_client)
+        self.dbfs_client = DbfsApi(api_client)
+
+    def deploy(self, spec, headers=None):
```
andrewmchen: Is there a place where we document where this spec might look like?

databricks_cli/pipelines/api.py
```diff
@@ -0,0 +1,187 @@
...
+        lib_objects = LibraryObject.convert_from_libraries(spec.get('libraries', []))
+        local_lib_objects, rest_lib_objects = \
+            self._partition_libraries_and_extract_local_paths(lib_objects)
+        remote_lib_objects = list(map(lambda llo:
```
andrewmchen: nit: can we use list comprehensions wherever we try to use map?

databricks_cli/pipelines/api.py
```diff
@@ -0,0 +1,187 @@
...
+            return {'user': config.username, 'password': config.password}
+
+
+class LibraryObject(object):
+    def __init__(self, lib_type, lib_path):
+        self.path = lib_path
+        self.lib_type = lib_type
+
+    @classmethod
+    def convert_from_libraries(cls, libraries):
```
andrewmchen: nit: can we rename this to something a bit more descriptive? `from_libraries` is a bit weird since you're returning a list of LibraryObjects. Maybe `from_json`?

databricks_cli/pipelines/api.py
arulajmani: There is a link to an example spec in the doc https://docs.google.com/document/d/1K3iaOzfv0OvIsvZsdqiFDfvVFMh8Oh58wA2x0P8rtHQ/edit#heading=h.twb6crp48eby

databricks_cli/pipelines/cli.py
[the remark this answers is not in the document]
arulajmani: In later versions, we will be supporting other formats like yaml specs, thats why the check

databricks_cli/pipelines/api.py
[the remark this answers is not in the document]
andrewmchen: I'm wondering about this as well. How come we don't use something standard like what is suggested here? https://stackoverflow.com/questions/5977576/is-there-a-convenient-way-to-map-a-file-uri-to-os-path

databricks_cli/pipelines/api.py
```diff
@@ -0,0 +1,187 @@
...
+                                         data=spec,
+                                         headers=headers)
+
+    def delete(self, pipeline_id, headers=None):
+        self.client.delete(pipeline_id, self._get_credentials_for_request(), headers)
+
+    @staticmethod
+    def _partition_libraries_and_extract_local_paths(lib_objects):
+        """
+        Partitions the given set of libraries into local and remote by checking uri scheme
```
andrewmchen: Can you describe the criteria of what makes a local/remote library?

databricks_cli/pipelines/api.py
andrewmchen: I think it'd be very valuable to document it in this repo as well: in code as a comment and as an example somewhere as well.

databricks_cli/sdk/service.py
```diff
@@ -706,3 +706,45 @@ def get_groups_for_principal(self, user_name=None, group_name=None, headers=None
             _data['group_name'] = group_name
         return self.client.perform_query('GET', '/groups/list-parents', data=_data, headers=headers)
 
+
+class DeltaPipelinesService(object):
```
andrewmchen: Is this hand written? We should avoid doing that since this file is autogenerated and these changes will get clobbered next time someone runs the autogenerator.

arulajmani: It is autogenerated, but the autogeneration PR is not merged into Universe yet. I've created a JIRA task to ensure that when the Pipelines branch on the CLI gets merged with master, so does the PR that does the autogeneration get merged into Universe. 
The reason to hold off on the merging into Universe is we don't want someone trying to add their component and accidentally adding the Pipelines service as part of that.

databricks_cli/pipelines/api.py
arulajmani: This is specifically library json objects (not any json). I'm thinking of renaming it to from_json_libraries to be more clear. Does that work?

andrewmchen: I don't feel strongly about it but I think `from_json` is better still. Library is implicit from `LibraryObjects.from_json`. Your choice though.

databricks_cli/pipelines/api.py
arulajmani: Local libraries are present on the user machine which can be either prefixed by file: or no scheme at all (described in the doc)
All other libraries in the spec are remote. Currently, they should be dbfs paths, but this will probably be expanded in the future. I'll update the documentation to briefly describe this.

databricks_cli/pipelines/api.py
andrewmchen: nit: `Once the API requirement is relaxed, this function can be stripped out includes for this function removed.` -> `Once the API requirement is relaxed, we can remove this function`.

databricks_cli/pipelines/api.py
mukulmurthy: I don't think we should put an example spec here, since the spec may change in the future and this function is (mostly) forwarding the spec to our delta pipelines API. The spec specification and examples should live on the docs site (and we'll add a link to that from here when we get that up). 

I do think we should add documentation here saying:
 - What a spec is (a Delta Pipelines deployment specification that explains how to run a Delta Pipeline on Databricks)
 - That it gets forwarded to the Delta Pipelines deployment API
 - What we actually do check for and do with it in this function (look for all its libraries that are specified as local paths and upload them to DBFS).

databricks_cli/pipelines/api.py
andrewmchen: Ah, I see why I'm confused. I think you overload the term "remote" in your PR here. One describes the remote versions of the local libraries you are trying to upload and the other describes libraries which actually sit on S3/DBFS. Can we make sure that we don't overload the word `remote` in docstrings and variable names?

andrewmchen: It might be more clear if you split PipelinesApi.deploy into three clear steps
1. Identify what are the local libs
2. Write a function called `uploadLocalLibs` which identifies where the local libs should go and which ones should be uploaded. The function can return the uploaded versions of the local libs.
3. Replace the libraries field with the union of return of 2 and remotes identified in 1.

databricks_cli/pipelines/api.py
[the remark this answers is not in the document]
arulajmani: This doesn't seem to work when the uri is of the form "file://abc..." 

Strictly speaking, the standard says that there should be either 1 or 3 slashes in the file scheme, but 2 slashes are common. [5:] means that 1, 2, or 3 slashes work. The method described in the stackoverflow will mean that only 1 and 3 slashes work. I can switch this if that is the functionality we want. 
@mukulmurthy

databricks_cli/pipelines/api.py
arulajmani: Changed the deploy flow to what you mentioned. 
To decrease confusion further, I've renamed "rest" to "external".

databricks_cli/pipelines/api.py
arulajmani: I've added this documentation in pipelines/cli.py deploy

databricks_cli/pipelines/api.py
```diff
@@ -0,0 +1,182 @@
...
+        :return: List[List[LibraryObject], List[LibraryObject]] ([Local, External])
+        """
+        local_lib_objects, external_lib_objects = [], []
+        for lib_object in lib_objects:
+            parsed_uri = urllib.parse.urlparse(lib_object.path)
+            if lib_object.lib_type == 'jar' and parsed_uri.scheme == '':
+                local_lib_objects.append(lib_object)
+            elif lib_object.lib_type == 'jar' and parsed_uri.scheme.lower() == 'file':
+                # exactly 1 or 3
+                if parsed_uri.path.startswith('//') or parsed_uri.netloc != '':
```
andrewmchen: How does the path start with a `//`?

databricks_cli/pipelines/api.py
```diff
@@ -0,0 +1,182 @@
...
+        """
+        hash_buffer = sha1()
+        try:
+            with open(path, 'rb') as f:
+                while True:
+                    data = f.read(BUFFER_SIZE)
+                    if not data:
+                        break
+                    hash_buffer.update(data)
+        except Exception as e:
```
andrewmchen: Why do we need to catch these?

databricks_cli/pipelines/api.py
```diff
@@ -0,0 +1,182 @@
...
+            except Exception as e:
+                raise RuntimeError('Error \'{}\' while uploading {}'.format(e, llo.path))
+
+        return remote_lib_objects
+
+    @staticmethod
+    def _get_hashed_path(path):
+        """
+        Finds the corresponding dbfs file path for the file located at the supplied path by
+        calculating its hash.
```
andrewmchen: nit: specify what hash algorithm.

databricks_cli/pipelines/api.py
```diff
@@ -0,0 +1,182 @@
...
+                while True:
+                    data = f.read(BUFFER_SIZE)
+                    if not data:
+                        break
+                    hash_buffer.update(data)
+        except Exception as e:
+            raise RuntimeError('Error \'{}\' while processing {}'.format(e, path))
+
+        file_hash = hash_buffer.hexdigest()
+        path = '{}{}{}'.format(base_pipelines_dir, file_hash, os.path.splitext(path)[1])
```
andrewmchen: typically, these look better when `base_pipelines_dir` doesn't end with a slash. Also are you missing a period here?

andrewmchen: I take back the period question. Could you leave a small comment that says splitext includes the period though?

tests/pipelines/test_api.py
```diff
@@ -0,0 +1,207 @@
...
+    3. Local files already present in dbfs
+    4. Local files that need to be uploaded to dbfs
+    Every test local file which has '123' written to it is expected to be already present in Dbfs.
+    A test local file which has '456' written to it is not present in Dbfs and therefore must be.
+    uploaded to dbfs.
+    """
+    get_credentials_mock.return_value = CREDENTIALS
+    deploy_mock = pipelines_api.client.client.perform_query
+    # set-up the test
+    jar1 = tmpdir.join('/jar1.jar').strpath
```
andrewmchen: I think you just need to do tmpdir.join('jar1.jar'). Btw, I found this when I was reading your tests. tmpdir is a `py.path.local` object I believe so you can test in an interpreter...

```
NameError: name 'py' is not defined
>>> import py
>>> py.path.local('a')
local('/Users/andrew.chen/databricks-cli/a')
>>> a = _
>>> a.join('a')
local('/Users/andrew.chen/databricks-cli/a/a')
>>>
```

tests/pipelines/test_api.py
```diff
@@ -0,0 +1,207 @@
...
+    with open(jar3, 'w') as f:
+        f.write('456')
+    with open(jar4, 'w') as f:
+        f.write('456')
+    libraries = [{'jar': 'dbfs:/pipelines/code/file.jar'},
+                 {'jar': jar1},
+                 {'jar': jar2},
+                 {'jar': jar3_relpath},
+                 {'jar': jar4_file_prefix}]
+    SPEC['libraries'] = libraries
```
andrewmchen: this looks really weird. Why are we mutating an module level object which looks like a constant?

databricks_cli/pipelines/api.py
```diff
@@ -0,0 +1,182 @@
...
+
+        transformed_remote_lib_objects = [LibraryObject(rlo.lib_type, DbfsPath(rlo.path))
+                                          for rlo in remote_lib_objects]
+        upload_files = [llo_tuple for llo_tuple in
+                        zip(local_lib_objects, transformed_remote_lib_objects)
+                        if not self.dbfs_client.file_exists(llo_tuple[1].path)]
+
+        for llo, rlo in upload_files:
+            try:
+                self.dbfs_client.put_file(llo.path, rlo.path, False)
```
andrewmchen: Won't this fail your test case below with the same file contents?

databricks_cli/pipelines/api.py
arulajmani: If 4 or more slashes, like file:////

databricks_cli/pipelines/api.py
arulajmani: This function is mocked in the test

databricks_cli/pipelines/api.py
arulajmani: Should we not?

tests/pipelines/test_api.py
arulajmani: I agree it looks weird, I'm switching to deepcopying the SPEC to spec and then adding the libraries field to it.

PR #244, opened 2019-07-08
    The password provided to configure shouldn't be echo'd.

areese:
Add a --stdin option to configure, so that the password will always be read from stdin instead of stored in a file.

### DIFF
--- a/databricks_cli/configure/cli.py
+++ b/databricks_cli/configure/cli.py
@@ -85,7 +85,7 @@ def configure_cli(token, insecure, stdin):
     if token:
         _configure_cli_token(profile, insecure_str)
     elif stdin:
-        _configure_cli_stdin()
+        _configure_cli_stdin(profile, insecure_str)
     else:
         _configure_cli_password(profile, insecure_str)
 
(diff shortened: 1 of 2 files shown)


areese:
This is very similar to: https://github.com/databricks/databricks-cli/pull/123

databricks_cli/configure/cli.py
```diff
@@ -32,18 +32,27 @@
...
+    token = click.prompt(PROMPT_TOKEN, default=config.token, hide_input=True)
     new_config = DatabricksConfig.from_token(host, token, insecure)
     update_and_persist_config(profile, new_config)
 
 
+def _configure_cli_stdin(profile, insecure):
+    config = ProfileConfigProvider(profile).get_config() or DatabricksConfig.empty()
+    host = click.prompt(PROMPT_HOST, default=config.host, type=_DbfsHost())
+    username = click.prompt(PROMPT_USERNAME, default=config.username)
+    password = 'stdin'
```
andrewmchen: What if my password is `stdin`? I know this password is probably not allowed in most sign up workflows in Databricks but I'd rather not special case this password..

Can we just add a flag called `read_password_from_stdin` or something?

areese: excellent point, let me rework this that way.
How about empty password?  Would that be an ok choice or would you rather use a flag?

andrewmchen: I think a flag is probably just the cleanest way.

areese:
I'm going to close this.  I ended up with a new patch to hide the token, and I never ended up using the stdin option after all.

PR #247, opened 2019-07-08
    Add runs get output command

areese:


databricks_cli/runs/cli.py
```diff
@@ -147,7 +147,23 @@ def runs_group():
...
+    The output schema is documented https://docs.databricks.com/api/latest/jobs.html#runs-get-output
+    """
+    click.echo(pretty_format(RunsApi(api_client).get_run_output(run_id)))
+
+
 runs_group.add_command(submit_cli, name='submit')
 runs_group.add_command(list_cli, name='list')
 runs_group.add_command(get_cli, name='get')
 runs_group.add_command(cancel_cli, name='cancel')
+runs_group.add_command(get_output_cli, name='get_output')
```
andrewmchen: This should be `get-output`.

areese: Thank you for catching that. I had fixed that in my branch, but I didn't get those changes to this branch.

PR #248, opened 2019-07-08
    Add a debugging flag

areese:
If --debug is used or if DATABRICKS_DEBUGGING is set in the environment the entire contents of the http requests will be printed to stdout.

Depends on https://github.com/databricks/databricks-cli/pull/244

### DIFF
--- a/databricks_cli/click_types.py
+++ b/databricks_cli/click_types.py
@@ -20,7 +20,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import logging
 
 import click
 from click import ParamType, Option, MissingParameter, UsageError
--- a/databricks_cli/click_types.py
+++ b/databricks_cli/click_types.py
@@ -20,6 +20,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import logging
 
 from click import ParamType, Option, MissingParameter, UsageError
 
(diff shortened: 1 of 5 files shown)


databricks_cli/sdk/api_client.py
```diff
@@ -129,6 +130,27 @@ def perform_query(self, method, path, data = {}, headers = None):
...
+            import http.client as http_client
+        except ImportError:
+            # Python 2
+            import httplib as http_client
+
+        print ("HTTP debugging enabled")
+        http_client.HTTPConnection.debuglevel = 1
+
+        # You must initialize logging, otherwise you'll not see debug output.
+        logging.basicConfig()
```
andrewmchen: What do you think about putting the logic here? https://github.com/databricks/databricks-cli/blob/master/databricks_cli/click_types.py#L104.

areese: Let me move it there, that is a better place.  This set of patches is older, so I did have the same understanding I do now.

areese:
@andrewmchen Updated, using the --debug flag and ContextObject made the code a lot cleaner.

databricks_cli/sdk/api_client.py
areese: Moved thank you.

databricks_cli/click_types.py
```diff
@@ -104,6 +106,29 @@ def __init__(self):
...
+            import http.client as http_client
+        except ImportError:
+            # Python 2
+            import httplib as http_client
+
+        click.echo("HTTP debugging enabled")
+        http_client.HTTPConnection.debuglevel = 1
+
+        # You must initialize logging, otherwise you'll not see debug output.
+        logging.basicConfig()
```
andrewmchen: Why do we need this line? Looks like they will go to stderr by default?
> By default, no destination is set for any logging messages. You can specify a destination (such as console or file) by using basicConfig() as in the tutorial examples. If you call the functions debug(), info(), warning(), error() and critical(), they will check to see if no destination is set; and if one is not set, they will set a destination of the console (sys.stderr) and a default format for the displayed message before delegating to the root logger to do the actual message output.

andrewmchen: In fact, why do we need anything to do with requests.packages.urllib3? I removed all of that and it still worked.

areese: we might not need the request_log lines, lemme go back and check.  It's possible that since the patch was written, things have changed enough these lines are no longer needed.

areese: Fixed.

PR #249, opened 2019-07-31
Add support for wheels when deploying pipelines

arulajmani:
We should upload wheels to dbfs similar to how we upload jars to support Python APIs.


arulajmani:
@mukulmurthy

PR #258, opened 2019-09-07
[ES-14934] Retrying!

databricks-david-lewis:
Retry failed chunks!

### DIFF
--- a/databricks_cli/dbfs/api.py
+++ b/databricks_cli/dbfs/api.py
@@ -129,7 +129,7 @@ class DbfsApi(object):
                 except:
                     if retries_left > 0:
                         time.sleep(1)
-                        print 'Error reading chunk, trying again (', retries_left, ' left)!'
+                        print('Error reading chunk, trying again (' + str(retries_left) + ' left)!')
                         retries_left = retries_left - 1
                     else:
                         raise
--- a/databricks_cli/dbfs/api.py
+++ b/databricks_cli/dbfs/api.py
@@ -26,6 +26,7 @@ from base64 import b64encode, b64decode
 import os
 import shutil
 import tempfile
+import time
 
 import click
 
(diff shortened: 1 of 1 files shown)


databricks_cli/dbfs/api.py
```diff
@@ -115,13 +116,23 @@ def get_file(self, dbfs_path, dst_path, overwrite, headers=None):
...
-                local_file.write(b64decode(data))
+                try:
+                    response = self.client.read(dbfs_path.absolute_path, offset, BUFFER_SIZE_BYTES,
+                                                headers=headers)
+                    bytes_read = response['bytes_read']
+                    data = response['data']
+                    offset += bytes_read
+                    local_file.write(b64decode(data))
+                    retries_left = 5
+                except:
```
mparkhe: except as e:

databricks_cli/dbfs/api.py
```diff
@@ -115,13 +116,23 @@ def get_file(self, dbfs_path, dst_path, overwrite, headers=None):
...
+                    if retries_left > 0:
+                        time.sleep(1)
+                        print 'Error reading chunk, trying again (', retries_left, ' left)!'
```
mparkhe: if you want to print exception ...
`print(str(e) + 'Error reading ... ')`

databricks_cli/dbfs/api.py
```diff
@@ -115,13 +116,23 @@ def get_file(self, dbfs_path, dst_path, overwrite, headers=None):
...
+                        retries_left = retries_left - 1
+                    else:
+                        raise
```
mparkhe: `raise e`

databricks_cli/dbfs/api.py
```diff
@@ -115,13 +116,23 @@ def get_file(self, dbfs_path, dst_path, overwrite, headers=None):
...
+                try:
+                    response = self.client.read(dbfs_path.absolute_path, offset, BUFFER_SIZE_BYTES,
+                                                headers=headers)
+                    bytes_read = response['bytes_read']
+                    data = response['data']
+                    offset += bytes_read
+                    local_file.write(b64decode(data))
+                    retries_left = 5
+                except:
+                    if retries_left > 0:
```
andrewmchen: I think we should add a unit test for this addition piece of logic.

andrewmchen: Something like this should do the trick:
```

    def test_get_file_with_error(self, dbfs_api, tmpdir):
        api_mock = dbfs_api.client
        api_mock.get_status.return_value = TEST_FILE_JSON
        api_mock.read.side_effect = [Exception(), {
            'bytes_read': 1,
            'data': b64encode(b'x'),
        }]

        test_file_path = os.path.join(tmpdir.strpath, 'test')
        dbfs_api.get_file(TEST_DBFS_PATH, test_file_path, True)

        with open(test_file_path, 'r') as f:
            assert f.read() == 'x'
```

PR #259, opened 2019-09-07
Retry even harder on requests

thunterdb:
This is a follow-up to #258

### DIFF
--- a/databricks_cli/dbfs/api.py
+++ b/databricks_cli/dbfs/api.py
@@ -129,7 +129,7 @@ class DbfsApi(object):
                 except:
                     if retries_left > 0:
                         time.sleep(1)
-                        print 'Error reading chunk, trying again (', retries_left, ' left)!'
+                        print('Error reading chunk, trying again (' + str(retries_left) + ' left)!')
                         retries_left = retries_left - 1
                     else:
                         raise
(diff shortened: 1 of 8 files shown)


databricks_cli/dbfs/api.py
```diff
@@ -115,13 +116,24 @@ def get_file(self, dbfs_path, dst_path, overwrite, headers=None):
...
         with open(dst_path, 'wb') as local_file:
+            retries_left = 20
             while offset < length:
-                response = self.client.read(dbfs_path.absolute_path, offset, BUFFER_SIZE_BYTES,
-                                            headers=headers)
-                bytes_read = response['bytes_read']
-                data = response['data']
-                offset += bytes_read
-                local_file.write(b64decode(data))
+                try:
```
thunterdb: I think this logic should be moved to `DbfsClient`, or even to `ApiClient`. `GET` are supposed to be idempotent and the client can configure the retries.

bogdanghita-db:
The retries we added were only for 429 (see [here](https://github.com/databricks/databricks-cli/pull/343/files)).

nfx:
closing in favour of https://github.com/databricks/databricks-cli/commit/945690f2ac669e5b20be0bbd121976891b105f2d

Issue #264, opened 2019-10-24
Add support for cluster-scoped init_scripts when creating clusters.


adampicwell:
It looks like the proper way to specify cluster-scoped init_scripts is when creating clusters: https://docs.databricks.com/clusters/init-scripts.html

It looks like it was added in this PR #196 but only to `edit_cluster`. In any case, it was (inadvertently?) removed in PR #200.

Can we have this as a kwarg to `create_cluster` and make it available on the CLI as well?

chrisst:
**This repository has been archived.**

The legacy `databricks/databricks-cli` is deprecated and has been fully replaced by the new Databricks CLI at **https://github.com/databricks/cli**.

We are closing all open issues here as part of archiving this repository. If your issue still affects the new CLI, we'd appreciate you re-opening it at https://github.com/databricks/cli/issues so the team can track and address it there.

PR #266, opened 2019-10-29
Add 'init_scripts' optional argument

malthe:
This fixes issue #264.

Note that the 'init_scripts' argument was added previously to 'edit_cluster', but (inadvertently?) removed later.

This change adds it to both 'create_cluster' and 'edit_cluster'.


malthe:
Bump!

PR #274, opened 2019-12-12
Fix pipelines request paths for delete, get, and reset commands

mukulmurthy:
Previous change passed unit tests which made sure the correct call was made to the generated client, but didn't verify the generated client actually makes the right requests. Fixed service.py to send the right requests and updated unit tests to make sure the called paths are correct.

Also made the `databricks pipelines get` command print the server's response.

### DIFF
--- a/databricks_cli/pipelines/api.py
+++ b/databricks_cli/pipelines/api.py
@@ -48,16 +48,14 @@ class PipelinesApi(object):
 
         spec['libraries'] = LibraryObject.to_json(external_lib_objects +
                                                   self._upload_local_libraries(local_lib_objects))
-        self.client.client.perform_query('PUT',
-                                         '/pipelines/{}'.format(spec['id']),
-                                         data=spec,
-                                         headers=headers)
+        pipeline_id = spec['id']
+        self.client.deploy_spec(pipeline_id, spec, headers)
 
     def delete(self, pipeline_id, headers=None):
         self.client.delete(pipeline_id, headers)
 
     def get(self, pipeline_id, headers=None):
-        self.client.get(pipeline_id, headers)
+        return self.client.get(pipeline_id, headers)
 
     def reset(self, pipeline_id, headers=None):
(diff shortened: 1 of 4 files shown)


databricks_cli/pipelines/api.py
```diff
@@ -48,16 +48,14 @@ def deploy(self, spec, headers=None):
...
-                                         headers=headers)
+        pipeline_id = spec['id']
+        self.client.deploy_spec(pipeline_id, spec, headers)
 
     def delete(self, pipeline_id, headers=None):
         self.client.delete(pipeline_id, headers)
 
     def get(self, pipeline_id, headers=None):
-        self.client.get(pipeline_id, headers)
+        return self.client.get(pipeline_id, headers)
```
anew: :)

databricks_cli/sdk/service.py
```diff
@@ -821,21 +821,24 @@ def deploy(self, pipeline_id=None, id=None, name=None, storage=None, configurati
...
 
+    def deploy_spec(self, pipeline_id=None, spec={}, headers=None):
+        return self.client.perform_query('PUT', '/pipelines/{}'.format(pipeline_id), data=spec, headers=headers)
+
     def delete(self, pipeline_id=None, headers=None):
         _data = {}
         if pipeline_id is not None:
             _data['pipeline_id'] = pipeline_id
-        return self.client.perform_query('DELETE', '/pipelines/{pipeline_id}', data=_data, headers=headers)
+        return self.client.perform_query('DELETE', '/pipelines/{}'.format(pipeline_id), data=_data, headers=headers)
```
anew: this is an unrelated question because existing code did this: Why are we passing the pipeline_id in the URL as well as in the body of the request? Is it required in the body?

mukulmurthy: Not required, but that's just the generated way. Seems to be fine to send it in the body if it's the same.

PR #278, opened 2020-03-01
Introduce new profiles command group

thegreatmoonmoon:
* add new function to get all profiles from config
* add and register new cmd group profiles
* add new test module for the cmd group


thegreatmoonmoon:
Hi - would like to propose an addition of a new command group 'profiles' that lets the user check which profiles are present in the databricks configuration file directly from the command line utility. Could make it easier to check which profiles had been already configured.

chrisst:

PR #280, opened 2020-03-09
Add support for DBFS partial deletes

bogdanghita-db:
Introduce retry logic for handling `DBFS` partial deletes that occur as a result of a recursive delete with limit. A partial delete means that not all the contents in the directory were deleted because the limit was reached.

Tested with unit tests and manually.

### DIFF
--- a/databricks_cli/dbfs/api.py
+++ b/databricks_cli/dbfs/api.py
@@ -124,7 +124,17 @@ class DbfsApi(object):
                 local_file.write(b64decode(data))
 
     def delete(self, dbfs_path, recursive, headers=None):
-        self.client.delete(dbfs_path.absolute_path, recursive=recursive, headers=headers)
+        while True:
+            try:
+                self.client.delete(dbfs_path.absolute_path, recursive=recursive, headers=headers)
+            except HTTPError as e:
+                if (e.response.status_code and
+                    e.response.json()['error_code'] == 'REQUEST_LIMIT_EXCEEDED'):
+                    print "[debug] partial delete: %s" % e.response.json()['message']
+                    continue
+                else:
+                    raise e
+            break
 
     def mkdirs(self, dbfs_path, headers=None):
         self.client.mkdirs(dbfs_path.absolute_path, headers=headers)
--- a/databricks_cli/dbfs/api.py
+++ b/databricks_cli/dbfs/api.py
@@ -27,8 +27,8 @@ import os
 import shutil
 import tempfile
 
-import click
 import re
+import click
 
 from requests.exceptions import HTTPError
 
--- a/databricks_cli/dbfs/api.py
+++ b/databricks_cli/dbfs/api.py
@@ -39,7 +39,7 @@ from databricks_cli.dbfs.dbfs_path import DbfsPath
 from databricks_cli.dbfs.exceptions import LocalFileExistsException
 
 BUFFER_SIZE_BYTES = 2**20
-DELETE_MAX_CONSECUTIVE_503_ERRORS = 3
+DELETE_MAX_CONSECUTIVE_503_RETRIES = 3
 DELETE_503_RETRY_DELAY_MILLIS = 30 * 1000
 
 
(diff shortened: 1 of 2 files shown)


databricks_cli/dbfs/api.py
```diff
@@ -124,7 +125,17 @@ def get_file(self, dbfs_path, dst_path, overwrite, headers=None):
                 local_file.write(b64decode(data))
 
     def delete(self, dbfs_path, recursive, headers=None):
-        self.client.delete(dbfs_path.absolute_path, recursive=recursive, headers=headers)
+        while True:
+            try:
+                self.client.delete(dbfs_path.absolute_path, recursive=recursive, headers=headers)
+            except HTTPError as e:
+                # Handle partial delete exceptions and retry until all the files have been deleted
+                if (e.response.status_code == 429 and
```
mswit-databricks: If we go for 503, is it going to be possible to tell apart 503 "service unavailable" from 503 "partial delete" here?

databricks_cli/dbfs/api.py
```diff
@@ -124,7 +125,17 @@ def get_file(self, dbfs_path, dst_path, overwrite, headers=None):
...
+                        e.response.json()['error_code'] == DbfsErrorCodes.REQUEST_LIMIT_EXCEEDED):
+                    continue
```
mswit-databricks: It would be great to log sth to give indication of progress.

databricks_cli/dbfs/api.py
bogdanghita-db: No, unless we add a special field in the exception or change `error_code` to something specific to partial delete.

bogdanghita-db: Solved by using a dedicated `error_code` in the response.

databricks_cli/dbfs/api.py
bogdanghita-db: Done.

databricks_cli/dbfs/api.py
```diff
@@ -123,8 +134,48 @@ def get_file(self, dbfs_path, dst_path, overwrite, headers=None):
...
+                if e.response.status_code == 503:
+                    try:
+                        error_code = e.response.json()['error_code']
+                    except (AttributeError, KeyError):
+                        error_code = None
+                    # Handle partial delete exceptions: retry until all the files have been deleted
+                    if error_code == DbfsErrorCodes.PARTIAL_DELETE:
+                        try:
+                            num_files_deleted += self.get_num_files_deleted(e)
+                            click.echo("\rDeleted {} files. Delete in progress...\033[K".format(
```
andrewmchen: hm, what are the \r and \033[K? Are they color codes? There are helpers to do this here: https://click.palletsprojects.com/en/7.x/api/#click.style

bogdanghita-db: `\r` moves the cursor at the beginning of the line and `\033[K` clears the remaining chars until the end of the line. Their purpose is to interactively update the number of delete files.

databricks_cli/dbfs/api.py
```diff
@@ -123,8 +134,48 @@ def get_file(self, dbfs_path, dst_path, overwrite, headers=None):
...
+                                num_files_deleted), nl=False)
+                        except ParseException:
+                            click.echo("\rDelete in progress...\033[K", nl=False)
+                        num_consecutive_503_errors = 0
+                        continue
+                    # Retry at most DELETE_MAX_CONSECUTIVE_503_ERRORS times for other 503 errors
+                    elif num_consecutive_503_errors < DELETE_MAX_CONSECUTIVE_503_ERRORS:
```
mswit-databricks: I think it would be cleaner to change it such that the code bails out after exactly DELETE_MAX_CONSECUTIVE_503_ERRORS errors instead of DELETE_MAX_CONSECUTIVE_503_ERRORS + 1.

databricks_cli/dbfs/api.py
```diff
@@ -123,8 +134,48 @@ def get_file(self, dbfs_path, dst_path, overwrite, headers=None):
                 offset += bytes_read
                 local_file.write(b64decode(data))
 
+    def get_num_files_deleted(self, partial_delete_error):
```
andrewmchen: Could we add a unit test for this function?

bogdanghita-db: Done.

databricks_cli/dbfs/api.py
bogdanghita-db: My initial intention was to retry 3 times, but I chose a poor name for that constant. I performed renames such that the variable names reflect the intended behaviour.

PR #287, opened 2020-04-07
require configparser only in python < 3.6

acxz:
configparser comes in default python >= 3.6

See below for corresponding issue on the Arch Linux package: https://github.com/acxz/pkgbuilds/issues/21 and https://github.com/acxz/pkgbuilds/issues/20

TLDR: users should not be forced to install configparser from the pypi when it is included in default for python versions >= 3.6. This PR ensures that only for versions below 3.6 this requirement is kept.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setup(
         'requests>=2.17.3',
         'tabulate>=0.7.7',
         'six>=1.10.0',
-        'configparser >= 0.3.5',
+        'configparser>=0.3.5;python_version < "3.6"',
     ],
     entry_points='''
         [console_scripts]


acxz:
@andrewmchen or anyone else with maintainership authority, can you take a look at this? This is a crucial issue for python > 3.6. Thx in advance.

andrewmchen:
Could I ask some clarifying questions? I tried reproducing locally and could not. Here is what I tried:
```
$conda create -n test-configparser python=3
$ python --version
Python 3.8.2
$ pip install configparser>=0.3.5
$ pip list
Package      Version
------------ -------------------
certifi      2020.4.5.1
configparser 5.0.0
pip          20.0.2
setuptools   46.1.3.post20200330
wheel        0.34.2
```

acxz:
So while having configparser does not break the install. It overrides the existing configparser in python >= 3.6. The configparser that is in the setup.py is for backward compat only and redundant for python >= 3.6

acxz:
@andrewmchen any verdict on this?

andrewmchen:
Let's ship it. Thanks for pushing this over the finish line.

PR #296, opened 2020-05-19
Pipelines CLI Fixes

null-sleep:
This PR makes the following changes to the Pipelines CLI:

- A successful spec deployment prints the URL to the pipeline.
- If the provided spec file does not have an ID, one is generated and the spec file is updated with the new ID.
- Adds a new exception eater to print cleaner error messages from the pipelines API
- Adds exception handing for specs with invalid json
- A validation check for pipeline id: only -, _ and alphanumeric characters are allowed

Testing:
Unit Tests
Manually verified deployment URLs, new exception formatting, validation of pipeline ids

### DIFF
--- a/databricks_cli/pipelines/cli.py
+++ b/databricks_cli/pipelines/cli.py
@@ -1,0 +1,112 @@
+# Databricks CLI
+# Copyright 2017 Databricks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"), except
+# that the use of services to which certain application programming
+# interfaces (each, an "API") connect requires that the user first obtain
+# a license for the use of the APIs from Databricks, Inc. ("Databricks"),
+# by creating an account at www.databricks.com and agreeing to either (a)
+# the Community Edition Terms of Service, (b) the Databricks Terms of
+# Service, or (c) another written agreement between Licensee and Databricks
+# for the use of the APIs.
+#
+# You may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from json import loads as json_loads
+import os
+
+import click
+
+from databricks_cli.click_types import PipelineSpecClickType
+from databricks_cli.utils import eat_exceptions, CONTEXT_SETTINGS
+from databricks_cli.version import print_version_callback, version
+from databricks_cli.pipelines.api import PipelinesApi
+from databricks_cli.configure.config import provide_api_client, profile_option, debug_option
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Deploys a delta pipeline according to the pipeline specification')
+@click.argument('spec_arg', default=None, required=False)
+@click.option('--spec', default=None, help=PipelineSpecClickType.help)
+@debug_option
+@profile_option
+@eat_exceptions
+@provide_api_client
+def deploy_cli(api_client, spec_arg, spec):
+    """
+    Deploys a delta pipeline according to the pipeline specification.
+    * The pipeline spec is a deployment specification that explains how to run a
+    Delta Pipeline on Databricks.
+    * The CLI simply forwards the spec to Databricks.
+    * All the local libraries referenced in the spec are uploaded to DBFS.
+    """
+    if bool(spec_arg) == bool(spec):
+        raise RuntimeError('The spec should be provided either by an option or argument')
+    src = spec_arg if bool(spec_arg) else spec
+    spec_obj = _read_spec(src)
+    PipelinesApi(api_client).deploy(spec_obj)
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Stops a delta pipeline and cleans '
+                          'up Databricks resources associated with it')
+@click.argument('spec_arg', default=None, required=False)
+@click.option('--spec', default=None, help=PipelineSpecClickType.help)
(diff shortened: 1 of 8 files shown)


databricks_cli/utils.py
```diff
@@ -46,9 +46,39 @@ def decorator(*args, **kwargs):
                                'reconfigure with ``dbfs configure``')
             else:
                 error_and_quit(exception.response.content)
-        except Exception as exception: # noqa
+        except Exception as exception:  # noqa
             if not DEBUG_MODE:
                 error_and_quit('{}: {}'.format(type(exception).__name__, str(exception)))
+
+    decorator.__doc__ = function.__doc__
```
null-sleep, line 53: This is just formatting

databricks_cli/pipelines/cli.py
```diff
@@ -59,8 +69,20 @@ def deploy_cli(api_client, spec_arg, spec):
         raise RuntimeError('The spec should be provided either by an option or argument')
     src = spec_arg if bool(spec_arg) else spec
     spec_obj = _read_spec(src)
+    if 'id' not in spec_obj:
+        pipeline_id = str(uuid.uuid4())
+        click.echo("Updating spec with id: {}".format(pipeline_id))
+        spec_obj['id'] = pipeline_id
+        _write_spec(src, spec_obj)
+    elif not _validate_pipeline_id(spec_obj['id']):
+        error_and_quit("Pipeline id has invalid characters")
```
mukulmurthy, line 78: Say what's allowed and also print the id they provided. Also, maybe the error_and_quit should just be inside the _validate_pipeline_id call so you don't have to repeat those bits?

databricks_cli/pipelines/cli.py
```diff
@@ -59,8 +69,20 @@ def deploy_cli(api_client, spec_arg, spec):
...
     PipelinesApi(api_client).deploy(spec_obj)
 
+    pipeline_id = spec_obj['id']
+    base_url = "{0.scheme}://{0.netloc}/".format(urlparse(api_client.url))
+    pipeline_url = urljoin(base_url, "#joblist/pipelines/{}".format(pipeline_id))
```
mukulmurthy, line 83: File a task on our side to return this from deployment (and then also update this to use that).

mukulmurthy: Does this work for MT shards?

databricks_cli/pipelines/cli.py
```diff
@@ -59,8 +69,20 @@ def deploy_cli(api_client, spec_arg, spec):
...
+    click.echo(pipeline_url)
```
mukulmurthy, line 84: Better message, like "Pipeline {id} has been deployed and can be viewed at url"

databricks_cli/utils.py
```diff
@@ -46,9 +46,39 @@ def decorator(*args, **kwargs):
...
+    """
+
+    @six.wraps(function)
+    def decorator(*args, **kwargs):
+        try:
+            return function(*args, **kwargs)
+        except HTTPError as exception:  # noqa
+            if exception.response.status_code == 401:
+                error_and_quit('Your authentication information may be incorrect. Please '
+                               + 'reconfigure with ``dbfs configure``')
```
mukulmurthy, line 70: databricks configure

databricks_cli/utils.py
```diff
@@ -46,9 +46,39 @@ def decorator(*args, **kwargs):
...
+
+            elif exception.response.status_code == 400:
+                exp_context = json_loads(exception.response.content.decode('utf-8'))
+                message = exp_context['error_code'] + '\n' + exp_context['message']
```
mukulmurthy, line 74: Can we do this for all exceptions first (even before the 401), and only do the other ones if it fails to get this info?

tests/pipelines/test_cli.py
```diff
@@ -124,6 +124,19 @@ def test_delete_cli_incorrect_parameters(pipelines_api_mock, tmpdir):
...
+def test_deploy_spec_updated_with_id_if_pipeline_id_not_in_spec(pipelines_api_mock, tmpdir):  # noqa
+    path = tmpdir.join('/spec.json').strpath
+    spec_with_no_id = '{"name": "no id"}'
+    with open(path, 'w') as f:
+        f.write(spec_with_no_id)
+    runner = CliRunner()
+    runner.invoke(cli.deploy_cli, ['--spec', path])
+    with open(path, 'r') as f:
+        updated_spec = json.loads(f.read())
+    assert 'id' in updated_spec
```
mukulmurthy, line 137: Couple missing tests for not touching the ID if it's valid, and what happens if the id is an empty string.

mukulmurthy: Let's also test for invalid json?

databricks_cli/pipelines/cli.py
```diff
@@ -21,16 +21,26 @@
...
-from json import loads as json_loads
 import os
+import uuid
+import json
+import string
+
+try:
+    from urlparse import urlparse, urljoin
+except ImportError:
+    from urllib.parse import urlparse, urljoin
```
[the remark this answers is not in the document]
null-sleep: It's a pattern used to support python2/3. The library was renamed in py3

databricks_cli/utils.py
null-sleep: this part is replicating `eat_exceptions`. if we want to change this, it should be in a separate PR and we should do it in both places

databricks_cli/pipelines/cli.py
null-sleep: Yes `api_client.url` cleans up the `?o=123` part and just returns the host url

tests/pipelines/test_cli.py
null-sleep: Invalid json input is tested in `_read_spec`. I think validate should raise an error for an empty or invalid id and I'll add a test for the case when an id is provided.

databricks_cli/utils.py
null-sleep: Have a handler that tries to parse for error_code and message and if it fails, raises the Exception?

PR #298, opened 2020-05-26
Pipelines CLI Fixes

null-sleep:
This PR makes the following changes to the Pipelines CLI:

A successful spec deployment prints the URL to the pipeline.
If the provided spec file does not have an ID, one is generated and the spec file is updated with the new ID.
Adds a new exception eater to print cleaner error messages from the pipelines API
Adds exception handing for specs with invalid json
A validation check for pipeline id: only -, _ and alphanumeric characters are allowed
Testing:
Unit Tests
Manually verified deployment URLs, new exception formatting, validation of pipeline ids

Continues from https://github.com/databricks/databricks-cli/pull/296 addressing PR comments, for some reason I was able to push straight to main and repo and can no longer can.

### DIFF
--- a/databricks_cli/pipelines/cli.py
+++ b/databricks_cli/pipelines/cli.py
@@ -74,9 +74,7 @@ def deploy_cli(api_client, spec_arg, spec):
         click.echo("Updating spec at {} with id: {}".format(src, pipeline_id))
         spec_obj['id'] = pipeline_id
         _write_spec(src, spec_obj)
-    msg = _validate_pipeline_id(spec_obj['id'])
-    if msg is not None:
-        error_and_quit(msg)
+    _validate_pipeline_id(spec_obj['id'])
     PipelinesApi(api_client).deploy(spec_obj)
 
     pipeline_id = spec_obj['id']
@@ -215,9 +213,7 @@ def _get_pipeline_id(spec_arg, spec, pipeline_id):
     if bool(spec_arg) or bool(spec):
         src = spec_arg if bool(spec_arg) else spec
         pipeline_id = _read_spec(src)["id"]
-    msg = _validate_pipeline_id(pipeline_id)
-    if msg is not None:
-        error_and_quit(msg)
+    _validate_pipeline_id(pipeline_id)
     return pipeline_id
 
 
@@ -226,11 +222,11 @@ def _validate_pipeline_id(pipeline_id):
     Checks if the pipeline_id only contain -, _ and alphanumeric characters
     """
     if len(pipeline_id) == 0:
-        return u'Empty pipeline id provided'
+        error_and_quit(u'Empty pipeline id provided')
     if not set(pipeline_id) <= PIPELINE_ID_PERMITTED_CHARACTERS:
         message = u'Pipeline id {} has invalid character(s)\n'.format(pipeline_id)
         message += u'Valid characters are: _ - a-z A-Z 0-9'
-        return message
+        error_and_quit(message)
 
 
 @click.group(context_settings=CONTEXT_SETTINGS,
--- a/databricks_cli/pipelines/cli.py
+++ b/databricks_cli/pipelines/cli.py
@@ -117,6 +117,7 @@ def delete_cli(api_client, spec_arg, spec, pipeline_id):
     """
     pipeline_id = _get_pipeline_id(spec_arg=spec_arg, spec=spec, pipeline_id=pipeline_id)
     PipelinesApi(api_client).delete(pipeline_id)
+    click.echo("Pipeline {} successfully deleted".format(pipeline_id))
 
 
 @click.command(context_settings=CONTEXT_SETTINGS,
@@ -178,6 +179,7 @@ def reset_cli(api_client, spec_arg, spec, pipeline_id):
     """
     pipeline_id = _get_pipeline_id(spec_arg=spec_arg, spec=spec, pipeline_id=pipeline_id)
     PipelinesApi(api_client).reset(pipeline_id)
+    click.echo("Pipeline {} successfully reset".format(pipeline_id))
 
 
 def _read_spec(src):
--- a/databricks_cli/pipelines/cli.py
+++ b/databricks_cli/pipelines/cli.py
@@ -24,6 +24,7 @@
 import os
 import uuid
 import json
+import string
 
 try:
     from urlparse import urlparse, urljoin
@@ -38,6 +39,8 @@ from databricks_cli.pipelines.api import PipelinesApi
 from databricks_cli.configure.config import provide_api_client, profile_option, debug_option
 from databricks_cli.utils import pipelines_exception_eater, CONTEXT_SETTINGS, pretty_format, \
     error_and_quit
+
+PIPELINE_ID_PERMITTED_CHARACTERS = set(string.ascii_letters + string.digits + '-_')
 
 
 @click.command(context_settings=CONTEXT_SETTINGS,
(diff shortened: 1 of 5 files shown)


databricks_cli/pipelines/cli.py
```diff
@@ -180,9 +215,24 @@ def _get_pipeline_id(spec_arg, spec, pipeline_id):
...
+def _validate_pipeline_id(pipeline_id):
+    """
+    Checks if the pipeline_id only contain -, _ and alphanumeric characters
+    """
+    if len(pipeline_id) == 0:
+        return 'Empty pipeline id provided'
+    if not set(pipeline_id) <= PIPELINE_ID_PERMITTED_CHARACTERS:
+        message = 'Pipeline id {} has invalid character(s)\n'.format(pipeline_id)
+        message += "Valid characters are: _ - a-z A-Z 0-9"
+        return message
```
null-sleep, line 233: Having validate call error_and_quit makes the code untestable. Click tries to capture the actual error and drops it to raise a Click error for having a context during testing.

databricks_cli/pipelines/cli.py
```diff
@@ -59,8 +69,21 @@ def deploy_cli(api_client, spec_arg, spec):
...
+        _write_spec(src, spec_obj)
+    msg = _validate_pipeline_id(spec_obj['id'])
+    if msg is not None:
+        error_and_quit(msg)
     PipelinesApi(api_client).deploy(spec_obj)
 
+    pipeline_id = spec_obj['id']
+    base_url = "{0.scheme}://{0.netloc}/".format(urlparse(api_client.url))
+    pipeline_url = urljoin(base_url, "#joblist/pipelines/{}".format(pipeline_id))
+    click.echo("Pipeline successfully deployed: {}".format(pipeline_url))
```
null-sleep, line 85: The URL will have the pipeline id

databricks_cli/pipelines/cli.py
mukulmurthy: That makes this API a lot more awkward to use. Can the tests set up a context so the real error gets surfaced?

tests/pipelines/test_cli.py
```diff
@@ -124,6 +124,31 @@ def test_delete_cli_incorrect_parameters(pipelines_api_mock, tmpdir):
     assert pipelines_api_mock.delete.call_count == 0
 
 
+@provide_conf
+def test_deploy_spec_updated_with_id_if_pipeline_id_not_in_spec(pipelines_api_mock, tmpdir):  # noqa
```
mukulmurthy, line 128: what is noqa?

databricks_cli/pipelines/cli.py
null-sleep: I spent some time trying to do the same, I'll have a look again

tests/pipelines/test_cli.py
null-sleep: It tells the linter to ignore the line. Removing, no longer needed.

databricks_cli/utils.py
```diff
@@ -58,7 +88,7 @@ def error_and_quit(message):
     context_object = ctx.ensure_object(ContextObject)
     if context_object.debug_mode:
         traceback.print_exc()
-    click.echo('Error: {}'.format(message))
+    click.echo(u'Error: {}'.format(message))
```
null-sleep, line 91: Python 2's default encoding is ascii, not unicode causing any error message with unicode to raise an `UnicodeEncodeError` exception

databricks_cli/pipelines/cli.py
```diff
@@ -59,8 +69,19 @@ def deploy_cli(api_client, spec_arg, spec):
         raise RuntimeError('The spec should be provided either by an option or argument')
     src = spec_arg if bool(spec_arg) else spec
     spec_obj = _read_spec(src)
+    if 'id' not in spec_obj:
+        pipeline_id = str(uuid.uuid4())
+        click.echo("Updating spec at {} with id: {}".format(src, pipeline_id))
+        spec_obj['id'] = pipeline_id
+        _write_spec(src, spec_obj)
+    _validate_pipeline_id(spec_obj['id'])
```
mukulmurthy, line 77: Why are we validating for deploy and get, but not reset and delete?

databricks_cli/pipelines/cli.py
```diff
@@ -59,8 +69,19 @@ def deploy_cli(api_client, spec_arg, spec):
...
     PipelinesApi(api_client).deploy(spec_obj)
 
+    pipeline_id = spec_obj['id']
+    base_url = "{0.scheme}://{0.netloc}/".format(urlparse(api_client.url))
+    pipeline_url = urljoin(base_url, "#joblist/pipelines/{}".format(pipeline_id))
+    click.echo("Pipeline successfully deployed: {}".format(pipeline_url))
```
mukulmurthy, line 83: Can we make sure there are nice messages returned for delete and reset as well?

databricks_cli/utils.py
```diff
@@ -46,9 +46,39 @@ def decorator(*args, **kwargs):
...
+            return function(*args, **kwargs)
+        except HTTPError as exception:  # noqa
+            if exception.response.status_code == 401:
+                error_and_quit('Your authentication information may be incorrect. Please '
+                               + 'reconfigure with ``dbfs configure``')
+
+            elif exception.response.status_code == 400:
+                exp_context = json_loads(exception.response.content.decode('utf-8'))
+                message = exp_context['error_code'] + '\n' + exp_context['message']
+                error_and_quit(message)
```
mukulmurthy, line 75: Idr if I mentioned this earlier, but let's check for error_code and message being there even if it's not a 400? If they're not, we can always fall back to the other (generic) code path.

databricks_cli/utils.py
```diff
@@ -46,9 +46,39 @@ def decorator(*args, **kwargs):
...
-        except Exception as exception: # noqa
+        except Exception as exception:  # noqa
             if not DEBUG_MODE:
                 error_and_quit('{}: {}'.format(type(exception).__name__, str(exception)))
+
+    decorator.__doc__ = function.__doc__
+    return decorator
+
+
+def pipelines_exception_eater(function):
```
mukulmurthy, line 57: Is it not possible to unit test this by mocking the client?

databricks_cli/pipelines/cli.py
null-sleep: all other methods except deploy call `_get_pipeline_id` which calls `_validate_pipeline_id`. So only deploy needs to call validate explicitly, all commands are covered

databricks_cli/pipelines/cli.py
```diff
@@ -91,6 +117,7 @@ def delete_cli(api_client, spec_arg, spec, pipeline_id):
     """
     pipeline_id = _get_pipeline_id(spec_arg=spec_arg, spec=spec, pipeline_id=pipeline_id)
     PipelinesApi(api_client).delete(pipeline_id)
+    click.echo("Pipeline {} successfully deleted".format(pipeline_id))
```
mukulmurthy, line 120: Nit: Submitting a delete requests deletion of resources, so maybe just say "Pipeline {} deleted". It very rarely fails, and we do retry (even if the service starts up), but probably still best to not say it's successful ;).

databricks_cli/pipelines/cli.py
```diff
@@ -152,6 +179,7 @@ def reset_cli(api_client, spec_arg, spec, pipeline_id):
     """
     pipeline_id = _get_pipeline_id(spec_arg=spec_arg, spec=spec, pipeline_id=pipeline_id)
     PipelinesApi(api_client).reset(pipeline_id)
+    click.echo("Pipeline {} successfully reset".format(pipeline_id))
```
mukulmurthy, line 182: Same thing here, maybe say something like "Reset triggered for pipeline" .

PR #299, opened 2020-05-27
Add/Fix support for maven deps in pipelines spec

null-sleep:
Previously a pipelines spec with maven dependencies would fail with the error: `Error: AttributeError: 'dict' object has no attribute 'decode'`. This PR fixes the bugs.

Testing:
Updated tests and manual testing

### DIFF
--- a/databricks_cli/pipelines/api.py
+++ b/databricks_cli/pipelines/api.py
@@ -33,7 +33,7 @@ from databricks_cli.dbfs.dbfs_path import DbfsPath
 
 BUFFER_SIZE = 1024 * 64
 base_pipelines_dir = 'dbfs:/pipelines/code'
-supported_lib_types = {'jar', 'whl'}
+supported_lib_types = {'jar', 'whl', 'maven'}
 
 
 class PipelinesApi(object):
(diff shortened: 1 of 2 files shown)


databricks_cli/pipelines/api.py
```diff
@@ -33,7 +33,7 @@
 
 BUFFER_SIZE = 1024 * 64
 base_pipelines_dir = 'dbfs:/pipelines/code'
-supported_lib_types = {'jar', 'whl'}
+supported_lib_types = {'jar', 'whl', 'maven'}
```
null-sleep, line 36: add `egg` here as well?

null-sleep:
@mukulmurthy

tests/pipelines/test_api.py
```diff
@@ -225,7 +231,8 @@ def test_library_object_serialization_deserialization():
         LibraryObject('egg', 'FiLe:/weird/case.ext'),
         LibraryObject('whl', 'file.ext'),
         LibraryObject('whl', 's3:/s3/path/file.ext'),
-        LibraryObject('jar', 'dbfs:/dbfs/path/file.ext')
+        LibraryObject('jar', 'dbfs:/dbfs/path/file.ext'),
+        LibraryObject('maven', {'coordinates': 'com.package.name'})
```
mukulmurthy, line 235: Can we also add an exclusion for this test since it's a serialization test, and that thing's a list?

databricks_cli/pipelines/api.py
mukulmurthy: When we did python before we had a good reason for not supporting eggs. I don't remember exactly what that was, but we can do it in the future when it's needed.

PR #303, opened 2020-06-30
Cluster policies CLI

itaiw:
Hey Andrew,

I added the policies, my tests fail and I think this is caused by Unicode response.

Let me know if you can help.

Thanks,
Itai


itaiw:
Hi Andrew,

I fixed all of those, and I passed the tox tests
There's a tiny change in databricks_cli/cli.py I added the cluster policy, but it insists on removing a line and adding the same one again.

I hope this passes.
Thanks,
Itai

itaiw:
I modified the services api file to match the on generated by Bazel.
Here is the Universe PR for reference https://github.com/databricks/universe/pull/61883

databricks_cli/cli.py
```diff
@@ -63,4 +65,4 @@ def cli():
 cli.add_command(pipelines_group, name='pipelines')
 
 if __name__ == "__main__":
-    cli()
+    cli()
```
andrewmchen, line 68: Could we add back the newline here?

databricks_cli/cli.py
```diff
@@ -54,6 +55,7 @@ def cli():
 cli.add_command(workspace_group, name='workspace')
 cli.add_command(jobs_group, name='jobs')
 cli.add_command(clusters_group, name='clusters')
+cli.add_command(cluster_policies_group, name='cluster_policies')
```
andrewmchen, line 58: can we make this `cluster-policies`? That would be consistent with `instance-pools`.

databricks_cli/cluster_policies/cli.py
```diff
@@ -0,0 +1,164 @@
...
+
+    Returns information about all currently cluster policiess.
+
+    By default the output format will be a human readable table with the following fields
+
+      - Policy ID
+
+      - Policy name
+
+      - Policy state
```
andrewmchen, line 136: I think this should be `Policy definition` to be consistent with the API naming.

databricks_cli/cluster_policies/cli.py
```diff
@@ -0,0 +1,164 @@
...
+
+def _cluster_policies_to_table(policies_json):
+    ret = []
+    for c in policies_json.get('policies', []):
+        ret.append((c['policy_id'], truncate_string(c['name']), c['definition']))
+    return ret
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Lists Cluster Policiess.')
```
andrewmchen, line 118: Can we run a quick `s/Policiess/Policies/g` on this file?

databricks_cli/cli.py
itaiw: Done

databricks_cli/cli.py
itaiw: Done

databricks_cli/cluster_policies/cli.py
itaiw: fixed. Thanks

databricks_cli/cluster_policies/cli.py
itaiw: fixed. Thanks

PR #308, opened 2020-07-29
Use create pipeline API

rgovind3:
Previously, we used the same API to both create and update a pipeline. The apis have been split, and this PR updates the cli to use the new create API for pipelines.

Testing:
- Manual testing
- Unit tests

### DIFF
--- a/databricks_cli/pipelines/api.py
+++ b/databricks_cli/pipelines/api.py
@@ -21,9 +21,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 from hashlib import sha1
 import os
+import copy
 
 from six.moves import urllib
 
@@ -41,13 +41,13 @@ class PipelinesApi(object):
         self.client = DeltaPipelinesService(api_client)
         self.dbfs_client = DbfsApi(api_client)
 
-    def deploy(self, spec, headers=None):
-        lib_objects = LibraryObject.from_json(spec.get('libraries', []))
-        local_lib_objects, external_lib_objects = \
-            self._identify_local_libraries(lib_objects)
+    def create(self, spec, headers=None):
+        spec = self._upload_libraries_and_update_spec(spec)
+        return self.client.client.perform_query('POST', '/pipelines/'.format(), data=spec,
+                                                headers=headers)
 
-        spec['libraries'] = LibraryObject.to_json(external_lib_objects +
-                                                  self._upload_local_libraries(local_lib_objects))
+    def deploy(self, spec, headers=None):
+        spec = self._upload_libraries_and_update_spec(spec)
         pipeline_id = spec['id']
         self.client.client.perform_query('PUT', '/pipelines/{}'.format(pipeline_id), data=spec,
                                          headers=headers)
@@ -60,6 +60,16 @@ class PipelinesApi(object):
 
     def reset(self, pipeline_id, headers=None):
         self.client.reset(pipeline_id, headers)
+
+    def _upload_libraries_and_update_spec(self, spec):
+        spec = copy.deepcopy(spec)
+        lib_objects = LibraryObject.from_json(spec.get('libraries', []))
+        local_lib_objects, external_lib_objects = \
+            self._identify_local_libraries(lib_objects)
+
+        spec['libraries'] = LibraryObject.to_json(external_lib_objects +
+                                                  self._upload_local_libraries(local_lib_objects))
+        return spec
 
     @staticmethod
     def _identify_local_libraries(lib_objects):
--- a/databricks_cli/pipelines/api.py
+++ b/databricks_cli/pipelines/api.py
@@ -51,6 +51,9 @@ class PipelinesApi(object):
         self.client.client.perform_query('PUT', '/pipelines/{}'.format(pipeline_id), data=spec,
                                          headers=headers)
 
+    def list(self, headers=None):
+        return self.client.client.perform_query("GET", "/pipelines", headers=headers)
+
     def delete(self, pipeline_id, headers=None):
         self.client.delete(pipeline_id, headers)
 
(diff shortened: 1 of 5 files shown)


databricks_cli/pipelines/api.py
```diff
@@ -61,6 +60,15 @@ def get(self, pipeline_id, headers=None):
     def reset(self, pipeline_id, headers=None):
         self.client.reset(pipeline_id, headers)
 
+    def _upload_libraries_and_update_spec(self, spec):
+        spec = copy.deepcopy(spec)
```
mukulmurthy, line 64: Why'd you add this thing?

databricks_cli/pipelines/cli.py
```diff
@@ -70,22 +70,25 @@ def deploy_cli(api_client, spec_arg, spec):
...
+        click.echo("Successfully deployed pipeline: {}".format(
+            _get_pipeline_url(api_client, new_pipeline_id)))
 
-    pipeline_id = spec_obj['id']
-    base_url = "{0.scheme}://{0.netloc}/".format(urlparse(api_client.url))
-    pipeline_url = urljoin(base_url, "#joblist/pipelines/{}".format(pipeline_id))
-    click.echo("Pipeline successfully deployed: {}".format(pipeline_url))
+        spec_obj['id'] = new_pipeline_id
+        _write_spec(src, spec_obj)
+        click.echo("Updated spec at {} with ID {}".format(src, new_pipeline_id))
```
mukulmurthy, line 86: @anew, where did we land on if we're going to update the spec in place or not? In a Rahul's implementation, we always do, which is fine with me.

databricks_cli/pipelines/api.py
rgovind3: The deepcopy makes sure that `_upload_libraries_and_update_spec` is a pure function. It's not really necessary here but it make sure that the spec passed through the arguments is not modified.

databricks_cli/pipelines/api.py
```diff
@@ -41,17 +41,19 @@ def __init__(self, api_client):
...
-        spec['libraries'] = LibraryObject.to_json(external_lib_objects +
-                                                  self._upload_local_libraries(local_lib_objects))
+    def deploy(self, spec, headers=None):
+        spec = self._upload_libraries_and_update_spec(spec)
         pipeline_id = spec['id']
         self.client.client.perform_query('PUT', '/pipelines/{}'.format(pipeline_id), data=spec,
                                          headers=headers)
 
+    def list(self, headers=None):
+        return self.client.client.perform_query("GET", "/pipelines", headers=headers)
```
mukulmurthy, line 55: I thought we can't do this without pagination since it's a breaking API change? It feels wrong to put all this name validation logic in the CLI; it should be done on the deployment service. The CLI always tries to deploy a pipeline (optionally supplying the force or ignore_name_collision arg), and the deployment service decides whether or not to throw.

databricks_cli/pipelines/cli.py
```diff
@@ -52,40 +52,65 @@
                short_help='Deploys a delta pipeline according to the pipeline specification')
 @click.argument('spec_arg', default=None, required=False)
 @click.option('--spec', default=None, type=PipelineSpecClickType(), help=PipelineSpecClickType.help)
+@click.option('--allow-duplicate-names', is_flag=True,
+              help="Skip duplicate name check while deploying pipeline")
+@click.option('--no-update-spec', is_flag=True,
+              help="Do not update spec with pipeline ID after creating a new pipeline.")
```
mukulmurthy, line 58: Nit: spec file

databricks_cli/pipelines/cli.py
```diff
@@ -52,40 +52,65 @@
...
-def deploy_cli(api_client, spec_arg, spec):
+def deploy_cli(api_client, spec_arg, spec, allow_duplicate_names, no_update_spec):
     """
     Deploys a delta pipeline according to the pipeline specification. The pipeline spec is a
     specification that explains how to run a Delta Pipeline on Databricks. All local libraries
     referenced in the spec are uploaded to DBFS.
 
+    The deploy command creates a new pipeline and adds the ID of this pipeline to your spec if your
+    spec does not already contain an ID. The spec will not be updated if the --no-update-spec
+    option is added.
```
mukulmurthy, line 71: Minor: To make this a little clearer:
```
If the pipeline spec contains an "id" field, attempts to update an existing pipeline with that ID. If it does not, creates a new pipeline and edits the spec to add the ID of the created pipeline. The spec will not be updated if the --no-update-spec option is added.

The deploy command will not create a new pipeline if a pipeline with the same name already exists. You can disable this check by using the --allow-duplicate-names option.
```

databricks_cli/pipelines/cli.py
```diff
@@ -236,6 +266,13 @@ def _validate_pipeline_id(pipeline_id):
         error_and_quit(message)
 
 
+def _handle_duplicate_name_exception(spec, exception):
+    if json.loads(exception.response.text).get('error_code', "") == 'RESOURCE_CONFLICT':
```
mukulmurthy, line 270: Is this safe to do? It's JSON if it's a DatabricksServiceException, but I'm not sure it's guaranteed to be?

tests/pipelines/test_api.py
```diff
@@ -105,36 +126,70 @@ def test_deploy(put_file_mock, dbfs_path_validate, pipelines_api, tmpdir):
...
+    spec = copy.deepcopy(SPEC_WITHOUT_ID)
+    spec['libraries'] = []
+
+    pipelines_api.create(spec, False)
+    data = copy.deepcopy(spec)
+    data['allow_duplicate_names'] = False
+    client_mock.assert_called_with("POST", "/pipelines", data=data, headers=None)
+    assert client_mock.call_count == 1
+
+    pipelines_api.create(spec, True, HEADERS)
```
mukulmurthy, line 170: Nit: Named parameters , for the bools at least

databricks_cli/pipelines/cli.py
rgovind3: That's true. Added some additional error handling to account for this.

databricks_cli/pipelines/cli.py
rgovind3: Thanks! This is definitely more clear. Updated the comment.

PR #317, opened 2020-09-03
Update Pipelines CLI to take a pipeline-id parameter + stop overwriting spec with new ID + remove the spec argument from all commands except deploy.

aakash-db:
This PR adds a pipeline-id parameter to the deploy command, and removes automatic rewriting of the spec when a new pipeline is created. It also removes the spec parameter from all commands except for deploy, only taking an explicitly specified --pipeline_id argument instead. This will create a more intuitive CLI experience for pipeline users, now that all id's are service-generated (rather than user-generated), as pipeline_id, not a spec, becomes the core identifying parameter across pipelines commands.

Testing
* Unit tests
* Manual testing

### DIFF
--- a/databricks_cli/pipelines/cli.py
+++ b/databricks_cli/pipelines/cli.py
@@ -100,8 +100,8 @@ def deploy_cli(api_client, spec_arg, spec, allow_duplicate_names, pipeline_id):
     else:
         if (pipeline_id and 'id' in spec_obj) and pipeline_id != spec_obj["id"]:
             raise ValueError(
-                "The ID provided from --pipeline_id '{}' is different than the id provided the spec '{}'. Please "
-                "resolve the conflict and try the command again.")
+                "The ID provided in --pipeline_id '{}' is different from the id provided the spec '{}'. Please "
+                "resolve the conflict and try the command again.".format(pipeline_id, spec["id"]))
 
         spec_obj['id'] = pipeline_id or spec_obj.get('id', None)
         _validate_pipeline_id(spec_obj['id'])
--- a/databricks_cli/pipelines/cli.py
+++ b/databricks_cli/pipelines/cli.py
@@ -66,9 +66,11 @@ def deploy_cli(api_client, spec_arg, spec, allow_duplicate_names, pipeline_id):
     specification that explains how to run a Delta Pipeline on Databricks. All local libraries
     referenced in the spec are uploaded to DBFS.
 
-    If the pipeline spec contains an "id" field, attempts to update an existing pipeline with
-    that ID. If it does not, creates a new pipeline and edits the spec file to add the ID of the
-    created pipeline. The spec file will not be updated if the --no-update-spec option is added.
+    If the pipeline spec contains an "id" field, or if a pipeline id is specified directly
+    (using the  --pipeline-id argument), attempts to update an existing pipeline
+    with that ID. If it does not, creates a new pipeline and logs the id of the new pipeline
+    to STDOUT. Note that if an id is both specified in the spec and passed via --pipeline-id,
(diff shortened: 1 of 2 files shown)


aakash-db:
@marmbrus  @mukulmurthy @rgovind3 Please take a look!

databricks_cli/pipelines/cli.py
```diff
@@ -95,14 +95,19 @@ def deploy_cli(api_client, spec_arg, spec, allow_duplicate_names, no_update_spec
...
             _get_pipeline_url(api_client, new_pipeline_id)))
 
-        if not no_update_spec:
-            spec_obj['id'] = new_pipeline_id
-            _write_spec(src, spec_obj)
-            click.echo("Updated spec at {} with ID {}".format(src, new_pipeline_id))
-        else:
-            click.echo("Pipeline has been assigned ID {}".format(new_pipeline_id))
+        click.echo("Pipeline has been assigned ID {}".format(new_pipeline_id))
+        click.echo(new_pipeline_id, err=True)
```
mukulmurthy, line 99: What's this second line for?

mukulmurthy: Also, can we swap that "assigned ID" line to be before the link?

databricks_cli/pipelines/cli.py
```diff
@@ -95,14 +95,19 @@ def deploy_cli(api_client, spec_arg, spec, allow_duplicate_names, no_update_spec
...
     else:
+        if (pipeline_id and 'id' in spec_obj) and pipeline_id != spec_obj["id"]:
+            raise ValueError(
+                "The ID provided in --pipeline_id '{}' is different from the id provided the "
+                "spec '{}'. Please resolve the conflict and try the command again."
+                .format(pipeline_id, spec["id"])
+            )
```
mukulmurthy, line 106: I'd say something like "Because pipeline IDs are no longer persisted after being deleted, we recommend removing the ID field from your spec."

databricks_cli/pipelines/cli.py
aakash-db: > What's this second line for?

This was the thing where we log only the pipeline_id to STDERR so people can pipe it more easily. 

> Also, can we swap that "assigned ID" line to be before the link?

What do you mean by this?

mukulmurthy: Oh, just moving the new logging you added to be before the previous one.

aakash-db: Done.

databricks_cli/pipelines/cli.py
aakash-db: Done.

databricks_cli/pipelines/cli.py
```diff
@@ -80,29 +80,36 @@ def deploy_cli(api_client, spec_arg, spec, allow_duplicate_names, no_update_spec
     OR
```
mukulmurthy, line 80: Can you fix the comment above to reflect the new behavior?

databricks_cli/pipelines/cli.py
mukulmurthy: Sorry, meant to add that in addition to what you had before. It's still good to tell the user they've passed conflicting id params.

mukulmurthy (approved):
Two minor comments; I can merge it for you after that.

databricks_cli/pipelines/cli.py
aakash-db: Done.

databricks_cli/pipelines/cli.py
aakash-db: Ah, understood. Updated.

PR #323, opened 2020-09-16
[SC-49206] Add run and stop commands to the cli

calving-db:
Changes:
Adds run/stop commands which underneath just call the REST API.

Tested:
Ran `tox` tests.

### DIFF
--- a/databricks_cli/pipelines/api.py
+++ b/databricks_cli/pipelines/api.py
@@ -63,6 +63,12 @@ class PipelinesApi(object):
     def reset(self, pipeline_id, headers=None):
         self.client.reset(pipeline_id, headers)
 
+    def run(self, pipeline_id, headers=None):
+        self.client.run(pipeline_id, headers)
+
+    def stop(self, pipeline_id, headers=None):
+        self.client.stop(pipeline_id, headers)
+
     def _upload_libraries_and_update_spec(self, spec):
         spec = copy.deepcopy(spec)
         lib_objects = LibraryObject.from_json(spec.get('libraries', []))
--- a/databricks_cli/pipelines/cli.py
+++ b/databricks_cli/pipelines/cli.py
@@ -204,6 +204,7 @@ def run_cli(api_client, pipeline_id):
     databricks pipelines run --pipeline-id 1234
     """
     _validate_pipeline_id(pipeline_id)
+        PipelinesApi(api_client).run(pipeline_id)
     click.echo("Run triggered for pipeline {}".format(pipeline_id))
 
 
--- a/databricks_cli/pipelines/cli.py
+++ b/databricks_cli/pipelines/cli.py
@@ -205,6 +205,69 @@ def reset_cli(api_client, spec_arg, spec, pipeline_id):
     click.echo("Reset triggered for pipeline {}".format(pipeline_id))
 
 
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Starts the execution of a delta pipeline run')
+@click.argument('spec_arg', default=None, required=False)
+@click.option('--spec', default=None, type=PipelineSpecClickType(), help=PipelineSpecClickType.help)
+@click.option('--pipeline-id', default=None, type=PipelineIdClickType(),
+              help=PipelineIdClickType.help)
+@debug_option
+@profile_option
+@pipelines_exception_eater
+@provide_api_client
+def run_cli(api_client, spec_arg, spec, pipeline_id):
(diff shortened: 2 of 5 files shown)


mukulmurthy (commented):
We're removing the ability to pass anything but a --pipeline-id to non-deploy commands (see https://github.com/databricks/databricks-cli/pull/317). Guessing that one will go in first, so might have to hold off on this for a second and build on top of that.

databricks_cli/pipelines/cli.py
```diff
@@ -187,6 +187,49 @@ def reset_cli(api_client, pipeline_id):
...
+    """
+    Starts the execution of a delta pipelines run by starting the cluster and processing data.
+
+    Usage:
+
+    databricks pipelines run --pipeline-id 1234
+    """
+    _validate_pipeline_id(pipeline_id)
+        PipelinesApi(api_client).run(pipeline_id)
+    click.echo("Run triggered for pipeline {}".format(pipeline_id))
```
mukulmurthy, line 208: Fix the indenting here?

I don't know Python well enough, but I thought this was an error.

databricks_cli/pipelines/cli.py
```diff
@@ -187,6 +187,49 @@ def reset_cli(api_client, pipeline_id):
...
+    Stops the execution of a delta pipelines run by terminating the cluster. Processing of data can be resumed by
+    calling `run`.
+
+    Usage:
+
+    databricks pipelines stop --pipeline-id 1234
+    """
+    _validate_pipeline_id(pipeline_id)
+    PipelinesApi(api_client).stop(pipeline_id)
+    click.echo("Stop triggered for pipeline {}".format(pipeline_id))
```
mukulmurthy, line 230: Nit: "Stopped pipeline"

databricks_cli/pipelines/api.py
```diff
@@ -63,6 +63,12 @@ def get(self, pipeline_id, headers=None):
     def reset(self, pipeline_id, headers=None):
         self.client.reset(pipeline_id, headers)
 
+    def run(self, pipeline_id, headers=None):
+        self.client.run(pipeline_id, headers)
+
+    def stop(self, pipeline_id, headers=None):
```
mukulmurthy, line 69: I know this is pretty basic, but we should still add tests for it in test_api.py.

databricks_cli/sdk/service.py
```diff
@@ -938,3 +938,13 @@ def reset(self, pipeline_id=None, headers=None):
         _data = {}
 
         return self.client.perform_query('POST', '/pipelines/{pipeline_id}/reset'.format(pipeline_id=pipeline_id), data=_data, headers=headers)
+
+    def stop(self, pipeline_id=None, headers=None):
+        _data = {}
+
+        return self.client.perform_query('POST', '/pipelines/{pipeline_id}/stop'.format(pipeline_id=pipeline_id), data=_data, headers=headers)
+
+    def run(self, pipeline_id=None, headers=None):
```
mukulmurthy, line 947: Are these generated from the protos or did you do it manually?

calving-db: Oh, makes sense why the code looks so weird. Didn't realize they were generated. Do you happen to know the command to do that?

databricks_cli/pipelines/cli.py
calving-db: Yeah that is strange, all the tests passed though. Going to look into it.

databricks_cli/sdk/service.py
mukulmurthy: `bb //api:doc_and_sdk`. In theory you just copy the service.py file that gets generated, but they're out of sync, so just grab the pipelines bits that get generated. I think it's probably going to be exactly what you have.

databricks_cli/pipelines/cli.py
mukulmurthy: Actually it looks like the build on this PR did fail

calving-db: Yeah you're right, I think i just messed up when running tox then

databricks_cli/sdk/service.py
calving-db: Nice, will confirm that it's the same

calving-db:
Sorry for the million commit emails, something between virtualenv/tox was really not cooperating on my machine and I didn't anticipate having so many bugs lol

mukulmurthy (approved):

PR #325, opened 2020-09-22
[PLAT-21956] Add az-token config functionality to db cli

sushi1998:
Added a new option to the databricks-cli wherein users can run ```databricks-cli configure --az-token``` and add an az-token to the configuration. Environment variables with the bearer token (DATABRICKS_TOKEN) and the az-token (DATABRICKS_AZ_TOKEN) need to be set.


databricks_cli/configure/cli.py
```diff
@@ -31,6 +32,8 @@
 from databricks_cli.configure.config import profile_option, get_profile_from_context, debug_option
 
 PROMPT_HOST = 'Databricks Host (should begin with https://)'
+PROMPT_RESOURCE_ID = 'Resource/Workspace ID'
+PROMPT_AZ_TOKEN = 'Azure Token'
```
mglee72, line 36: nit: This field doesn't seem to be in use anymore.

databricks_cli/configure/cli.py
```diff
@@ -44,6 +47,16 @@ def _configure_cli_token(profile, insecure):
     update_and_persist_config(profile, new_config)
 
 
+def _configure_cli_aad_token(profile, insecure):
+    config = ProfileConfigProvider(profile).get_config() or DatabricksConfig.empty()
+    host = click.prompt(PROMPT_HOST, default=config.host, type=_DbfsHost())
+    token = os.environ.get('DATABRICKS_TOKEN')
+    az_token = os.environ.get('DATABRICKS_AZ_TOKEN')
```
mglee72, line 54: Do we have any guard against missing/invalid tokens? Seems like if users do are not guided to use the envvar approach yet, and if they don't write the variable correctly, the CLI would just silently fail by passing over the AZ token login option.

Also, it would be worth it to actually check-in with the PMs that env-var based approach would be acceptable.

databricks_cli/sdk/service.py
```diff
@@ -616,15 +616,22 @@ class SecretService(object):
...
         if scope is not None:
             _data['scope'] = scope
         if initial_manage_principal is not None:
             _data['initial_manage_principal'] = initial_manage_principal
         if scope_backend_type is not None:
-            _data['scope_backend_type'] = scope_backend_type
+            if scope_backend_type == 'databricks':
+                _data['scope_backend_type_str'] = 1
+            elif azure_keyvault is not None and scope_backend_type == 'azure_keyvault':
+                _data['scope_backend_type'] = 2
```
mglee72, line 630: Just making sure, `scope_backend_type` and `scope_backend_type_str` differ by design right?

databricks_cli/secrets/cli.py
```diff
@@ -50,15 +50,29 @@
               ' principal for this option is the group "users", which contains all users in the'
               ' workspace. If not specified, the initial ACL with MANAGE permission applied to the'
               ' scope is assigned to the request issuer\'s user identity.')
+@click.option('--scope-backend-type', type=click.Choice(['azure_keyvault', 'databricks'], case_sensitive=True),
```
mglee72, line 53: It would be important to clarify that one must log in with AZ tokens if the person desires to create an AKV backed scope. Also, I think we must print a user-friendly error message when an user attempts to create an AKV backed scope while not logged in with AZ tokens.

mglee72 (changes requested):

databricks_cli/configure/cli.py
```diff
@@ -44,6 +52,27 @@ def _configure_cli_token(profile, insecure):
     update_and_persist_config(profile, new_config)
 
 
+def _configure_cli_aad_token(profile, insecure):
+    config = ProfileConfigProvider(profile).get_config() or DatabricksConfig.empty()
+    host = click.prompt(PROMPT_HOST, default=config.host, type=_DbfsHost())
+
+    is_token_env_set = click.prompt(PROMPT_ENV_TOKEN, type=bool)
```
andrewmchen, line 59: Is prompting in this function twice helpful? Can't we just check the `ENV_TOKEN` and `ENV_AZ_TOKEN` keys in environ without the prompt and save the user a couple of keystrokes?

databricks_cli/sdk/service.py
andrewmchen: Can we put in the strings `DATABRICKS` and `AZURE_KEYVAULT` in the _data[`scope_backend_type`] field directly? Our rpc system should be able to handle the string constants as part of the enum I think.

Also we should not make these changes in this file since it is autogenerated. Could you lift them to `databricks_cli/secrets/api.py` and regenerate this file to include the `backend_azure_keyvault` argument?

databricks_cli/configure/cli.py
sushi1998: `PROMPT_AZ_TOKEN` is not being used anymore - removed it

databricks_cli/configure/cli.py
sushi1998: The CLI itself doesn't have any test against invalid tokens - it never did. Talked with PMs about env variable approach and they said it was acceptable

databricks_cli/configure/cli.py
```diff
@@ -31,19 +32,39 @@
 from databricks_cli.configure.config import profile_option, get_profile_from_context, debug_option
 
 PROMPT_HOST = 'Databricks Host (should begin with https://)'
+PROMPT_RESOURCE_ID = 'Resource/Workspace ID'
 PROMPT_USERNAME = 'Username'
 PROMPT_PASSWORD = 'Password' #  NOQA
 PROMPT_TOKEN = 'Token' #  NOQA
+ENV_AAD_TOKEN = 'DATABRICKS_TOKEN'
```
andrewmchen, line 39: This seems like a bug?

databricks_cli/configure/cli.py
```diff
@@ -31,19 +32,39 @@
 from databricks_cli.configure.config import profile_option, get_profile_from_context, debug_option
 
 PROMPT_HOST = 'Databricks Host (should begin with https://)'
+PROMPT_RESOURCE_ID = 'Resource/Workspace ID'
```
andrewmchen, line 35: Remove please.

databricks_cli/sdk/__init__.py
```diff
@@ -24,7 +24,7 @@
 """
 Databricks Python REST Client 2.0 for interacting with various services.
 
-Currently supports services including clusters, clusters policies and jobs.
+Currently supports services including clusters and jobs.
```
andrewmchen, line 27: Let's revert changes to this file.

databricks_cli/sdk/api_client.py
```diff
@@ -113,10 +113,10 @@ def perform_query(self, method, path, data = {}, headers = None):
             if method == 'GET':
                 translated_data = {k: _translate_boolean_to_query_param(data[k]) for k in data}
                 resp = self.session.request(method, self.url + path, params = translated_data,
```
andrewmchen, line 115: Is it possible to revert all changes to this file?

databricks_cli/configure/cli.py
sushi1998: Removed

databricks_cli/sdk/__init__.py
sushi1998: Revered back to old file

databricks_cli/sdk/api_client.py
sushi1998: Reverted back to old file

databricks_cli/configure/cli.py
sushi1998: What's the bug here?

databricks_cli/configure/cli.py
sushi1998: Removed this and changed workflow

databricks_cli/secrets/cli.py
sushi1998: Added more detailed description saying that users will need AAD token when creating an AKV backed scope - but I'm not aware of any tests to check if users have configured an AAD token aside from maybe the size of the token (compared to the DB PAT). But that seems like a flimsy check, do you know of any others?

databricks_cli/configure/cli.py
andrewmchen: The name of the env var is `DATABRICKS_TOKEN` that's a bit confusing if we're asking for the AAD token right?

databricks_cli/secrets/cli.py
```diff
@@ -50,15 +50,31 @@
               ' principal for this option is the group "users", which contains all users in the'
               ' workspace. If not specified, the initial ACL with MANAGE permission applied to the'
               ' scope is assigned to the request issuer\'s user identity.')
+@click.option('--scope-backend-type', type=click.Choice(['AZURE_KEYVAULT', 'DATABRICKS'], case_sensitive=True),
+              default='DATABRICKS', help='The backend that will be used for this secret scope. '
+                                         'Options are (case-sensitive): 1) \'azure_keyvault\' and 2) \'databricks\' '
```
andrewmchen, line 55: This is very confusing, the message which shows up from help is this
```
  --scope-backend-type [AZURE_KEYVAULT|DATABRICKS]
                                  The backend that will be used for this
                                  secret scope. Options are (case-sensitive):
                                  1) 'azure_keyvault' and 2) 'databricks'
                                  (default option) Note: To create an Azure
                                  Keyvault, be sure to configure an AAD Token
                                  using'databricks-cli configure --aad-token'
```
Which one is it? Lower case or upper case?

databricks_cli/secrets/cli.py
```diff
@@ -50,15 +50,31 @@
...
+                                         '(default option)'
+                                         '\nNote: To create an Azure Keyvault, be sure to configure an AAD Token using'
```
andrewmchen, line 57: missing trailing space here!

databricks_cli/secrets/cli.py
sushi1998: It is uppercase - changed the message

databricks_cli/configure/cli.py
sushi1998: You're right. I thought about this and for the users, to make it as less confusing as possible, I think the var should be DATABRICKS_AAD_TOKEN even though we deal with it in the same as a DATABRICKS TOKEN in the backend. I've changed the variables names to make it consistent with it as well.

databricks_cli/secrets/cli.py
sushi1998: Addressed

databricks_cli/sdk/service.py
```diff
@@ -112,7 +112,7 @@ def submit_run(self, run_name=None, existing_cluster_id=None, new_cluster=None,
         if timeout_seconds is not None:
             _data['timeout_seconds'] = timeout_seconds
         return self.client.perform_query('POST', '/jobs/runs/submit', data=_data, headers=headers)
-
+
```
mglee72, line 115: Seems like we are adding spaces. Is it desired?

databricks_cli/secrets/cli.py
```diff
@@ -316,7 +336,7 @@ def get_acl(api_client, scope, principal, output):
 @debug_option
 @profile_option
 @eat_exceptions
-def secrets_group():
+def secrets_group():  # pragma: no cover
```
mglee72, line 339: is it intended?

databricks_cli/sdk/service.py
```diff
@@ -510,51 +529,105 @@ def put(self, path, contents=None, overwrite=None, headers=None):
...
     def create(self, path, overwrite=None, headers=None):
         _data = {}
         if path is not None:
             _data['path'] = path
         if overwrite is not None:
             _data['overwrite'] = overwrite
         return self.client.perform_query('POST', '/dbfs/create', data=_data, headers=headers)
-
+    
+    def create_test(self, path, overwrite=None, headers=None):
```
mglee72, line 595: Where did these set of functions come from?

databricks_cli/sdk/service.py
```diff
@@ -169,55 +169,55 @@ def list_runs(self, job_id=None, active_only=None, completed_only=None, offset=N
...
         _data = {}
-
+    
         return self.client.perform_query('GET', '/clusters/list', data=_data, headers=headers)
-
+    
     def create_cluster(self, num_workers=None, autoscale=None, cluster_name=None, spark_version=None,
                        spark_conf=None, aws_attributes=None, node_type_id=None,
                        driver_node_type_id=None, ssh_public_keys=None, custom_tags=None,
-                       cluster_log_conf=None, init_scripts=None, spark_env_vars=None,
```
mglee72, line 218: Seems like you are removing the init_scripts param and the following variable-setting. I'm wondering this is part of what PR intends, and if so, we need to update the descriptions & possibly split the PR I think.

databricks_cli/secrets/cli.py
sushi1998: I think this was a result of resolving the merge conflict? Not my change, but certainly came from master

databricks_cli/sdk/service.py
sushi1998: This file was autogenerated save for a few necessary changes, but nothing around this particular line was perturbed manually.

databricks_cli/sdk/service.py
sushi1998: from autogeneration of this file - has nothing to do with my change. Seems like it was out of sync with changes made in universe.

databricks_cli/sdk/service.py
sushi1998: once again, this was a result of autogeneration, wasn't relevant to my changes at all. Should I revert this particular function back to it's original form?

mglee72 (approved):

databricks_cli/configure/cli.py
nfx: e.g. this means that AAD tokens are not going to be refreshed by our cli and we may get errors for commands running for too long? MSAL/ADAL libs provide token refresh functionality.

databricks_cli/secrets/cli.py
nfx: can we already create AKV scopes?

databricks_cli/sdk/service.py
```diff
@@ -488,19 +485,41 @@ def read(self, path, offset=None, length=None, headers=None):
         if length is not None:
             _data['length'] = length
         return self.client.perform_query('GET', '/dbfs/read', data=_data, headers=headers)
-
+    
+    def read_test(self, path, offset=None, length=None, headers=None):
```
nfx, line 489: why do we need `/dbfs-testing` for?

nfx:
this PR doesn't cover token refresh functionality. `az cli` refreshes tokens every 10 minutes and tokens that are about to expire in 30 seconds will fail requests.

databricks_cli/secrets/api.py
```diff
@@ -28,8 +28,10 @@ class SecretApi(object):
     def __init__(self, api_client):
         self.client = SecretService(api_client)
 
-    def create_scope(self, scope, initial_manage_principal):
-        return self.client.create_scope(scope, initial_manage_principal)
+    def create_scope(self, scope, initial_manage_principal, scope_backend_type,
```
marekbrysa, line 31: In the future, please try to avoid making breaking API changes. These new required positional arguments break anything that use this API.
Prefer to add new arguments with default values that retain the previous functionality, or instead add new methods if that cannot be done.

PR #326, opened 2020-09-22
DBFS API 429 retries: Fix bug on computing the time until the next retry

bogdanghita-db:
In https://github.com/databricks/databricks-cli/pull/319 we added 429 retries for the DBFS API.

Computing the time until the next retry as `retry_state.idle_for - time_for_last_retry` is not going to work as expected in the case where more than 1 call to functions wrapped with `retry_429` happens (e.g. in the case of `put_file`). The problem is that `time_for_last_retry` is only initialised once per databricks-cli command, while it should be initialised every time a function wrapped with `retry_429` is called, in the first attempt.

We fix this by initialising `time_for_last_retry` if `retry_state.attempt_number == 1`. This only works if the code is single-threaded, which is the case here.

Additional improvements:
* Remove unnecessary retry log, since we already have one.
* Increase max number of retries to 8, so that we have one additional retry with long delay.
* Use click.echo instead of logger, to maintain consistency with the rest of the code.
* Add clarifying note on `retry_state.idle_for`.
* Display seconds with 2 decimal places only.


databricks_cli/dbfs/api.py
```diff
@@ -113,7 +110,6 @@ def wrapped_function(*args, **kwargs):
             return func(*args, **kwargs)
         except HTTPError as e:
             if e.response.status_code == 429:
-                click.echo("Rate limit exceeded. Retrying with exponential backoff.")
```
bogdanghita-db, line 116: We already have the following logs, no need for this extra one.
```
Received 429 REQUEST_LIMIT_EXCEEDED for attempt 1. Retrying in 0.820262549404 seconds.
Received 429 REQUEST_LIMIT_EXCEEDED for attempt 2. Retrying in 0.0355085254586 seconds.
Received 429 REQUEST_LIMIT_EXCEEDED for attempt 3. Retrying in 1.40377268966 seconds.
Received 429 REQUEST_LIMIT_EXCEEDED for attempt 4. Retrying in 4.30875714394 seconds.
Received 429 REQUEST_LIMIT_EXCEEDED for attempt 5. Retrying in 15.1568721823 seconds.
Received 429 REQUEST_LIMIT_EXCEEDED for attempt 6. Retrying in 4.13532917455 seconds.
Received 429 REQUEST_LIMIT_EXCEEDED for attempt 7. Retrying in 49.7773688326 seconds.
```

tests/dbfs/test_api.py
```diff
@@ -40,6 +40,7 @@
     'file_size': 1
 }
 TEST_FILE_INFO = api.FileInfo(TEST_DBFS_PATH, False, 1)
+MAX_RETRY_429_ATTEMPTS = 8
```
gotibhai, line 43: nit: Consistency with `api.py` and keep it as `MAX_RETRY_ATTEMPTS`

databricks_cli/dbfs/api.py
```diff
@@ -94,13 +86,21 @@ class DbfsErrorCodes(object):
 def before_sleep_on_429(retry_state):
     global time_for_last_retry
```
gotibhai, line 87: Is this still needed now? This is using the global declaration. We can remove this since you're now initializing locally everytime now.

bogdanghita-db: Yes, we need it, since the function is called multiple times and `time_for_last_retry` needs to be incremented every time. Otherwise, on attempt 2 `time_for_last_retry` will be uninitialised and we'll get: `local variable 'time_for_last_retry' referenced before assignment`.

gotibhai: Ah of course! The same logic still holds. Gotcha!

gotibhai (approved):
Nice catch. Thanks for doing this.

PR #330, opened 2020-09-24
Add a --token-file flag to the databricks configure command

areese:
This flag causes the databricks configure command to read the token from the specified file instead of from stdin.
This can be used for places where the token is exposed as a secret in a file, but a complete configuration is not.


databricks_cli/configure/cli.py
```diff
@@ -84,29 +108,41 @@ def _configure_cli_password(profile, insecure):
...
     profile = get_profile_from_context()
     insecure_str = str(insecure) if insecure is not None else None
+
     if token:
-        _configure_cli_token(profile, insecure_str)
+        _configure_cli_token(profile=profile, insecure=insecure_str, host=host)
+    elif token_file:
+        _configure_cli_token_file(profile=profile, insecure=insecure_str, host=host,
+                                  token_file=token_file)
+        _configure_cli_token(profile=profile, insecure=insecure_str, host=host)
```
andrewmchen, line 134: Is this maybe a bug? I don't understand why we want to call this here.

areese: oops. that's a merge bug thanks.

andrewmchen (approved):

PR #331, opened 2020-09-25
import IPython directly instead of PythonShell

mengxr:
PythonShell is a private class. It should be the same if we import `IPython` directly.


mengxr:
cc: @aarondav

aarondav (approved):

aarondav:
Jenkins Trigger: MLPlatform/NightlyPR

PR #333, opened 2020-10-07
Add the `pipelines list` command to allow users to view all pipelines visible to them

aakash-db:
This PR adds the `pipelines list` command to the CLI, which allows users to view metadata about all the pipelines that they own. The list() API is a paginated request, and this command continually requests new pages until they've been exhausted before displaying them.

This API also includes updates to the sdk/api_client.py and sdk/service.py files, which come from newly auto-generated versions of the files.


aakash-db:
@mukulmurthy PTAL when you can!

databricks_cli/pipelines/cli.py
```diff
@@ -165,6 +165,25 @@ def get_cli(api_client, pipeline_id):
...
+@provide_api_client
+def list_cli(api_client):
+    pipelines_api = PipelinesApi(api_client)
+
+    response = pipelines_api.list()
+    pipelines = response["statuses"]
+
+    while "next_page_token" in response["pagination"]:
+        response = pipelines_api.list(page_token=response["pagination"]["next_page_token"])
+        pipelines.extend(response["statuses"])
```
mukulmurthy, line 182: Shouldn't this logic should go inside api.py, not cli.py?

databricks_cli/sdk/service.py
```diff
@@ -997,29 +997,43 @@ def deploy(self, pipeline_id=None, id=None, name=None, storage=None, configurati
...
+                                         headers=headers)
+
     def delete(self, pipeline_id=None, headers=None):
         _data = {}
-    
-        return self.client.perform_query('DELETE', '/pipelines/{pipeline_id}'.format(pipeline_id=pipeline_id), data=_data, headers=headers)
-    
+
+        return self.client.perform_query('DELETE', '/pipelines/{pipeline_id}'.format(pipeline_id=pipeline_id),
+                                         data=_data, headers=headers)
```
mukulmurthy, line 1007: Is this spacing what got generated or a copy/paste error?

tests/pipelines/test_api.py
```diff
@@ -313,3 +313,19 @@ def test_library_object_serialization_deserialization():
...
+    client_mock.assert_called_with('GET', '/pipelines',
+                                   data={"pagination": {"max_results": 10, "order_by": ["name asc"]}}, headers=None)
+
+
+def test_list_with_page_token(pipelines_api):
+    pipelines_api.list(page_token="a")
+    client_mock = pipelines_api.client.client.perform_query
+    assert client_mock.call_count == 1
+    client_mock.assert_called_with('GET', '/pipelines',
+                                   data={"pagination": {"page_token": "a"}}, headers=None)
```
mukulmurthy, line 331: There's no test coverage for paginated requests?

mukulmurthy (changes requested):

aakash-db:
@andrewmchen This PR includes changes to the `perform_query` method in `api_client.py`, which is shared across all of the CLI. The file is auto-generated, and the changes are simply copy-pasted from a regenerated version of the file, but they're causing other tests in the CLI to break. 

Specifically, what's happening is that the api_client.py previously had an if-case that turned request params for GET request into query params, while our API expects them to be part of the request body. 

What's the best approach to this? Should we create a new version of the file for our own CLI code, and keep that in the pipelines/ directory?

databricks_cli/sdk/service.py
aakash-db: Yup, it's all generated.

aakash-db: I'll revert of it except for the `list` function - it unnecessarily bloats this PR.

databricks_cli/sdk/service.py
```diff
@@ -997,29 +997,43 @@ def deploy(self, pipeline_id=None, id=None, name=None, storage=None, configurati
...
+
+        return self.client.perform_query('GET', '/pipelines/{pipeline_id}'.format(pipeline_id=pipeline_id), data=_data,
+                                         headers=headers)
+
+    def list(self, pagination=None, headers=None):
+        _data = {}
+        if pagination is not None:
+            if not isinstance(pagination, dict):
+                raise TypeError('Expected databricks.Pagination() or dict for field pagination')
+            _data = {"pagination." + key: value for (key, value) in pagination.items()}
```
aakash-db, line 1020: Note that this is not generated, but it's the best place I could think of to add this line (which is necessary for the request to work).

tests/pipelines/test_api.py
aakash-db: Added one in test_cli.py

databricks_cli/pipelines/cli.py
mukulmurthy: I'd say cli.py is for understanding and validating the user request, and api.py is for the business logic. For example, when deploying, we validate the parameters in cli.py, but api.py actually uploads local files to DBFS and updates the spec before calling our deploy endpoints.

databricks_cli/sdk/service.py
mukulmurthy: Is there something about the Pagination message or this API that makes this line necessary? As we've been told before, having manual changes here is an anti-pattern. If it's something like proto-messages-don't-work-with-GET-requests, we should change the service generation to do this so it works for all APIs.

aakash-db: True, there's nothing specific about this API. But, we do need special handling in the generated code.

Namely, [around this line](https://github.com/databricks/universe/blob/master/api-base/src/main/scala/com/databricks/api/base/PythonProtoClientGenerator.scala#L163) in the code generator needs to change, to path-compress data dicts for GET requests. 

I feel like this is an easier stopgap for now? Doing what I suggested will probably require a much larger discussion.

andrewmchen: We've needed to work around the SDK generation in other places like https://github.com/databricks/databricks-cli/blob/master/databricks_cli/jobs/api.py#L45. Can we do something similar here?

mukulmurthy: I like Andrew's suggestion better than making manual changes in service.py. Though I'd like to understand the issue a little more, so at the very least we can file a ticket with all the details. But best of all is to be good citizens and fix it ourselves.

Is the problem that the CLI is constructing incorrect requests, or that the server is incorrectly parsing correct requests? Why does your fix actually fix the issue?

aakash-db: The former. The code generated by our RPC parser doesn't work for our use-case. Basically, the problem is that for GET requests with nested data dicts , like `{"pagination" : { "max_results": 1 }}` the request the CLI produces is `GET /pipelines?pagination=max_results` which is obviously wrong. 

My fix turns `{"pagination" : { "max_results": 1 }}` into `{"pagination.max_results": 1}, which produces `/pipelines?pagination.max_results=1`, which the server can parse and handle correctly. What we need to do in code generator is this, essentially: recursively flatten all nested dicts for GET requests, while path-compressing the keys.

I think Andrew's idea makes sense here. I'll make a bug for actually fixing this upstream, which should be pretty straightforward.

aakash-db: Done.

tests/pipelines/test_api.py
aakash-db: Actually in test_api now, but it's there.

databricks_cli/pipelines/cli.py
aakash-db: Moved this into api.py.

databricks_cli/sdk/service.py
```diff
@@ -997,29 +997,42 @@ def deploy(self, pipeline_id=None, id=None, name=None, storage=None, configurati
...
+
+        return self.client.perform_query('GET', '/pipelines/{pipeline_id}'.format(pipeline_id=pipeline_id), data=_data,
+                                         headers=headers)
+
+    def list(self, pagination=None, headers=None):
+        _data = {}
+        if pagination is not None:
+            if not isinstance(pagination, dict):
+                raise TypeError('Expected databricks.Pagination() or dict for field pagination')
+        return self.client.perform_query('GET', '/pipelines', data=_data, headers=headers)
```
mukulmurthy, line 1020: Hmm, is this what gets auto-generated? Probably best to leave the line adding pagination into _data here, since we don't call it anyway and we don't want to confuse anyone else that regenerates and updates this file.

tests/pipelines/test_api.py
```diff
@@ -313,3 +313,107 @@ def test_library_object_serialization_deserialization():
...
+                       'cluster_id': '1023-090256-woke18',
+                       'name': 'Discovery Staging',
+                       'health': 'HEALTHY'},
+                      {'pipeline_id': '8',
+                       'state': 'RUNNING',
+                       'cluster_id': '1023-093334-cluck1',
+                       'name': 'Very Large Pipeline',
+                       'health': 'HEALTHY'}],
+         'pagination': {'prev_page_token': 'page4'}
+         }
```
mukulmurthy, line 397: Nit: Does the 4th page test anything that the first 3 don't? Should we shorten the test by just making it 3 pages?

mukulmurthy (commented):
Couple minor comments. Can you send me the bug to fix the generation behavior?

tests/pipelines/test_api.py
aakash-db: No, not really. I just took the actual response from list on dogfood and turned into a test.

Removed the last page.

databricks_cli/pipelines/api.py
```diff
@@ -60,6 +60,27 @@ def delete(self, pipeline_id, headers=None):
     def get(self, pipeline_id, headers=None):
         return self.client.get(pipeline_id, headers)
 
+    def list(self, max_results=None, order_by=None, headers=None):
```
mukulmurthy, line 63: Can we just get rid of all these default parameters if we're not using them yet? eventually these should come from user args, but better to add them then than to have them sit around unused now.

aakash-db: They're useful for unit testing, fwiw. It's useful to ensure they're plumbed through as expected.

mukulmurthy: Plumbed through from where to where? You're testing a variants of the API that are never actually used in the production code.

aakash-db: Making sure the perform_query calls have the right params. You're right it doesn't change much, so I'll remove it for now.

aakash-db: Done.

mukulmurthy (approved):

Issue #337, opened 2020-10-12
databricks workspace import_dir fails since 0.12.1


rom1gal:
Hi all,

Since version 0.12.1 (the same code works correctly when using version 0.12.0) we are not able to issue the following command (running from an Azure DevOps agent running Ubuntu 18.06) : 

`databricks workspace import_dir /home/vsts/work/r1/a/Notebooks /ASI.Test/20201012072959-release-47
`

The following error is triggered : 

```
Traceback (most recent call last):
  File "/usr/local/bin/databricks", line 5, in <module>
    from databricks_cli.cli import cli
  File "/usr/local/lib/python3.6/dist-packages/databricks_cli/cli.py", line 37, in <module>
    from databricks_cli.secrets.cli import secrets_group
  File "/usr/local/lib/python3.6/dist-packages/databricks_cli/secrets/cli.py", line 54, in <module>
    type=click.Choice(['AZURE_KEYVAULT', 'DATABRICKS'], case_sensitive=True),
TypeError: __init__() got an unexpected keyword argument 'case_sensitive'
```

Thanks in advance

rom1gal:
Hi,

Any news about that issue ?

Regards

nfx:
It seems that the version you're referring is too old.

PR #339, opened 2020-10-12
update minimal click version to allow case_sensitive option

rom1gal:
When using the CLI within specific environment (in my case an Azure DevOps agent running Ubuntu 18.06) some librairies need to be upgraded for the CLI to work properly (_click_ when running the 6.7 version triggers the stacktrace displayed here : https://github.com/databricks/databricks-cli/issues/337)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ setup(
     version=version,
     packages=find_packages(exclude=['tests', 'tests.*']),
     install_requires=[
-        'click>=6.7',
+        'click>=7.0',
         'requests>=2.17.3',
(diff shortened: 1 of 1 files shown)


andrewmchen:
Thanks for reporting and thank you for the fix. I'm wondering if we can just remove this option instead. https://github.com/databricks/databricks-cli/blob/master/databricks_cli/secrets/cli.py#L54.

@sushi1998

rom1gal:
Actually I do not understand the purpose of that option since it is currently not possible to create a Azure KV-backed secret scope :/

Issue #348, opened 2020-11-06
Doesn't seem to work with minimal version of click (i.e., click 6.7)?


bgreenwell:
Was not able to get any of the commands  working until I upgraded from click 6.7 to click 7.1.2; for example, with click 6.7 I  was  getting:

```bash
$ databricks --version
.
.
.
    type=click.Choice(['AZURE_KEYVAULT', 'DATABRICKS'], case_sensitive=True),
TypeError: __init__() got an unexpected keyword argument 'case_sensitive'
```

bgreenwell:
Looks like the case_sensitive argument was added in 0.7.0; per the change log: 

Add case_sensitive=False as an option to Choice. :issue:`569`

chrisst:
**This repository has been archived.**

The legacy `databricks/databricks-cli` is deprecated and has been fully replaced by the new Databricks CLI at **https://github.com/databricks/cli**.

We are closing all open issues here as part of archiving this repository. If your issue still affects the new CLI, we'd appreciate you re-opening it at https://github.com/databricks/cli/issues so the team can track and address it there.

PR #349, opened 2020-11-09
Update minimal version of click to 7.0

bgreenwell:
Patch to fix issue https://github.com/databricks/databricks-cli/issues/348 regarding incorrect minimal version of [click](https://github.com/pallets/click/blob/master/CHANGES.rst).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ setup(
     version=version,
     packages=find_packages(exclude=['tests', 'tests.*']),
     install_requires=[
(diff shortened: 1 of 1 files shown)


pietern:
Belated thanks for making the fix and posting a PR. We're slowly making our way through the backlog this past month.

The `case_sensitive` argument was removed in #347.

However, in the mean time we merged #471 which depends on `click >= 7.0` for a different reason.

PR #353, opened 2020-11-30
Avoid indexing potentially empty dictionaries in the `pipelines list` command

aakash-db:
The CLI should gracefully handle cases where there are missing fields in the response. This PR replaces all direct dictionary indexes with `dict.get()` calls, with sane default values, to ensure we don't throw an error when the API response does not have all fields present.

### DIFF
--- a/databricks_cli/pipelines/api.py
+++ b/databricks_cli/pipelines/api.py
@@ -74,11 +74,11 @@ class PipelinesApi(object):
                 'GET', '/pipelines', data=_data, headers=headers)
 
         response = call()
-        pipelines = response["statuses"]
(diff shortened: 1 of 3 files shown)


aakash-db:
@mukulmurthy please take a look when you can!

mukulmurthy (commented):
Change looks good, but please fix the description. Also the build is failing, I'm not sure why.

aakash-db:
@andrewmchen I'm not sure why this build is broken, but it doesn't seem related to this change. How should I go about fixing this?

mukulmurthy (approved):

PR #354, opened 2020-12-03
Add runs export command including model parsing

marekbrysa:
* Add support for the [runs export API](https://docs.databricks.com/api/latest/jobs.html#runs-export).
* It returns a JSON with HTML content and the actual notebook model is embedded it in in an unreadable format, which makes it not very useful for seeing e.g. errors in the run. This also optionally parses the model to improve that.


chrisst:

PR #357, opened 2020-12-21
Update library type support in pipelines cli

rgovind3:
This PR enables the pipelines cli to support arbitrary library types, thereby enabling support for new library types in the future without additional cli updates.


rgovind3:
cc @mukulmurthy @marmbrus

marmbrus:
Makes sense to me.  Clients in general should not be doing validations that make the system less evolvable.

mukulmurthy (approved):

PR #360, opened 2021-02-09
[SC-67303] Add modification_time field to dbfs FileInfo Object

stormwindy:
This is a change that will soon be added to the Databricks DBFS REST API. Change adds a new `modification_time` Long field to `FileInfo` Object. This field is intended to be received through `dbfs ls -l <path>` or `databricks fs ls -l <path>`. Changes are backwards compatible. Hence, this change can be merged before the backend functionality becomes available.

### DIFF
--- a/databricks_cli/dbfs/api.py
+++ b/databricks_cli/dbfs/api.py
@@ -57,16 +57,20 @@ class FileInfo(object):
         if is_long_form:
             filetype = 'dir' if self.is_dir else 'file'
             # Do not return modification time if it is not available from server.
-            return ([filetype, self.file_size, stylized_path] if self.modification_time is None else 
-                    [filetype, self.file_size, stylized_path, self.modification_time])
+            return (
+                [filetype, self.file_size, stylized_path] if self.modification_time is None else
+                [filetype, self.file_size, stylized_path, self.modification_time]
+            )
         return [stylized_path]
 
     @classmethod
     def from_json(cls, json):
         dbfs_path = DbfsPath.from_api_path(json['path'])
         # If JSON doesn't include modification_time data, replace it with None.
-        return (cls(dbfs_path, json['is_dir'], json['file_size'], json['modification_time']) if 'modification_time' in json else
-                cls(dbfs_path, json['is_dir'], json['file_size'], None))
+        return (
+            cls(dbfs_path, json['is_dir'], json['file_size'], None) if 'modification_time' not in json else
+            cls(dbfs_path, json['is_dir'], json['file_size'], json['modification_time'])
+        )
 
     def __eq__(self, other):
         if isinstance(other, self.__class__):
--- a/databricks_cli/dbfs/api.py
+++ b/databricks_cli/dbfs/api.py
@@ -57,20 +57,20 @@ class FileInfo(object):
         if is_long_form:
             filetype = 'dir' if self.is_dir else 'file'
             # Do not return modification time if it is not available from server.
-            return (
-                [filetype, self.file_size, stylized_path] if self.modification_time is None else
-                [filetype, self.file_size, stylized_path, self.modification_time]
-            )
+            if self.modification_time is None:
+                return [filetype, self.file_size, stylized_path]
+            else:
+                return [filetype, self.file_size, stylized_path, self.modification_time]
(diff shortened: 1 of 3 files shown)


andrewmchen:
Could we update docs here? https://github.com/databricks/databricks-cli/blob/master/databricks_cli/dbfs/cli.py#L39 Also please make sure you document what the time format is.

stormwindy:
@andrewmchen Thanks you for having a look! Just updated the doc as you described.

tests/dbfs/test_api.py
```diff
@@ -34,12 +36,14 @@
 from databricks_cli.dbfs.exceptions import LocalFileExistsException
 
 TEST_DBFS_PATH = DbfsPath('dbfs:/test')
+DUMMY_TIME = time.time() * 1000
```
bogdanghita-db, line 39: I think it's better to use a fixed value here, instead of the current time. This is to make the test deterministic and consistent for each run. It may not matter that much for this particular test, though.

databricks_cli/dbfs/api.py
```diff
@@ -45,29 +45,39 @@ class ParseException(Exception):
...
 
     @classmethod
     def from_json(cls, json):
         dbfs_path = DbfsPath.from_api_path(json['path'])
-        return cls(dbfs_path, json['is_dir'], json['file_size'])
+        # If JSON doesn't include modification_time data, replace it with None.
+        if 'modification_time' not in json:
+            return cls(dbfs_path, json['is_dir'], json['file_size'], None)
+        else:
+            return cls(dbfs_path, json['is_dir'], json['file_size'], json['modification_time'])
```
bogdanghita-db, line 73: nit: Another way in which you could do this is the following:
```
modification_time = json['modification_time'] if `modification_time` in json else None
return cls(dbfs_path, json['is_dir'], json['file_size'], modification_time)
```
Similar to the other comment, I don't have a strong opinion about it, it's just a matter of style.

databricks_cli/dbfs/api.py
```diff
@@ -45,29 +45,39 @@ class ParseException(Exception):
...
         path = self.dbfs_path.absolute_path if is_absolute else self.dbfs_path.basename
         stylized_path = click.style(path, 'cyan') if self.is_dir else path
         if is_long_form:
             filetype = 'dir' if self.is_dir else 'file'
-            return [filetype, self.file_size, stylized_path]
+            # Do not return modification time if it is not available from server.
+            if self.modification_time is None:
+                return [filetype, self.file_size, stylized_path]
+            else:
+                return [filetype, self.file_size, stylized_path, self.modification_time]
```
bogdanghita-db, line 63: nit: Another way in which you could do this is the following:
```
row = [filetype, self.file_size, stylized_path]
if self.modification_time is not None:
  row.append(self.modification_time)
return row
```
This avoids duplicating the code and scales if we want to add more optional parameters. But I don't have a strong opinion about it, it's just a matter of style.

bogdanghita-db (commented):
LGTM overall. I left a few comments, mostly nits. It would be good to address the one about initialising `DUMMY_TEST` with a fixed value, though.

Also, could you add the result of an end-to-end manual test to the PR description?  I mean an actual `dbfs ls -l <path>` command against a shard that has the backend changes.

databricks_cli/dbfs/api.py
stormwindy: I prefer the way you have shown as well but a similar example right below these were written with the current style. However, having a second look to the file both styles seem to be used. I will apply these changes to avoid code duplication.

tests/dbfs/test_api.py
```diff
@@ -25,6 +25,8 @@
 from base64 import b64encode
 
 import os
+import time
```
bogdanghita-db, line 28: I think this is what `lint.sh` complains about. You probably forgot to remove it in your last commit.

bogdanghita-db (approved):
LGTM, pending fixing `lint.sh` error and passing tests.

tests/dbfs/test_api.py
stormwindy: Thanks for pointing out.

PR #361, opened 2021-02-25
[SC-71155] Fix TypeError in Pipelines CLI

jon-mio:
We were indexing by key into a string instead of a dict

### DIFF
--- a/tests/pipelines/test_cli.py
+++ b/tests/pipelines/test_cli.py
@@ -307,5 +307,6 @@ def test_deploy_pipeline_conflicting_ids(pipelines_api_mock, tmpdir):
 
     result = CliRunner().invoke(cli.deploy_cli, ['--spec', path, '--pipeline-id', "fake"])
     assert result.exit_code == 1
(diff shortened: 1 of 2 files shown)


tests/pipelines/test_cli.py
```diff
@@ -307,4 +307,5 @@ def test_deploy_pipeline_conflicting_ids(pipelines_api_mock, tmpdir):
 
     result = CliRunner().invoke(cli.deploy_cli, ['--spec', path, '--pipeline-id', "fake"])
     assert result.exit_code == 1
+    assert "ValueError" in result.stdout
```
jon-mio, line 310: weirdly the CliRunner does not include the ValueError exception as an object so we have to check the output directly

mukulmurthy: Instead of asserting "ValueError", shouldn't we assert some snippet of that helpful error message?

jon-mio: added more of the msg

mukulmurthy (approved):

PR #365, opened 2021-03-10
[SC-71389] Better exceptions for Pipelines CLI

jon-mio:
Add validation on missing pipeline id and throw `ValueError` instead of `RuntimeError`.

Can you review @mukulmurthy? Can't tag like normal.

### DIFF
--- a/databricks_cli/pipelines/cli.py
+++ b/databricks_cli/pipelines/cli.py
@@ -88,7 +88,7 @@ def deploy_cli(api_client, spec_arg, spec, allow_duplicate_names, pipeline_id):
     databricks pipelines deploy --pipeline-id 1234 --spec example.json
     """
     if bool(spec_arg) == bool(spec):
-        raise RuntimeError('The spec should be provided either by an option or argument')
+        raise ValueError('The spec should be provided either by an option or argument')
     src = spec_arg if bool(spec_arg) else spec
     spec_obj = _read_spec(src)
     if not pipeline_id and 'id' not in spec_obj:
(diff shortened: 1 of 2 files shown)


databricks_cli/pipelines/cli.py
```diff
@@ -271,27 +271,11 @@ def _write_spec(src, spec):
         f.write(data)
 
 
-def _get_pipeline_id(spec_arg, spec, pipeline_id):
```
jon-mio, line 274: unused code

databricks_cli/pipelines/cli.py
```diff
@@ -88,7 +88,7 @@ def deploy_cli(api_client, spec_arg, spec, allow_duplicate_names, pipeline_id):
     databricks pipelines deploy --pipeline-id 1234 --spec example.json
     """
     if bool(spec_arg) == bool(spec):
-        raise RuntimeError('The spec should be provided either by an option or argument')
+        raise ValueError('The spec should be provided either by an option or argument')
```
mukulmurthy, line 91: What does RuntimeError do vs ValueError?

mukulmurthy (approved):

databricks_cli/pipelines/cli.py
jon-mio: Just slightly more accurate exception IMO (https://stackoverflow.com/questions/256222/which-exception-should-i-raise-on-bad-illegal-argument-combinations-in-python) RuntimeError is pretty generic

PR #367, opened 2021-03-17
Initial support for Databricks Repos

alexott:
This PR adds an initial support for [Databricks Repos](https://docs.databricks.com/repos.html) feature using the [Repos API](https://docs.databricks.com/dev-tools/api/latest/repos.html).  

Summary of changes:
* the new top level command was added: `repos`
* the `update` subcommand is used to checkout the repository specified via `--repos-id` switch to the branch specified via `--branch` switch
* the `get-repos-id` subcommand uses Workspaces API to find the Repos ID based on the file path.  It will be used to simplify the integration with CI/CD systems, etc., so people can write something like this instead of looking for Repos ID in UI:
```sh
REPOS_ID=$(databricks repos get-repos-id /Repos/Production/some-project)
databricks repos update --repos-id $REPOS_ID --branch releases
```

### DIFF
--- a/databricks_cli/repos/api.py
+++ b/databricks_cli/repos/api.py
@@ -20,12 +20,32 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from databricks_cli.sdk import ReposService
+import os.path
+import requests
+from databricks_cli.sdk import ReposService, WorkspaceService
 
 
 class ReposApi(object):
     def __init__(self, api_client):
         self.client = ReposService(api_client)
+        self.ws_client = WorkspaceService(api_client)
 
     def update(self, repo_id, branch):
         return self.client.update(repo_id, branch)
+
+    def get_repos_id(self, path):
+        if not path.startswith("/Repos/"):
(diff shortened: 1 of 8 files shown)


databricks_cli/repos/api.py
```diff
@@ -0,0 +1,51 @@
...
+
+class ReposApi(object):
+    def __init__(self, api_client):
+        self.client = ReposService(api_client)
+        self.ws_client = WorkspaceService(api_client)
+
+    def update(self, repo_id, branch):
+        return self.client.update(repo_id, branch)
+
+    def get_repos_id(self, path):
```
anders-databricks, line 36: Can we leave this feature out for now? We're planning on extending the Repos API with additional CRUD endpoints, and they will achieve this same goal.

alexott: you mean `get_repos_id` or both?

anders-databricks: Yeah I meant just the get-repos-id subcommand!

nfx:
i wonder if we can compress it to single command. e.g. `databricks repos checkout /Repos/Production/some-project master`.

alexott:
let's wait for the more APIs then...

PR #368, opened 2021-04-27
Interpret pipeline libraries relative to spec

anew:
The library paths in a pipeline spec should be interpreted as relative to the spec's location, and not relative to the current working directory.

### DIFF
--- a/tests/pipelines/test_api.py
+++ b/tests/pipelines/test_api.py
@@ -103,10 +103,12 @@ def _test_library_uploads(pipelines_api, api_method, spec, put_file_mock, dbfs_p
     # set-up the test
     jar1 = tmpdir.join('jar1.jar').strpath
     jar2 = tmpdir.join('jar2.jar').strpath
-    jar3 = tmpdir.join('jar3.jar').strpath
+    jar3dir = 'some/relative/path'
+    jar3absdir = tmpdir.join(jar3dir).strpath
+    jar3relpath = os.path.join(jar3dir, 'jar3.jar')
+    jar3abspath = tmpdir.join(jar3relpath).strpath
     jar4 = tmpdir.join('jar4.jar').strpath
     wheel1 = tmpdir.join('wheel-name-conv.whl').strpath
-    jar3_relpath = os.path.relpath(jar3, os.getcwd())
     jar4_file_prefix = 'file:{}'.format(jar4)
     remote_path_456 = 'dbfs:/pipelines/code/51eac6b471a284d3341d8c0c63d0f1a286262a18.jar'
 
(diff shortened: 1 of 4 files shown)


anew:
@mukulmurthy please take a look

tests/pipelines/test_api.py
```diff
@@ -171,13 +174,13 @@ def test_create(pipelines_api):
     spec = copy.deepcopy(SPEC_WITHOUT_ID)
     spec['libraries'] = []
 
-    pipelines_api.create(spec, allow_duplicate_names=False)
+    pipelines_api.create(spec, spec_dir='.', allow_duplicate_names=False)
```
mukulmurthy, line 177: If all these tests assume that the spec dir is the same as the current dir, would any of them actually fail even without the code changes?

tests/pipelines/test_cli.py
```diff
@@ -273,19 +273,19 @@ def test_allow_duplicate_names_flag(pipelines_api_mock, tmpdir):
         f.write(DEPLOY_SPEC_NO_ID)
     runner = CliRunner()
     runner.invoke(cli.deploy_cli, [path])
-    assert pipelines_api_mock.create.call_args_list[0][0][1] is False
+    assert pipelines_api_mock.create.call_args_list[0][0][2] is False
```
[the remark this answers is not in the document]
anew: These test that the `allow_duplicate_names` parameter of `create` and `deploy` is set correctly. Because I added a new parameter `spec_dir`, this shifts by one from position 1 to 2.

tests/pipelines/test_api.py
anew: these tests use a spec without libraries, so it makes no difference to them.

mukulmurthy: So which one's the test that would fail without the prod code changes?

anew: This one: https://github.com/databricks/databricks-cli/pull/368/files#diff-66dd45a6659dae8bccf642b5526207c7af3ab20f872109aa759bd9fbfd76b070R88

tests/pipelines/test_api.py
```diff
@@ -152,12 +155,12 @@ def _test_library_uploads(pipelines_api, api_method, spec, put_file_mock, dbfs_p
     ]
     expected_data['allow_duplicate_names'] = allow_duplicate_names
 
-    api_method(spec, allow_duplicate_names)
+    api_method(spec, tmpdir.strpath, allow_duplicate_names)
```
anew, line 158: it's not obvious, but `api_method` is either `create` or `deploy`. So this is the test that would actually fail.

mukulmurthy: Ah, didn't realize that you switched the spec to use a relative path to itself rather than to the cwd.

mukulmurthy (approved):

PR #371, opened 2021-05-07
Refactor DBFS CLI Put to support new backend.

stormwindy:
Refactors PUT methods in CLI (without creating user facing APIs) for CLI to use new `put` backend of DBFS rather than using `create`, `add_block` and `close` methods to achieve same results. Changes, in short, create a multipart/form request and sent to `/dbfs/put` backend. Files with `>=2Gb` fall back to using `create, add_block, close` (streaming upload) to not break any pipelines.

Version number is not increased for this change since there will not be a new release specific for this change. It will be piggy backed to next release.

Changes are tested on a staging shard. Files sizes of `1kb`, `1mb`, `10mb`, `500mb`, `750mb`, `1gb`, `1.5gb` and `3gb` have been tested by exposing an interface for `put_file` (not included in the PR) and initiating an upload command `dbfs put_file <file-path>, <dbfs-path>`. Later the files on dbfs and local are compared. `3gb` file upload has returned an error as expected since `put` API supports up to 2Gbs.  (For this testing streaming upload fallback was removed.)

Moreover, changes are tested by using `cp` command of the CLI which internally uses `put`. Files have been copied from one directory to another with sizes `1mb`, `100mb`, `1gb`, `3gb`.

After the testing a fall-back logic has been added to `put_file` method so that file uploads larger than 2gb automatically uses streaming uploads with `open`, `add_block` and `close` instead of `put` API.


databricks_cli/dbfs/cli.py
```diff
@@ -165,4 +165,4 @@ def cat_cli(api_client, src):
 dbfs_group.add_command(rm_cli, name='rm')
 dbfs_group.add_command(cp_cli, name='cp')
 dbfs_group.add_command(mv_cli, name='mv')
-dbfs_group.add_command(cat_cli, name='cat')
+dbfs_group.add_command(cat_cli, name='cat')
```
gotibhai, line 168: nit: why removing the line?

databricks_cli/dbfs/api.py
```diff
@@ -114,15 +114,12 @@ def get_status(self, dbfs_path, headers=None):
...
-                contents = local_file.read(BUFFER_SIZE_BYTES)
-                if len(contents) == 0:
-                    break
-                # add_block should not take a bytes object.
-                self.client.add_block(handle, b64encode(contents).decode(), headers=headers)
-            self.client.close(handle, headers=headers)
+        self.client.put(dbfs_path.absolute_path, src_path=src_path,
+                        overwrite=overwrite, headers=headers)
+
+    def put_content(self, content, dbfs_path, overwrite, headers=None):
```
gotibhai, line 120: A comment here about the need and two different functions would be useful for future context.

databricks_cli/sdk/service.py
```diff
@@ -519,25 +523,45 @@ def list_test(self, path, headers=None):
...
             _data['overwrite'] = overwrite
-        return self.client.perform_query('POST', '/dbfs/put', data=_data, headers=headers)
-
-    def put_test(self, path, contents=None, overwrite=None, headers=None):
-        _data = {}
+        if src_path is not None:
+            # @self.client sets Content-Type 'text/json' by default.
+            # For multipart/form-data POST Content-Type should be set automatically
+            # to decode 'Boundary' parameter.
+            headers = {'Content-Type': None}
```
gotibhai, line 541: This doesn't seem right, The content Type is expected to be something of this format 
`Content-Type: multipart/form-data; boundary=something` 
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type

See how we check whether a request is a multipart upload or not // https://livegrep.dev.databricks.com/view/databricks/universe/daemon/data/daemon/src/main/scala/com/databricks/backend/daemon/data/server/meta/DbfsFileUploadDownloadBackend.scala#L305

databricks_cli/sdk/service.py
```diff
@@ -519,25 +523,45 @@ def list_test(self, path, headers=None):
...
+            filename = os.path.basename(src_path)
+            _files = {'file': (filename, open(src_path, 'rb'), 'multipart/form-data')}
```
gotibhai, line 543: What's this struct/tuple you're passing for files? Is this format defined somewhere or did you create it? How does the request know to send these files as a multipart upload?

databricks_cli/sdk/service.py
```diff
@@ -519,25 +523,45 @@ def list_test(self, path, headers=None):
...
         if path is not None:
             _data['path'] = path
         if contents is not None:
-            _data['contents'] = contents
+            # Because terminal might add trailing newlines, they need to be encoded properly.
+            encoded_contents = base64.b64encode(contents.encode('utf-8'))
+            _data['contents'] = encoded_contents.decode("utf-8")
         if overwrite is not None:
             _data['overwrite'] = overwrite
-        return self.client.perform_query('POST', '/dbfs-testing/put', data=_data, headers=headers)
```
gotibhai, line 540: What path is this exactly? I don't see a mention of this in the codebase except in `service.proto` ? I don't think these are used. We should create a ticket to clean this up from universe? cc @bogdanghita-db 

```
  rpc putTest(Put) returns (Put.Response) {
    option (rpc) = {
      endpoints: {
        method: "POST",
        path: "/dbfs-testing/put",
        since: { major: 2, minor: 0 },
      },
      visibility: PUBLIC,
    };
  }
  ```

File: tests/dbfs/test_api.py [hunk 6]
```diff
@@ -133,14 +133,9 @@ def test_put_file(self, dbfs_api, tmpdir):
         api_mock = dbfs_api.client
         test_handle = 0
         api_mock.create.return_value = {'handle': test_handle}
+        # Should succeed.
```
gotibhai, line 136: This comment isn't really helpful.
Can you assert that it succedded by using other API's like list and matching the content?

gotibhai (commented):
Took a first pass, will take a look after comments are addressed. Could you add a note about testing in the description.

databricks_cli/sdk/service.py
stormwindy: I have checked this quite a lot. If we set the `content-type` manually, requests library forces programmer to define other required fields such as `boundary`. If the `content-type` is not set (or `None`), requests library automatically fills them. If a `files` parameter is passed to the call, it will automatically generate `Content-Type: multipart/form-data; boundary=something`.

A lot of answers on stack I checked were against setting 'content-type' in this case.

databricks_cli/sdk/service.py
stormwindy: Thanks for letting me know about this. @bogdanghita-db should I edit the `service.proto` file to have new parameters added? I will have to add `src_path` if we want to keep current design choices on how to implement new put. (the ones I did.)

databricks_cli/sdk/service.py
stormwindy: If you go to `perform_query`, it passes a `file=` argument to the request. When that is the case, POST requests becomes a multipart upload. It is explained in requests docs: https://docs.python-requests.org/en/master/user/quickstart/#post-a-multipart-encoded-file.

tests/dbfs/test_api.py
stormwindy: I don't think this is possible since the API is a mock. It would only be possible to do this if I explicitly define what the return value of other APIs are which does not help us for testing. Please, correct me if I am wrong.

tests/dbfs/test_api.py
```diff
@@ -129,17 +129,27 @@ def test_put_file(self, dbfs_api, tmpdir):
         test_file_path = os.path.join(tmpdir.strpath, 'test')
         with open(test_file_path, 'wt') as f:
             f.write('test')
-
         api_mock = dbfs_api.client
         test_handle = 0
         api_mock.create.return_value = {'handle': test_handle}
         dbfs_api.put_file(test_file_path, TEST_DBFS_PATH, True)
 
-        assert api_mock.add_block.call_count == 1
```
bogdanghita-db, line 138: We can assert that `add_block.call_count == 0` instead, right?

tests/dbfs/test_api.py
```diff
@@ -129,17 +129,27 @@ def test_put_file(self, dbfs_api, tmpdir):
...
+    #     api_mock = dbfs_api.client
+    #     test_handle = 0
+    #     api_mock.create.return_value = {'handle': test_handle}
+    #     dbfs_api.put_file(test_file_path, TEST_DBFS_PATH, True)
+    #     add_block_expected = math.ceil(2254856676 / (2**20))
+    #     assert True
+    #     # assert api_mock.add_block.call_count == add_block_expected
+    #     # assert test_handle == api_mock.add_block.call_args[0][0]
+    #     # assert api_mock.close.call_count == 1
+    #     # assert test_handle == api_mock.close.call_args[0][0]
```
bogdanghita-db, line 152: We can keep this test if we override the 2GB constant in `class DbfsApi(object)` to something small (e.g. 1B) for the test. Let's do it if there's an easy and clean way of doing it. Otherwise, we can just remove this commented block.

databricks_cli/dbfs/api.py
```diff
@@ -113,16 +113,25 @@ def get_status(self, dbfs_path, headers=None):
         json = self.client.get_status(dbfs_path.absolute_path, headers=headers)
         return FileInfo.from_json(json)
 
+    # Single variation of put implemented.
+    # See https://docs.databricks.com/dev-tools/api/latest/dbfs.html#put
+    # @put_file() is for multipart file upload.
```
bogdanghita-db, line 118: This comment is a bit confusing for me. What do you mean by `Single variation`? Why do you say `@put_file() is for multipart file upload`? Did you mean `self.client.put` instead?

bogdanghita-db (commented):
LGTM overall, but I see the PR tests are failing.

Thanks for the description on how you tested. I understand from it that you made some changes to the code to test. It would be good to also test the final code end-to-end with `databricks fs cp` directly against a test shard, if you didn't do it already.

It would be good to get a review from @andrewmchen as well.

tests/dbfs/test_api.py
stormwindy: Ah yes! I was explicitly going to ask what you think about this test before requesting a review. Sorry about that. I will have a look if there is a way of doing that. Thanks you for the input.

databricks_cli/sdk/api_client.py
```diff
@@ -125,8 +125,13 @@ def perform_query(self, method, path, data = {}, headers = None):
...
                                             verify = self.verify, headers = headers)
             else:
-                resp = self.session.request(method, self.url + path, data = json.dumps(data),
-                                            verify = self.verify, headers = headers)
+                if files is None:
+                    resp = self.session.request(method, self.url + path, data = json.dumps(data),
+                                                verify = self.verify, headers = headers)
+                else:
+                    # Multipart file upload
+                    resp = self.session.request(method, self.url + path, files = files, data = data,
```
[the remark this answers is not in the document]
stormwindy: With `json.dumps` it used to fail creating the correct request. I saw somewhere the solution to the error was to pass the data object directly and request library handles it itself. I will try to find the thread about it.

databricks_cli/dbfs/api.py
stormwindy: I tried to mean the following method. I will make it more clear and also mention the underlying `put` method used.

databricks_cli/dbfs/api.py
```diff
@@ -113,16 +113,25 @@ def get_status(self, dbfs_path, headers=None):
...
-        with open(src_path, 'rb') as local_file:
-            while True:
-                contents = local_file.read(BUFFER_SIZE_BYTES)
-                if len(contents) == 0:
-                    break
-                # add_block should not take a bytes object.
-                self.client.add_block(handle, b64encode(contents).decode(), headers=headers)
-            self.client.close(handle, headers=headers)
+        # If file size is >2Gb use streaming upload.
+        if os.path.getsize(src_path) <= 2147483648:
```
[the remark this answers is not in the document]
stormwindy: I will make it `<` just to be safe. Sometimes dummy files I generate vary by 1 byte for some reason. So tests might not be a great indicator in fine grained cases.

bogdanghita-db (approved):

PR #377, opened 2021-06-29
Point out invalid characters during pipeline_id vaidation

z4f1r0v:
It would be useful for clients to get information not only regarding their `pipeline_id`'s being malformed, but also in what way. This PR gives the users back the actual characters that need to be removed so that the `pipeline_id` satisfies the validation requirements.


z4f1r0v:
Hmm, the validation failed but I haven't not made changes to the logic of the code. Can anyone assist here?

z4f1r0v:
PIng @andrewmchen

pietern:
Thanks for the PR.

Validation of pipeline IDs was removed in #437, which removes the need for this change.

PR #380, opened 2021-07-30
Repos API support

a-cong:
Adding CLI support for the repos API (docs to be published soon).

Summary of changes:

- added new top-level `repos` command
- added support for `list` (lists all repos the user has Manage permissions on), `get` (gets a repo), `create` (creates a repo and links it to the given remote Git repo), `update` (checks out a repo to the given branch or tag), and `delete` (deletes the repo)
- `get`, `update`, and `delete` take either the repo ID or path as the repo identifier


databricks_cli/repos/api.py
```diff
@@ -0,0 +1,81 @@
...
+    def __init__(self, api_client):
+        self.client = ReposService(api_client)
+        self.ws_client = WorkspaceService(api_client)
+
+    def get_repo_id(self, path):
+        if not path.startswith("/Repos/"):
+            raise ValueError("Path must start with /Repos/ !")
+
+        p = path
+        while p != "/Repos":
```
andrewmchen, line 38: Is it necessary for users to be able to do ```
databricks repos get --path /Repos/alexandra@databricks.com/test
```

Maybe it would be simpler if you could only pass in the `/Repos/alexandra@databricks.com` path?

</div></b>

File: tests/repos/test_cli.py [hunk 2]
```diff
@@ -0,0 +1,131 @@
...
+        None
+    )
+
+
+@provide_conf
+def test_update_tag_validation(repos_api_mock):
+    runner = CliRunner()
+    runner.invoke(cli.update_repo_cli,
+                  ["--repo-id", TEST_ID,
+                   "--tag", TEST_TAG])
```
andrewmchen, line 116: Is it possible to add a test which shows that if you update with both `--branch` and `--tag` it will fail?

andrewmchen (approved):

databricks_cli/repos/api.py
a-cong: Sure, that makes sense

hjsuh18 (approved):

tests/repos/test_cli.py
a-cong: I added a test like `test_add_member_validation` in `tests/groups/test_cli.py`, basically just asserting the method doesn't get called

PR #385, opened 2021-08-16
Relax requirement for specification of Git provider for Repos

alexott:
Before this PR, the `--provider` option was required, but in many cases we can guess the provider ID from the URL (for hosted providers as Gitlab, GitHub, etc.) - this PR is making `--provider` option not required, but if we can't guess Git provider, it will error out.

### DIFF
--- a/databricks_cli/sdk/service.py
+++ b/databricks_cli/sdk/service.py
@@ -24,6 +24,8 @@
 # limitations under the License.
 #
 import os
+import re
+
 from six.moves.urllib.parse import urlparse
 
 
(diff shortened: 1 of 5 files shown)


databricks_cli/sdk/service.py
```diff
@@ -1050,10 +1051,30 @@ def stop(self, pipeline_id=None, headers=None):
...
+    def __detect_repo_provider__(self, url):
+        provider = None
+        try:
+            netloc = urlparse(url).netloc
+            idx = netloc.rfind("@")
+            if idx != -1:
+                netloc = netloc[(idx+1):]
+            provider = ReposService.__git_providers__.get(netloc.lower())
+        except:
+            pass
```
victorsun123, line 1075: do we want to throw some error message here so it's clear to the user they need to specify the git provider if the request fails?

victorsun123:
Left some comments but otherwise looks good! Also, is there any documentation for this CLI, we should make it clear in the documentation that we will parse provider if not specified and it will throw an error if unable to parse.

databricks_cli/sdk/service.py
alexott: Added it in another place instead

alexott:
Added some tests, added explicit exception. Doc string for `--provider` parameter was already saying that it's required if we can't detect the git provider from URL

alexott:
@victorsun123 any chance to review it?

nfx:
@alexott please rebase :)

alexott:
@nfx rebased

nfx:
@alexott tests are failing

tests/repos/test_cli.py
```diff
@@ -75,6 +75,18 @@ def test_create_validation(repos_api_mock):
     )
 
 
+@provide_conf
+def test_create_validation_partial(repos_api_mock):
+    runner = CliRunner()
+    runner.invoke(cli.create_repo_cli,
+                  ["--url", TEST_URL])
+    repos_api_mock.create.assert_called_once_with(
```
nfx, line 83: can you also add a test for `awsCodeCommit`, as it has the deepest logic?..

nfx (changes requested):

tests/repos/test_cli.py
alexott: Added tests...

alexott:
@nfx all comments are addressed...

nfx (approved):

PR #390, opened 2021-09-24
Adding OAuth support to the Databricks CLI

sweisdb:
This change adds experimental support for OAuth to the Databricks CLI.

### DIFF
--- a/databricks_cli/configure/cli.py
+++ b/databricks_cli/configure/cli.py
@@ -27,6 +27,8 @@ from os import path
 import click
 from click import ParamType
 
+from databricks_cli.oauth.oauth import get_tokens
+
 from databricks_cli.configure.config import profile_option, get_profile_from_context, debug_option
 from databricks_cli.configure.provider import DatabricksConfig, update_and_persist_config, \
     ProfileConfigProvider
@@ -38,6 +40,15 @@ PROMPT_USERNAME = 'Username'
 PROMPT_PASSWORD = 'Password'  # NOQA
 PROMPT_TOKEN = 'Token'  # NOQA
 ENV_AAD_TOKEN = 'DATABRICKS_AAD_TOKEN'
+DEFAULT_SCOPES = [
+    'accounts',
+    'clusters',
+    'mlflow',
+    'offline_access',
+    'scim',
+    'sql',
+    'unity-catalog'
+]
 
 
 def _configure_cli_token_file(profile, token_file, host, insecure, jobs_api_version):
@@ -51,7 +62,11 @@ def _configure_cli_token_file(profile, token_file, host, insecure, jobs_api_vers
     if not host:
         host = click.prompt(PROMPT_HOST, default=config.host, type=_DbfsHost())
 
-    new_config = DatabricksConfig.from_token(host, token, insecure, jobs_api_version)
+    new_config = DatabricksConfig.from_token(host=host,
+                                             token=token,
+                                             refresh_token=None,
+                                             insecure=insecure,
+                                             jobs_api_version=jobs_api_version)
     update_and_persist_config(profile, new_config)
 
 
@@ -62,7 +77,36 @@ def _configure_cli_token(profile, insecure, host, jobs_api_version):
         host = click.prompt(PROMPT_HOST, default=config.host, type=_DbfsHost())
 
     token = click.prompt(PROMPT_TOKEN, default=config.token, hide_input=True)
-    new_config = DatabricksConfig.from_token(host, token, insecure, jobs_api_version)
+    new_config = DatabricksConfig.from_token(host=host,
+                                             token=token,
+                                             refresh_token=None,
+                                             insecure=insecure,
+                                             jobs_api_version=jobs_api_version)
+    update_and_persist_config(profile, new_config)
+
+
+def scope_format(user_input):
+    user_inputs = user_input.split(',')
+    choice = click.Choice(DEFAULT_SCOPES)
+    return list(map(lambda x: choice(x.strip()), user_inputs))
+
+
+def _configure_cli_oauth(profile, insecure, host, scope, jobs_api_version):
+    config = ProfileConfigProvider(profile).get_config() or DatabricksConfig.empty()
+
+    if not host:
+        host = click.prompt(PROMPT_HOST, default=config.host, type=_DbfsHost())
+    if not scope:
+        scope = click.prompt(
+            "Pick one or more comma-separated scopes from {scopes}".format(scopes=DEFAULT_SCOPES),
+            value_proc=scope_format,
+            type=click.Choice(DEFAULT_SCOPES), default=None)
+    access_token, refresh_token = get_tokens(host, scope)
+    new_config = DatabricksConfig.from_token(host=host,
+                                             token=access_token,
+                                             refresh_token=refresh_token,
+                                             insecure=insecure,
+                                             jobs_api_version=jobs_api_version)
     update_and_persist_config(profile, new_config)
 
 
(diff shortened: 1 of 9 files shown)


databricks_cli/oauth/oauth.py
```diff
@@ -0,0 +1,274 @@
...
+
+
+def get_tokens_from_response(oauth_response):
+    access_token = oauth_response['access_token']
+    refresh_token = oauth_response['refresh_token'] if 'refresh_token' in oauth_response else None
+    return access_token, refresh_token
+
+
+def check_and_refresh_access_token(hostname, access_token, refresh_token):
+    expiration_time = datetime.utcnow()
```
stormwindy, line 224: Does this mean if we cannot parse an expiration time, we assume token is expired?

databricks_cli/configure/cli.py
```diff
@@ -110,6 +158,9 @@ def _configure_cli_password(profile, insecure, host, jobs_api_version):
 
 @click.command(context_settings=CONTEXT_SETTINGS,
                short_help='Configures host and authentication info for the CLI.')
+@click.option('--oauth', '-o', 'oauth', show_default=True, is_flag=True, default=False)
+@click.option('--scope', '-s', 'scope', show_default=True, multiple=True,
+              type=click.Choice(DEFAULT_SCOPES))
```
stormwindy, line 163: Does it make sense to add `help` option to click here with a short description?

databricks_cli/configure/provider.py
```diff
@@ -258,34 +261,56 @@ def get_config(self):
...
+                                  refresh_token, insecure, jobs_api_version)
         if config.is_valid:
             return config
         return None
 
 
 class DatabricksConfig(object):
-    def __init__(self, host, username, password, token, insecure, jobs_api_version=None):  # noqa
+    def __init__(self, host, username, password, token,
+                 refresh_token=None, insecure=None, jobs_api_version=None):  # noqa
```
stormwindy, line 276: Is there an exhaustive way of knowing we have changed all the calls to init or `from_token` in the code-base if required? I am concerned if we have not used the PEP standards and specified optional parameters without giving their name. E.g. 
```
def optional_params(one, two=None):
  print(one)
  print(two)

optional_params("required_param", "first optional param")
```

databricks_cli/configure/provider.py
```diff
@@ -28,13 +28,13 @@
 
 from databricks_cli.utils import InvalidConfigurationError
 
-
```
stormwindy, line 31: Is this by mistake? I think PEP8 says there should be 2 lines between imports and any other code blocks.

databricks_cli/oauth/oauth.py
```diff
@@ -0,0 +1,274 @@
...
+    try:
+        decoded = jwt.decode(access_token, options={"verify_signature": False})
```
stormwindy, line 226: Sorry if it is a stupid question, I have limited knowledge of JWT. Isn't setting `verify_signature: False` unadvised? What is the reasoning for setting this?

stormwindy (approved):
Overall it looks good to me. However, I have limited knowledge of JWT. So, I cannot comment much about security aspects of this. I have some nits and question regarding formatting and some JWT usage.

stormwindy:
To add, since this is experimental, I think we can go forwards overall. I can spend some time learning JWT with Python and contribute to the functionality for the full (non-experimental) release.

databricks_cli/oauth/oauth.py
```diff
@@ -0,0 +1,274 @@
...
+        return "UTC"
+
+    def dst(self, dt):
+        #pylint: disable=unused-argument
+        return timedelta(0)
+
+
+# Some contant values
+OIDC_REDIRECTOR_PATH = "oidc"
+CLIENT_ID = "databricks-cli"
```
leidou-db, line 76: For the UI side, each account corresponds to one app id. For cli part, seems we are using one app id to cover all the workspaces. Is my understanding correct?

tests/configure/test_cli.py
```diff
@@ -120,3 +121,11 @@ def test_configure_cli_jobs_api_version_password():
     assert get_config().host == TEST_HOST
     assert get_config().username == TEST_USER
     assert get_config().password == TEST_PASSWORD
+
+
+def test_configure_cli_oauth():
+    runner = CliRunner()
+    runner.invoke(cli.configure_cli, ['--oauth'],
+                  input=(TEST_HOST + '\n' + TEST_SCOPE + '\n'))
+    # This is expected to error because the host doesn't exist and is not
```
leidou-db, line 130: Do we need to catch the exception here so as to verify the exception is thrown as expected?

leidou-db (commented):
Looks good to me in general! Ask some questions mostly for my own education :)

databricks_cli/configure/cli.py
```diff
@@ -38,6 +40,15 @@
 PROMPT_PASSWORD = 'Password'  # NOQA
 PROMPT_TOKEN = 'Token'  # NOQA
 ENV_AAD_TOKEN = 'DATABRICKS_AAD_TOKEN'
+DEFAULT_SCOPES = [
```
Rchakmak, line 43: Nit: I still think these scopes are a little weird to include in the CLI as I'm not sure there is like an API that uses SCIM here (I think it uses the Groups API, which is pretty unintuitive).

databricks_cli/oauth/oauth.py
```diff
@@ -0,0 +1,274 @@
...
+def get_client(client_id=CLIENT_ID):
+    return oauthlib.oauth2.WebApplicationClient(client_id)
+
+
+def get_redirect_url(port=REDIRECT_PORT):
+    return "http://localhost:{port}".format(port=port)
+
+
+def fetch_well_known_config(idp_url):
+    known_config_url = "{idp_url}/.well-known/oauth-authorization-server".format(idp_url=idp_url)
```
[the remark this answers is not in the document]
Rchakmak: It's part of OIDC protocol, so every IDP should have it or they're not in compliance.

Rchakmak (commented):
Left some comments and questions.

databricks_cli/configure/cli.py
sweisdb: Good call. Removed SCIM here.

databricks_cli/configure/provider.py
sweisdb: Fixed.

databricks_cli/oauth/oauth.py
sweisdb: This is correct.

databricks_cli/oauth/oauth.py
sweisdb: We have already verified this access token when we stored it and are just parsing out the expiration time.

This avoids having to fetch the public key from the issuer and perform an unnecessary signature verification.

If it has been tampered with, it'll fail when we replay it back to our API, which has to always verify the signature.

databricks_cli/oauth/oauth.py
```diff
@@ -0,0 +1,274 @@
...
+
+    # Try to refresh using the refresh token
+    click.echo("Attempting to refresh OAuth access token that expired on {expiration_time}"
+               .format(expiration_time=expiration_time))
+    oauth_response = send_refresh_token_request(hostname, refresh_token)
+    fresh_access_token, fresh_refresh_token = get_tokens_from_response(oauth_response)
+    return fresh_access_token, fresh_refresh_token, True
+
+
+def get_tokens(hostname, scope=None):
```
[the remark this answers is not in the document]
sweisdb: The CLI input makes you pick one or more scopes.

It could just be "offline_access" which would let you keep refreshing worthless tokens though.

databricks_cli/oauth/oauth.py
```diff
@@ -0,0 +1,274 @@
...
+            auth_url,
+            redirect_url,
+            scope,
+            state,
+            challenge,
+            REDIRECT_PORT)
+    except OAuth2Error as err:
+        error_and_quit("OAuth Authorization Error: {error}".format(error=err.description))
+
+    token_request_url = oauth_config["token_endpoint"]
```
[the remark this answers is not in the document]
sweisdb: The token endpoint is returned as part of the well known configuration. 

You can see an example here:
https://internal-idp-test.cloud.databricks.com/oidc/.well-known/oauth-authorization-server

PR #402, opened 2021-11-18
Add secure tunneling command to cli

deka108:
### Description
We're adding a new CLI command for secure tunneling to Databricks clusters on select environments. 

This new tool has a dependency with CommandExecution and ExecutionContext, hence we're introducing these services to the CLI as well.

### Testing
We have added the following test cases:
- [x] Tests for CommandExecution and ExecutionContext
- [x] Tests for Tunneling CLI
- [X] Tests for TunnelApi
- [x] Tests for TunnelClient
- [ ] Tests for StreamingServer (will be done on separate PR)
- [ ] Integration tests (will be done on separate PR / platform)
- [x] Manual e2e testing with select environment


ediie726 (approved):

joveyuan-db (commented):
Took an initial pass and left a few comments.

databricks_cli/clusters/cli.py
```diff
@@ -331,6 +333,56 @@ def cluster_events_cli(api_client, cluster_id, start_time, end_time, order, even
         click.echo(tabulate(_cluster_events_to_table(events_json), tablefmt='plain'))
 
 
+def find_free_port():
```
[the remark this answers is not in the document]
deka108: ok. we can move this part and bind it directly and lazily only when it's needed.

databricks_cli/sdk/api_client.py
```diff
@@ -104,14 +105,16 @@ def __init__(self, user=None, password=None, host=None, token=None,
...
+        self.config = config
 
     def close(self):
         """Close the client"""
         pass
 
     # helper functions starting here
 
-    def perform_query(self, method, path, data = {}, headers = None, files=None, version=None):
+    def perform_query(self, method, path, data = {}, headers = None, files=None, version=None,
```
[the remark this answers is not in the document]
deka108: yeah, the files inside sdk are skipped by the linter unfortunately. the files under `sdk` don't conform to pep8.

databricks_cli/sdk/v1_service.py
```diff
@@ -0,0 +1,84 @@
+import copy
```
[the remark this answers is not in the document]
deka108: 👍🏼👍🏼

databricks_cli/tunnel/api.py
```diff
@@ -0,0 +1,184 @@
...
+    def get_org_id(self):
+        resp = self.cluster_client.client.perform_query("GET", "/clusters/get",
+                                                        data={"cluster_id": self.cluster_id},
+                                                        return_raw_response=True)
+        # TODO(tunneling-cli): this assumes a multitenant workspace, we will need to confirm if
+        #  this exists for st workspace
+        return resp.headers["x-databricks-org-id"]
+
+    def create_context(self):
+        response = self.command_client.create_context(language=self.default_language,
```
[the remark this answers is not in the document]
deka108: currently, for simplicity only we're using python, the initial reason for the refactor was to standardize how `python` lang is defined in the cmd exec API. we're only using it twice though, so good point

tests/clusters/test_cli.py
```diff
@@ -130,6 +140,25 @@ def test_get_cli(cluster_api_mock):
...
+
+    @provide_conf
+    def test_tunnel_cli_with_cluster_id(self, tunnel_api_mock):
+        # with cluster id
+        self.runner.invoke(cli.tunnel_cli, ['--cluster-id', CLUSTER_ID])
+        assert tunnel_api_mock.start_tunneling.call_count == 1
+
+    def test_tunnel_cli_without_required_args(self):
+        res = self.runner.invoke(cli.tunnel_cli, [])
+        assert_cli_output(res.output,
```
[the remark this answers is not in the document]
deka108: yeah, this exception is caught at the decorator-level during cli invocation so we won't get that in this cli test. the one that we thrown is after it passes the cli validation checks.

databricks_cli/tunnel/server.py
```diff
@@ -0,0 +1,142 @@
...
+        await self.main_routine(stream)
+
+    async def main_routine(self, stream):
+        logger.info("Launching main routine for socket with id: %s", self.socket_id)
+        await self.client.connect()
+        while True:
+            try:
+                # read input stream from the client and emit this to socketio server
+                data = await stream.read_bytes(1024 * 50, partial=True)
+                await asyncio.sleep(0.05)
```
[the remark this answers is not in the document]
deka108: 👀 initially i thought it's necessary for coroutine progress / sequential ordering but actually let me check on that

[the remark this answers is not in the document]
deka108: cc: @marcom-db

databricks_cli/tunnel/server.py
```diff
@@ -0,0 +1,142 @@
...
+            logger.info("Connection to the socket.io server is not yet established")
+            await asyncio.sleep(1)
+
+    def is_connected(self):
+        return self.io.connected
+
+    async def send_data(self, data):
+        await self.io.emit("data_received", {"data": data})
+
+    async def disconnect(self):
```
deka108, line 95: @marcom-db, do we need to add an event handler for `disconnect` that sets `connection_established` to `False` as well?

databricks_cli/tunnel/api.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def send_public_key_to_driver(self, public_key):
+        print("Sending public key to the driver...")
+        public_key_str = public_key.decode('utf-8')
+        install_keys_on_driver_cmd = f"""
+from pathlib import Path
+
+Path("~/.ssh").expanduser().mkdir(exist_ok=True)
+Path("~/.ssh/id_rsa.pub").expanduser().write_text("{public_key_str}")
+Path("~/.ssh/authorized_keys").expanduser().write_text("{public_key_str}")
```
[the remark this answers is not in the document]
deka108: ok this is a bit tricky. is there any use case (outside of tunneling), that already modifies `authorized_keys`. if yes, then we'll need to  change the way we write to the file

databricks_cli/tunnel/api.py
```diff
@@ -0,0 +1,184 @@
...
+                                                context_id=self.context_id)
+
+    def send_public_key_to_driver(self, public_key):
+        print("Sending public key to the driver...")
+        public_key_str = public_key.decode('utf-8')
+        install_keys_on_driver_cmd = f"""
+from pathlib import Path
+
+Path("~/.ssh").expanduser().mkdir(exist_ok=True)
+Path("~/.ssh/id_rsa.pub").expanduser().write_text("{public_key_str}")
```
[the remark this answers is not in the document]
deka108: yep turned out we can

databricks_cli/clusters/cli.py
```diff
@@ -331,6 +331,44 @@ def cluster_events_cli(api_client, cluster_id, start_time, end_time, order, even
...
+@debug_option
+@profile_option
+@eat_exceptions
+@provide_api_client
+def tunnel_cli(api_client, cluster_id, cluster_name, local_port):
+    """
+    Start a secure TCP tunnel to a cluster over Databricks' identity proxy.
+    """
+    if sys.version_info < (3, 6):
+        raise RuntimeError("The tunneling command is not supported on Python version < 3.6")
```
marcom-db, line 349: Isn't the python version tied to the CLI version? If so can we give them instructions to change CLI versions

databricks_cli/tunnel/server.py
marcom-db: yeah, we can set it to false here if we disconnect

databricks_cli/tunnel/server.py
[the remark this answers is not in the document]
marcom-db: It may have been put in during early prototyping. Let's remove it and confirm there are no issues

databricks_cli/sdk/v1_service.py
```diff
@@ -0,0 +1,106 @@
...
+    def __init__(self, client):
+        self.v1_client = copy.deepcopy(client)
+        self.v1_client.api_version = OLD_API_VERSION
+
+
+def command_is_not_terminated(resp):
+    return resp["status"] not in {"Finished", "Cancelled", "Error"}
+
+
+class CommandExecutionService(APIV1Client):
```
jaipreet-s, line 40: The pattern I see in the v2 APIs is that there is a `Service` object for each API category. `context` and `commands` are two separate categories in API v1.2, do you think they should be separate `Services`? Or this might self-resolve if we're able to auto-generate the service objects for v1.2

databricks_cli/tunnel/api.py
```diff
@@ -0,0 +1,226 @@
...
+    private_key = key.private_bytes(
+        serialization.Encoding.PEM,  # noqa
+        serialization.PrivateFormat.TraditionalOpenSSL,  # noqa
+        serialization.NoEncryption()
+    )
+
+    public_key = key.public_key().public_bytes(
+        serialization.Encoding.OpenSSH,  # noqa
+        serialization.PublicFormat.OpenSSH  # noqa
+    )
```
jaipreet-s, line 54: Are these the same parameters that `ssh-keygen` would use when generating a key pair? For e.g., GitHub's recommendation is `ssh-keygen -t rsa -b 4096 ...` so we want to make sure we're using the recommended defaults here

databricks_cli/tunnel/server.py
```diff
@@ -0,0 +1,164 @@
...
+                }
+                await self.io.connect(self.tunnel_config.endpoint_url,
+                                      auth=None,
+                                      transports=["websocket"],
+                                      socketio_path=self.tunnel_config.socket_path,
+                                      headers=headers)
+            except socketio.exceptions.ConnectionError as e:
+                logger.error("Connection exception:")
+                logger.error(e)
+                raise e
```
jaipreet-s, line 105: nit: You can just use `logger.exception` here. See https://www.loggly.com/blog/exceptional-logging-of-exceptions-in-python/

databricks_cli/tunnel/server.py
```diff
@@ -0,0 +1,164 @@
...
+    def socket_path(self):
+        return f"{self.endpoint_path}socket.io"
+
+
+class TunnelClient:
+    def __init__(self, stream: IOStream, session: aiohttp.ClientSession,
+                 tunnel_config: TunnelConfig, debug=False):
+        self.stream = stream
+        self.tunnel_config = tunnel_config
+        self.io = socketio.AsyncClient(http_session=session, logger=debug, engineio_logger=debug)
```
jaipreet-s, line 73: nit: OK to start with `debug` logging, but let's have a TODO to remove this

setup.py
```diff
@@ -24,22 +24,37 @@
...
+if sys.version_info < (3, 6):
+    install_requires += ['configparser>=0.3.5;python_version < "3.6"']
+else:
+    install_requires += [
+        # additional deps for tunneling
+        'aiohttp>=3.7.4',
+        'cryptography>=3.4.6',
+        'python-socketio>=5.0.4',
+        'tenacity>=7.0.0',
+        'tornado>=6.0.1',
```
jaipreet-s, line 50: It's generally a good practice to pin these to the major version, for e.g., <7 for tornado to avoid getting breaking changes

jaipreet-s (commented):
First pass, would like to do another once the initial feedback is addressed 😃

databricks_cli/clusters/cli.py
deka108: That is true if the CLI is using only one python version. However, the current CLI version supports both python `2.7` and `3.6`, but the tunneling min. req is using `3.6`. So we need this check in case the user is using python2.

deka108: i'm not sure what the timeline is for the cli to only support python3 onwards.

databricks_cli/tunnel/server.py
deka108: i think this is important for debugging problems. currently it receives a boolean flag `--debug` from the cli. if it's true then the tunneling loggings will be shown. but currently, `--debug` flag is shared with other cli cmds. i wonder if we should surface new debug flag for tunneling

setup.py
deka108: hm.. pinning is a little tricky cos if we're being too restrictive then it might break user workflow (eg: user having different pkg dep versions in their env). so it'd be good to use a good deps version range that don't break the deps for the cli while also not being too restrictive.

databricks_cli/sdk/v1_service.py
deka108: cc: @andrewmchen

databricks_cli/tunnel/api.py
deka108: no... another thing that i notice is we don't have the user and the hostname. these aren't required, but usually there when we generate ssh keys. cc: @marcom-db

setup.py
jaipreet-s: Yea, I'm not suggested we pin to the patch version, just to the next major version. For e.g., `tornado>=6.01 < 7`

databricks_cli/tunnel/server.py
jaipreet-s: If there is a `--debug` flag that toggles the debug logging off/on, that's actually perfect.

databricks_cli/clusters/cli.py
```diff
@@ -331,6 +331,44 @@ def cluster_events_cli(api_client, cluster_id, start_time, end_time, order, even
...
+    if not api_client.config:
+        raise RuntimeError("Unable to initialize client."
+                           "Please check if the credentials are configured properly.")
```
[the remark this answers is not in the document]
deka108: oh yeah, this isn't needed, but since we require pat (require to create tunnel session), I'll modify api_client to store this value and then check it in the tunneling cli.

databricks_cli/sdk/api_client.py
```diff
@@ -68,7 +68,8 @@ class ApiClient(object):
     to be used by different versions of the client.
     """
     def __init__(self, user=None, password=None, host=None, token=None,
-                 api_version=version.API_VERSION, default_headers={}, verify=True, command_name="", jobs_api_version=None):
+                 api_version=version.API_VERSION, default_headers={}, verify=True, command_name="",
+                 jobs_api_version=None, config=None):
```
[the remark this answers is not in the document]
deka108: it's not required for `ApiClient`. but i'll modify `ApiClient` to store `host` and `token` which are used by `TunnelingClient` and `StreamingServer`

databricks_cli/sdk/v1_service.py
```diff
@@ -0,0 +1,106 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import copy
+
+from tenacity import retry, stop_after_attempt, wait_random_exponential, retry_if_result
+
+from databricks_cli.sdk.version import OLD_API_VERSION
+
+
+class APIV1Client(object):
```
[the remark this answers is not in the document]
deka108: yeah, i skipped the general hierarchy and combined the service + api for the cmd exec service because they can't be auto-generated... I'll try to separate the APIs (cmd and context), and low-level and business logic into different hierarchies and entities

databricks_cli/tunnel/server.py
jaipreet-s: I think you can log-and-raise here if the intention is to break the loop

setup.py
deka108: ok, i think we have two options here:
1. pin the upper limit to the next major version
2. pin the upper limit based on request / when we discover some version is not compatible

for 2, i wonder f it's possible that the user is using the next major version that is still compatible but not sure how common it is.

databricks_cli/tunnel/api.py
```diff
@@ -0,0 +1,235 @@
...
+            if sockets:
+                server.add_sockets(sockets)
+                for s in sockets:
+                    local_port = s.getsockname()[1]
+                    if local_port is not None:
+                        break
+            if local_port is None:
+                raise RuntimeError("Could not find a free port")
+        print(f"Starting local tunneling on localhost:{local_port}")
+        print("Please use the following command for ssh connection:")
```
jaipreet-s, line 233:
```suggestion
        print("Please use the following command for SSH connection:")
```

setup.py
jaipreet-s: I prefer 2), because more often that not things can break unexpectedly across major versions.

databricks_cli/tunnel/api.py
```diff
@@ -0,0 +1,235 @@
...
+
+    private_key = key.private_bytes(
+        encoding=serialization.Encoding.PEM,
+        format=serialization.PrivateFormat.OpenSSH,
+        encryption_algorithm=serialization.NoEncryption()
+    )
+
+    public_key = key.public_key().public_bytes(
+        encoding=serialization.Encoding.OpenSSH,
+        format=serialization.PublicFormat.OpenSSH
```
jaipreet-s, line 52: While we confirm the correct parameters to use here, recommend adding a TODO

jaipreet-s:
LGTM % the setup.py pinning and confirming the SSH key parameters

jaipreet-s (approved):

databricks_cli/clusters/cli.py
```diff
@@ -331,6 +331,41 @@ def cluster_events_cli(api_client, cluster_id, start_time, end_time, order, even
         click.echo(tabulate(_cluster_events_to_table(events_json), tablefmt='plain'))
 
 
+@click.command(context_settings=CONTEXT_SETTINGS)
+@click.option('--cluster-id', cls=OneOfOption, one_of=CLUSTER_OPTIONS,
+              type=ClusterIdClickType(), default=None, help=ClusterIdClickType.help)
+@click.option('--cluster-name', cls=OneOfOption, one_of=CLUSTER_OPTIONS,
```
joveyuan-db, line 337: Do you know if this is the standard way to define help for cluster name?  It's slightly confusing since it would indicate that users have to use cluster ID.

databricks_cli/tunnel/api.py
```diff
@@ -0,0 +1,236 @@
...
+import tornado
+from cryptography.hazmat.primitives.asymmetric import rsa
+from tornado.ioloop import IOLoop
+
+from databricks_cli.clusters.api import ClusterApi
+from databricks_cli.commands.api import CommandApi, ExecutionContextApi
+from databricks_cli.tunnel.server import TunnelConfig, StreamingServer
+
+
+def generate_key_pair():
```
joveyuan-db, line 36: nit: Can we add a comment here that this helper function returns (private key, public key) in case we switch the order at some point :)

databricks_cli/tunnel/api.py
```diff
@@ -0,0 +1,236 @@
...
+    return private_key, public_key
+
+
+class TunnelApi:
+    def __init__(self, api_client, debug=False):
+        self.cluster_client = ClusterApi(api_client)
+        self.command_client = CommandApi(api_client)
+        self.exec_ctx_client = ExecutionContextApi(api_client)
+        self._api_client = api_client
+        self._default_local_ssh_dir = "~/.ssh"
```
joveyuan-db, line 65: Just curious, this might not support Windows?  Is that OK?

databricks_cli/tunnel/api.py
```diff
@@ -0,0 +1,236 @@
...
+            execute_command_until_terminated(language="python",
+                                             cluster_id=self.cluster_id,
+                                             context_id=self.context_id,
+                                             command=install_keys_on_driver_cmd)
+        if resp["status"] != "Finished":
+            err_data = resp.get("results", {}).get("data", "")
+            raise RuntimeError(f"Sending public key to driver was not successful: {err_data}")
+        if resp["results"]["resultType"] == "error":
+            err_data = resp.get("results", {}).get("cause", "")
+            raise RuntimeError(f"Sending public key to driver was not successful: {err_data})")
```
joveyuan-db, line 129: nit: extra parentheses
```suggestion
            raise RuntimeError(f"Sending public key to driver was not successful: {err_data}")
```

databricks_cli/tunnel/api.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def setup_ssh_keys(self):
+        print("Setting up SSH keys...")
+
+        private_key, public_key = generate_key_pair()
+        self.send_public_key_to_driver(public_key)
+        self.setup_local_keys(private_key, public_key)
+
+    def setup_local_keys(self, private_key, public_key):
+        private_key_path = Path(f"{self.local_private_key_path}").expanduser()
```
joveyuan-db, line 163: nit: no need for f strings (I think?).  Also applies to lines 171 and 179

databricks_cli/tunnel/api.py
```diff
@@ -0,0 +1,236 @@
...
+            self.start_local_server(local_port)
+        finally:
+            self.cleanup()
+
+    def start_local_server(self, local_port=None):
+        server = StreamingServer(self.tunnel_config, debug=self.debug)
+        if local_port:
+            server.listen(local_port)
+        else:
+            sockets = tornado.netutil.bind_sockets(0, '')
```
joveyuan-db, line 224: Can you add a comment here that we're finding a random free port?

databricks_cli/tunnel/api.py
```diff
@@ -0,0 +1,236 @@
...
+                server.add_sockets(sockets)
+                for s in sockets:
+                    local_port = s.getsockname()[1]
+                    if local_port is not None:
+                        break
+            if local_port is None:
+                raise RuntimeError("Could not find a free port")
+        print(f"Starting local tunneling on localhost:{local_port}")
+        print("Please use the following command for SSH connection:")
+        print(f"ssh root@localhost -p {local_port} -i {self.local_private_key_path}")
```
joveyuan-db, line 235: Random, but I wonder whether we should use ANSI to print this line red or some other way to make it stand out.

joveyuan-db (commented):
Minor comments, otherwise looks good!

databricks_cli/clusters/cli.py
deka108: yeah.. that's the standard help for cluster name on cluster cli. it's wrong, but we can try to fix it for all clusters cli on a separate PR.

databricks_cli/tunnel/api.py
deka108: good point. cc: @jaipreet-s 

I think Windows users might be able to use WSL / OpenSSH? Or in the worst case, they might have to use docker container running a unix-based system...

Currently, in the context of the tunneling cli, what determines whether windows users can use it:
1. Can these keys be generated on windows?
2. Can the ssh command on Windows find the default ssh key location?

Related to 1, we're currently using Python's cryptography lib and std `Pathlib`. As long as cryptography lib is compatible with Windows and Python's `Pathlib` + path that we're using is abstract enough and compatible s.t. it works with Windows filesystem, then this req is satisfied. [This](https://stackoverflow.com/questions/4028904/what-is-the-correct-cross-platform-way-to-get-the-home-directory-in-python) and [this](https://stackoverflow.com/questions/2953834/windows-path-in-python) suggests that our implementation might work cross-platform.

2 needs some investigation. Although [this doc](https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_keymanagement) suggests it might be possible.

databricks_cli/tunnel/api.py
deka108: yeah, i was thinking of prettifying the logs... let's file followup tickets

databricks_cli/clusters/cli.py
```diff
@@ -331,6 +331,41 @@ def cluster_events_cli(api_client, cluster_id, start_time, end_time, order, even
...
+    elif cluster_name:
+        cluster = ClusterApi(api_client).get_cluster_by_name(cluster_name)
+        cluster_id = cluster["cluster_id"]
+    else:
+        raise RuntimeError('cluster_name and cluster_id must not be empty!')
+
+    # TODO(tunneling-cli): move this up once we support python3 only
+    from databricks_cli.tunnel.api import TunnelApi
+
+    print("Starting a secure tunnel to cluster with ID: {}...".format(cluster_id))
```
andrewmchen, line 365: nit: I think we use click.echo other places.

databricks_cli/sdk/v1_service.py
```diff
@@ -0,0 +1,87 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import copy
+
+from databricks_cli.sdk.version import OLD_API_VERSION
+
+
+class APIV1Client:
```
andrewmchen, line 28: Should we use "new-style" classes here? https://stackoverflow.com/questions/4015417/why-do-python-classes-inherit-object

databricks_cli/tunnel/api.py
```diff
@@ -0,0 +1,242 @@
...
+        if status != "working":
+            raise RuntimeError(f"Unexpected remote tunnel server status: {status}")
+
+        return True
+
+    def cleanup(self):
+        self.destroy_context()
+        self._init_params()
+
+    def start_tunneling(self, cluster_id, local_port=None):
```
andrewmchen, line 198: Why are `cluster_id` and `local_port` not provided as arguments for the constructor of `TunnelApi`?

databricks_cli/tunnel/api.py
```diff
@@ -0,0 +1,242 @@
...
+        print("Ping the remote tunnel server...")
+        resp = requests.get(f"{self.tunnel_config.healthcheck_url}",
+                            headers=self._api_client.default_headers)
+        if resp.status_code != 200:
+            raise RuntimeError(f"Unable to connect to the remote tunnel server: {resp.status_code}")
+        status = resp.json()["status"]
+        if status != "working":
+            raise RuntimeError(f"Unexpected remote tunnel server status: {status}")
+
+        return True
```
andrewmchen, line 192: Do we need to return True here?

databricks_cli/tunnel/api.py
andrewmchen: click.echo can help with this IIRC,

andrewmchen (approved):

databricks_cli/sdk/v1_service.py
deka108: By new style do you mean using `(object)`? It seems that this matters if we want to maintain compatibility with python 2. So in Python3 `object` is implicitly inherited by default.

databricks_cli/tunnel/api.py
deka108: no

databricks_cli/tunnel/api.py
deka108: Moving `cluster_id` to TunnelApi makes sense. But having `local_port` in `start_tunneling` can be helpful in case we want to reuse TunnelApi to create multiple tunnel connections with different ports

databricks_cli/tunnel/api.py
deka108: The output generated by `click.echo` is still similar as `print` actually, cos we haven't really added the ansi coloring. We'll file follow up tickets for this

PR #412, opened 2021-12-28
Better support for authentication with AAD tokens

alexott:
This PR adds support for authentication using AAD tokens that could be obtained by different means:

* using azure-cli's `az account get-access-token`
* Service Principal - with client ID & secret
* azure managed identity assigned to Azure VM

Support for AAD tokens enables use of databricks-cli in organizations that don't permit PATs

### DIFF
--- a/databricks_cli/cli.py
+++ b/databricks_cli/cli.py
@@ -23,7 +23,7 @@
 
 import click
 
-from databricks_cli.configure.config import profile_option, debug_option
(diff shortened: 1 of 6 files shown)


savbace:
Do we expect the PR to be merged? I faced with a need to authenticate using Service Principal too.

nfx:
We'll have a proper Databricks Python SDK that will support AAD auth, can't share the timelines. Stay tuned :)

PR #417, opened 2022-01-14
[ES-244938] Fix pipelines list pagination

dylanstee-db:
The `pagination` request and response property is now deprecated. Instead, the nested properties from inside `pagination` have been elevated to the top level (where `pagination` used to reside). This PR fixes the pagination-related property access so that more than one page of pipelines are returned.

### Before

```bash
$ databricks pipelines list | jq '. | length'              
25
```

### After

```bash
$ databricks pipelines list | jq '. | length'              
311
```

### DIFF
--- a/databricks_cli/pipelines/api.py
+++ b/databricks_cli/pipelines/api.py
@@ -63,11 +63,11 @@ class PipelinesApi(object):
         def call(page_token=None, max_results=None, order_by=None):
             _data = {}
             if page_token:
-                _data["pagination.page_token"] = page_token
+                _data["page_token"] = page_token
             if max_results:
-                _data["pagination.max_results"] = max_results
+                _data["max_results"] = max_results
             if order_by:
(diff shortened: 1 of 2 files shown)


mukulmurthy (approved):
LGTM, have you done a manual test with this against dogfood / some real env?

dylanstee-db:
> have you done a manual test with this against dogfood / some real env?

Yep, the commands in the description are run against e2-dogfood.

dylanstee-db:
@mukulmurthy, do you have permission to merge the PR? I don't have write access to the repo so I can't merge this.

dylanstee-db:
Confirmed that the CLI works on Python v2.7

Screen Shot 2022-01-19 at 3 48 47 PM

mukulmurthy:
Looks like the Python 2.7 build is broken (even on master), but the 3.6 one passed. Sync'd with Deka who gave us permission to merge this.

PR #420, opened 2022-01-31
[SC-92558] Deprecate "run/reset" commands in favor of start which uses StartUpdate endpoint

pradeepgv-db:
* Deprecate run/reset commands in favor of start command (run -> start, reset -> start --full-refresh=true).
* Update run/reset/start commands to use StartUpdate endpoint on deployment server, eventually we want to remove
   run/reset endpoints on deployment server.

Testing
------
* Unit tests
* Manualy testing of CLI against a test shard that start(with and without full-refresh)/reset/run trigger an update properly with correct params


pradeepgv-db:
(ve) ~/databricks-cli (SC-92558) $ databricks pipelines -h                                                                          
Usage: databricks pipelines [OPTIONS] COMMAND [ARGS]...

  Utility to interact with the Databricks pipelines.

Options:
  -v, --version   0.16.3.dev0
  --debug         Debug Mode. Shows full stack trace on error.
  --profile TEXT  CLI connection profile to use. The default profile is
                  "DEFAULT".
  -h, --help      Show this message and exit.

Commands:
  delete  Stops a delta pipeline and deletes its associated Databricks
          resources
  deploy  Deploys a delta pipeline according to the pipeline specification
  get     Gets a delta pipeline's current spec and status
  list    Gets a delta pipeline's current spec and status
  reset   (Deprecated) Resets a delta pipeline so data can be reprocessed from
          scratch
  run     (Deprecated) Starts the execution of a delta pipeline run
  start   Starts the execution of a delta pipeline run
  stop    Stops the execution of a delta pipeline run



(ve) ~/databricks-cli (SC-92558) $  databricks pipelines reset -h
Usage: databricks pipelines reset [OPTIONS]

  (Deprecated) "reset" command is deprecated, please use "start --full-
  refresh" command instead.

  Resets a delta pipeline by truncating tables and creating new checkpoint
  folders so data is reprocessed from scratch.


(ve) ~/databricks-cli (SC-92558) $  databricks pipelines run -h  
Usage: databricks pipelines run [OPTIONS]

  (Deprecated) "run" command is deprecated, please use "start" command
  instead.

  Starts the execution of a delta pipelines run by starting the cluster and
  processing data.


(ve) ~/databricks-cli (SC-92558) $  databricks pipelines start -h
Usage: databricks pipelines start [OPTIONS]

  Starts the execution of a delta pipelines run by starting the cluster and
  processing data.

pradeepgv-db:
@anew @mukulmurthy Don't have permissions to add reviewers, tagging you both  to help me with the review.

pradeepgv-db:
@dylanstee-db Maybe you can help me with review too, since you worked on this repo previously.

anew:
Please ask Jonathan Seidman to review the messages. Or ask him who is in charge of the CLI in the docs team. 
I could not figure out what his user id is in this repo.

databricks_cli/sdk/service.py
```diff
@@ -1058,6 +1058,14 @@ def run(self, pipeline_id=None, headers=None):
         return self.client.perform_query('POST', '/pipelines/{pipeline_id}/run'.format(pipeline_id=pipeline_id),
                                          data=_data, headers=headers)
 
+    def start(self, pipeline_id=None, full_refresh=False, headers=None):
+        _data = {}
+        if full_refresh:
+            _data['full_refresh'] = full_refresh
+        _data['cause'] = 'API_CALL'
```
anew, line 1065: @calving-db this is now `USER_ACTION`, correct?

databricks_cli/pipelines/cli.py
```diff
@@ -208,17 +213,45 @@ def reset_cli(api_client, pipeline_id):
...
+def start_cli(api_client, pipeline_id, full_refresh):
+    """
+    Starts the execution of a delta pipelines run by starting the cluster and processing data.
+
+    Usage:
+
+    databricks pipelines start --pipeline-id 1234 --full-refresh=(false|true)
+    """
+    _validate_pipeline_id(pipeline_id)
+    PipelinesApi(api_client).start(pipeline_id, full_refresh=full_refresh)
```
dylanstee-db, line 251: Can this fail? If so, how do the error messages look?

databricks_cli/pipelines/cli.py
```diff
@@ -208,17 +213,45 @@ def reset_cli(api_client, pipeline_id):
...
+    click.echo("Update triggered for pipeline {}".format(pipeline_id))
```
dylanstee-db, line 252: Maybe something like "Started an update for pipeline ..."?

tests/pipelines/test_api.py
```diff
@@ -226,20 +226,27 @@ def test_get(pipelines_api):
     assert (response['pipeline_id'] == PIPELINE_ID and response['state'] == 'RUNNING')
 
 
-def test_reset(pipelines_api):
```
dylanstee-db, line 229: I think it might be good to keep the `test_reset` and `test_run` tests until the run and reset APIs are fully removed.

dylanstee-db (commented):
Nice tests and deprecation messages!

pradeepgv-db:
Adding @jaseidman and @PaulCornellDB to review the messages

databricks_cli/sdk/service.py
calving-db: Yes

databricks_cli/pipelines/cli.py
```diff
@@ -186,20 +187,24 @@ def list_cli(api_client):
 @provide_api_client
 def reset_cli(api_client, pipeline_id):
     """
+    (Deprecated) "reset" command is deprecated, please use "start --full-refresh" command instead.
```
jaseidman, line 190:
```suggestion
    (Deprecated) Use the "start --full-refresh" command.
```

I think we can make this short and to the point, but let me know what you think.

Also, a naive question: do we have a pattern for documenting deprecated commands? For example, it seems like elsewhere we still include the command description along with the recommended alternative. Does it make sense to include the deprecation admonition, but still include the command description below?

databricks_cli/pipelines/cli.py
```diff
@@ -208,17 +213,45 @@ def reset_cli(api_client, pipeline_id):
 @provide_api_client
 def run_cli(api_client, pipeline_id):
     """
+    (Deprecated) "run" command is deprecated, please use "start" command instead.
```
jaseidman, line 216:
```suggestion
    (Deprecated) Use the "start" command instead.
```

databricks_cli/pipelines/cli.py
jaseidman: And also, (Deprecated) or [Deprecated]? Here again, I'm not sure if there's a standard.

databricks_cli/pipelines/cli.py
```diff
@@ -208,17 +213,45 @@ def reset_cli(api_client, pipeline_id):
...
+@click.option('--full-refresh', default=False, type=bool,
+              help='If true, truncates tables and creates new checkpoint' +
+                   ' folders so data is reprocessed from scratch')
+@debug_option
+@profile_option
+@pipelines_exception_eater
+@provide_api_client
+def start_cli(api_client, pipeline_id, full_refresh):
+    """
+    Starts the execution of a delta pipelines run by starting the cluster and processing data.
```
jaseidman, line 244: Ultimately I think this is a decision for engineering and product, but I'll just note that we seem to use different terminology here then is used elsewhere, for example the API uses the "update" request to trigger a pipeline run. "Update" is also consistent with terminology used elsewhere in our DLT documentation.

Also, I think "delta pipelines" is a reference to the old name? Should this be "a Delta Live Tables pipeline run", or just "a pipeline run"?

databricks_cli/pipelines/cli.py
jaseidman: So we do use "update" here, which would seem to add to the argument for using consistent terminology.

jaseidman (commented):
Some suggestions/comments.

databricks_cli/pipelines/cli.py
anew: I don't know about a standard. I was hoping our tech writer would know :)

anew: Looks like this is the only other place in the CLI where somethings deprecated:
https://github.com/databricks/databricks-cli/blob/25520f0b7a0ef8e37d318c50483a4f70393b6a08/databricks_cli/configure/provider.py#L129
```
def get_config_for_profile(profile):
    """
    [Deprecated] Reads from the filesystem and gets a DatabricksConfig for the
    specified profile. If it does not exist, then return a DatabricksConfig with fields set
    to None.
```
I agree with your suggestion of using square brackets.

databricks_cli/pipelines/cli.py
anew: I like `update`. Should have thought of this earlier.

databricks_cli/sdk/service.py
pradeepgv-db: updated to USER_ACTION, do we need to be careful about compatibility issue between cli and the databricks backend version (since this is a new enum that we are introducing)?

databricks_cli/pipelines/cli.py
pradeepgv-db: If the interface itself is not reachable, error looks something like this

`Error: b'<html>\n<head>\n<meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>\n<title>Error 403 Invalid access token.</title>\n</head>\n<body><h2>HTTP ERROR 403</h2>\n<p>Problem accessing /api/2.0/pipelines/{pipeline_id}/start. Reason:\n<pre>    Invalid access token.</pre></p>\n</body>\n</html>\n'
`
Error from backend looks like this:
```
Error: RESOURCE_DOES_NOT_EXIST
The specified pipeline {pipeline_id} was not found.

```

I believe pattern is Error: <error thrown either from backend or the library>

databricks_cli/pipelines/cli.py
pradeepgv-db: updated to "Started an update .."

Good point, switched the command visible to the user as "update" instead of "start"

tests/pipelines/test_api.py
pradeepgv-db: I don't have these methods anymore, cli.py directly calls start_update now. So tests are around start_udpate.

databricks_cli/pipelines/cli.py
pradeepgv-db: Changed to square brackets. Yeah, I couldn't find other examples, click.option has a deprecated param but the error message is generic and didn't have a easy way to modify that at the moment.

Let me know how the error messages look like now.

@jaseidman what do you mean by "deprecation admonition"?

databricks_cli/pipelines/cli.py
pradeepgv-db: updated

databricks_cli/pipelines/cli.py
pradeepgv-db: changed the command to "update".

Also, updated "delta pipelines run" to "a pipeline run" in most places in the file.

databricks_cli/pipelines/cli.py
```diff
@@ -49,7 +49,7 @@
 
 
 @click.command(context_settings=CONTEXT_SETTINGS,
-               short_help='Deploys a delta pipeline according to the pipeline specification')
+               short_help='Deploys a pipeline according to the pipeline specification')
```
PaulCornellDB, line 52: All descriptions should use proper punctuation (for example, end a sentence with a period), here and throughout.

databricks_cli/pipelines/cli.py
```diff
@@ -177,7 +177,7 @@ def list_cli(api_client):
 
 
 @click.command(context_settings=CONTEXT_SETTINGS,
-               short_help='Resets a delta pipeline so data can be reprocessed from scratch')
+               short_help='[Deprecated] Resets a pipeline so data can be reprocessed from scratch')
```
PaulCornellDB, line 180: For better internationalization:
"so data" -> "so that data"
"scratch" -> "from the beginning"

databricks_cli/pipelines/cli.py
```diff
@@ -186,20 +186,24 @@ def list_cli(api_client):
 @provide_api_client
 def reset_cli(api_client, pipeline_id):
     """
-    Resets a delta pipeline by truncating tables and creating new checkpoint folders so data is
+    [Deprecated] Use the "update --full-refresh" command instead.
+
+    Resets a pipeline by truncating tables and creating new checkpoint folders so data is
```
PaulCornellDB, line 191: "so data" -> "so that data"

databricks_cli/pipelines/cli.py
```diff
@@ -186,20 +186,24 @@ def list_cli(api_client):
...
     reprocessed from scratch.
 
     Usage:
 
     databricks pipelines reset --pipeline-id 1234
     """
+    click.echo("DeprecationWarning: \"reset\" command is deprecated, " +
```
PaulCornellDB, line 198: For internationalization and style:
"reset" -> "the reset"
Omit "please"

databricks_cli/pipelines/cli.py
```diff
@@ -186,20 +186,24 @@ def list_cli(api_client):
...
+               "please use \"start --full-refresh\" command instead")
     _validate_pipeline_id(pipeline_id)
-    PipelinesApi(api_client).reset(pipeline_id)
+    PipelinesApi(api_client).start_update(pipeline_id, full_refresh=True)
     click.echo("Reset triggered for pipeline {}".format(pipeline_id))
 
 
 @click.command(context_settings=CONTEXT_SETTINGS,
-               short_help='Starts the execution of a delta pipeline run')
+               short_help='[Deprecated] Starts the execution of a pipeline run')
```
PaulCornellDB, line 206: If possible, try to stay away from the word "execution." For example, "Starts a pipeline run."

databricks_cli/pipelines/cli.py
```diff
@@ -208,19 +212,47 @@ def reset_cli(api_client, pipeline_id):
...
 
 
 @click.command(context_settings=CONTEXT_SETTINGS,
-               short_help='Stops the execution of a delta pipeline run')
+               short_help='Starts the execution of a pipeline run')
+@click.option('--pipeline-id', default=None, type=PipelineIdClickType(),
+              help=PipelineIdClickType.help)
+@click.option('--full-refresh', default=False, type=bool,
+              help='If true, truncates tables and creates new checkpoint' +
+                   ' folders so data is reprocessed from scratch')
```
PaulCornellDB, line 236: "so data" -> "so that data"

PaulCornellDB (commented):
See my comments.

databricks_cli/pipelines/cli.py
pradeepgv-db: done, ptal

databricks_cli/pipelines/cli.py
pradeepgv-db: you mean <"reset" command> to <the "reset" command> right? 
ptal

databricks_cli/pipelines/cli.py
pradeepgv-db: done

databricks_cli/pipelines/cli.py
pradeepgv-db: done

databricks_cli/pipelines/cli.py
```diff
@@ -62,9 +62,9 @@
 @provide_api_client
 def deploy_cli(api_client, spec_arg, spec, allow_duplicate_names, pipeline_id):
     """
-    Deploys a delta pipeline according to the pipeline specification. The pipeline spec is a
-    specification that explains how to run a Delta Pipeline on Databricks. All local libraries
-    referenced in the spec are uploaded to DBFS.
+    Deploys a pipeline according to the pipeline specification. The pipeline spec is a
+    specification that explains how to run a Delta Live Tables pipeline on Databricks.
```
jaseidman, line 66:
```suggestion
    JSON document that defines the required settings to run a Delta Live Tables pipeline on Databricks.
```

Just a suggestion to remove the repeated use of "specification" here.

jaseidman (approved):
One more small comment. Otherwise, LGTM.

databricks_cli/pipelines/cli.py
pradeepgv-db: @jaseidman we had a long conversation w.r.t to whether naming the command to start an update as "start", "update", "start update". For now, we decided to go with start as it makes sense in both triggered update and continuous update case and it keeps the symmetry between start/stop. 

I will revert the naming from update => start

databricks_cli/pipelines/cli.py
```diff
@@ -49,11 +49,11 @@
 
 
 @click.command(context_settings=CONTEXT_SETTINGS,
-               short_help='Deploys a delta pipeline according to the pipeline specification')
+               short_help='Deploys a pipeline according to the pipeline specification.')
 @click.argument('spec_arg', default=None, required=False)
 @click.option('--spec', default=None, type=PipelineSpecClickType(), help=PipelineSpecClickType.help)
 @click.option('--allow-duplicate-names', is_flag=True,
-              help="Skip duplicate name check while deploying pipeline")
+              help="Skip duplicate name check while deploying pipeline.")
```
PaulCornellDB, line 56:
```suggestion
              help="Skips duplicate name checking while deploying the pipeline.")
```

databricks_cli/pipelines/cli.py
```diff
@@ -62,9 +62,9 @@
 @provide_api_client
 def deploy_cli(api_client, spec_arg, spec, allow_duplicate_names, pipeline_id):
     """
-    Deploys a delta pipeline according to the pipeline specification. The pipeline spec is a
-    specification that explains how to run a Delta Pipeline on Databricks. All local libraries
-    referenced in the spec are uploaded to DBFS.
+    Deploys a pipeline according to the pipeline specification. The pipeline spec is a
+    JSON document that defines the required settings to run a Delta Live Tables pipeline
```
PaulCornellDB, line 66:
```suggestion
    JSON document that defines the required settings to run a Delta Live Tables pipelines
```

databricks_cli/pipelines/cli.py
```diff
@@ -107,7 +107,7 @@ def deploy_cli(api_client, spec_arg, spec, allow_duplicate_names, pipeline_id):
         if (pipeline_id and 'id' in spec_obj) and pipeline_id != spec_obj["id"]:
             raise ValueError(
                 "The ID provided in --pipeline_id '{}' is different from the id provided "
```
PaulCornellDB, line 109:
```suggestion
                "The ID provided in --pipeline_id '{}' is different from the ID provided "
```

databricks_cli/pipelines/cli.py
```diff
@@ -107,7 +107,7 @@ def deploy_cli(api_client, spec_arg, spec, allow_duplicate_names, pipeline_id):
...
-                "in the spec '{}'. Please resolve the conflict and try the command again. "
+                "in the spec '{}'. Resolve the conflict and try the command again. "
                 "Because pipeline IDs are no longer persisted after being deleted, we "
```
PaulCornellDB, line 111:
```suggestion
                "Because pipeline IDs are no longer persisted after being deleted, Databricks "
```

databricks_cli/pipelines/cli.py
```diff
@@ -107,7 +107,7 @@ def deploy_cli(api_client, spec_arg, spec, allow_duplicate_names, pipeline_id):
...
                 "recommend removing the ID field from your spec."
```
PaulCornellDB, line 112:
```suggestion
                "recommends removing the ID field from your spec."
```

databricks_cli/pipelines/cli.py
pradeepgv-db: just fyi, reverted this from update to start, see my other comment.

databricks_cli/pipelines/cli.py
```diff
@@ -208,19 +213,47 @@ def reset_cli(api_client, pipeline_id):
...
+@profile_option
+@pipelines_exception_eater
+@provide_api_client
+def start_cli(api_client, pipeline_id, full_refresh):
+    """
+    Starts a pipelines run by starting the cluster and processing data.
+
+    Usage:
+
+    databricks pipelines start --pipeline-id 1234 --full-refresh=(false|true)
```
PaulCornellDB, line 248: Pick one, true or false.

PaulCornellDB (commented):
See my comments and proposed edits.

databricks_cli/pipelines/cli.py
```diff
@@ -186,20 +187,24 @@ def list_cli(api_client):
...
     _validate_pipeline_id(pipeline_id)
-    PipelinesApi(api_client).reset(pipeline_id)
-    click.echo("Reset triggered for pipeline {}".format(pipeline_id))
+    PipelinesApi(api_client).start_update(pipeline_id, full_refresh=True)
+    click.echo("Reset triggered for pipeline {}.".format(pipeline_id))
 
 
 @click.command(context_settings=CONTEXT_SETTINGS,
-               short_help='Starts the execution of a delta pipeline run')
+               short_help='[Deprecated] Starts a pipeline run.')
```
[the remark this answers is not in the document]
pradeepgv-db: PTAL, Changed this to [Deprecated] <Alternative usage>. <Help message>.

databricks_cli/pipelines/cli.py
pradeepgv-db: done

databricks_cli/pipelines/cli.py
```diff
@@ -62,9 +62,9 @@
 @provide_api_client
 def deploy_cli(api_client, spec_arg, spec, allow_duplicate_names, pipeline_id):
     """
-    Deploys a delta pipeline according to the pipeline specification. The pipeline spec is a
-    specification that explains how to run a Delta Pipeline on Databricks. All local libraries
-    referenced in the spec are uploaded to DBFS.
+    Deploys a pipeline according to the pipeline specification. The pipeline spec is a
+    JSON document that defines the required settings to run a Delta Live Tables pipelines
```
jaseidman, line 66:
```suggestion
    JSON document that defines the required settings to run a Delta Live Tables pipeline
```

This should be singular to match the the "run a".

databricks_cli/pipelines/cli.py
```diff
@@ -208,19 +215,47 @@ def reset_cli(api_client, pipeline_id):
 @provide_api_client
 def run_cli(api_client, pipeline_id):
     """
-    Starts the execution of a delta pipelines run by starting the cluster and processing data.
+    [Deprecated] Use the "start" command instead.
+
+    Starts a pipeline run by starting the cluster and processing data.
```
anew, line 220: run/update?

databricks_cli/pipelines/cli.py
```diff
@@ -186,20 +188,25 @@ def list_cli(api_client):
...
-    PipelinesApi(api_client).reset(pipeline_id)
-    click.echo("Reset triggered for pipeline {}".format(pipeline_id))
+    PipelinesApi(api_client).start_update(pipeline_id, full_refresh=True)
+    click.echo("Reset triggered for pipeline {}.".format(pipeline_id))
 
 
 @click.command(context_settings=CONTEXT_SETTINGS,
-               short_help='Starts the execution of a delta pipeline run')
+               short_help='[Deprecated] Use the "start" command instead. ' +
+                          'Starts a pipeline run.')
```
anew, line 209: run/update

databricks_cli/pipelines/cli.py
```diff
@@ -208,19 +215,47 @@ def reset_cli(api_client, pipeline_id):
...
 
     Usage:
 
     databricks pipelines run --pipeline-id 1234
     """
+    click.echo("Deprecation warning: the \"run\" command is deprecated." +
+               " Use the \"start\" command instead.")
+    _validate_pipeline_id(pipeline_id)
+    PipelinesApi(api_client).start_update(pipeline_id, full_refresh=False)
+    click.echo("Run triggered for pipeline {}.".format(pipeline_id))
```
anew, line 230: Update

databricks_cli/pipelines/cli.py
```diff
@@ -186,20 +188,25 @@ def list_cli(api_client):
...
 
     databricks pipelines reset --pipeline-id 1234
     """
+    click.echo("DeprecationWarning: the \"reset\" command is deprecated, " +
+               "use the \"start --full-refresh\" command instead.")
     _validate_pipeline_id(pipeline_id)
-    PipelinesApi(api_client).reset(pipeline_id)
-    click.echo("Reset triggered for pipeline {}".format(pipeline_id))
+    PipelinesApi(api_client).start_update(pipeline_id, full_refresh=True)
+    click.echo("Reset triggered for pipeline {}.".format(pipeline_id))
```
anew, line 204: I don't think "triggered" is a good word here. Since we recommend to use `start --full-refresh` here, should the message be the same? That is, `Started an update for pipeline {id}.`

databricks_cli/pipelines/cli.py
```diff
@@ -208,19 +215,47 @@ def reset_cli(api_client, pipeline_id):
...
     """
+    click.echo("Deprecation warning: the \"run\" command is deprecated." +
+               " Use the \"start\" command instead.")
+    _validate_pipeline_id(pipeline_id)
+    PipelinesApi(api_client).start_update(pipeline_id, full_refresh=False)
+    click.echo("Run triggered for pipeline {}.".format(pipeline_id))
+
+
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Starts a pipeline run.')
```
anew, line 234: run/update

databricks_cli/pipelines/cli.py
```diff
@@ -208,19 +215,47 @@ def reset_cli(api_client, pipeline_id):
...
+
+    Usage:
+
+    databricks pipelines start --pipeline-id 1234 --full-refresh=true
+    """
     _validate_pipeline_id(pipeline_id)
-    PipelinesApi(api_client).run(pipeline_id)
-    click.echo("Run triggered for pipeline {}".format(pipeline_id))
+    PipelinesApi(api_client).start_update(pipeline_id, full_refresh=full_refresh)
+    click.echo("Started an update for pipeline {}.".format(pipeline_id))
```
anew, line 254: should this message indicate whether it was a full refresh?

anew: This should print the update id of the update that was started.

databricks_cli/pipelines/cli.py
```diff
@@ -229,7 +264,7 @@ def run_cli(api_client, pipeline_id):
 @provide_api_client
 def stop_cli(api_client, pipeline_id):
     """
-    Stops the execution of a delta pipelines run by terminating the cluster. Processing of data can
+    Stops the pipeline run by terminating the cluster. Processing of data can
```
anew, line 267: This is not accurate any longer. I think it should say:
"Stops the pipelines by cancelling any active run".
and nothing about resuming processing, because that is only true in a few cases

anew (changes requested):

databricks_cli/pipelines/cli.py
pradeepgv-db: done

databricks_cli/pipelines/cli.py
pradeepgv-db: done

databricks_cli/pipelines/cli.py
pradeepgv-db: done

databricks_cli/pipelines/cli.py
pradeepgv-db: changed to a similar message across start/reset/run commands with update_id and mentioning if it's a full_refresh.

databricks_cli/pipelines/cli.py
pradeepgv-db: done

databricks_cli/pipelines/cli.py
pradeepgv-db: done

databricks_cli/pipelines/cli.py
pradeepgv-db: done

databricks_cli/pipelines/cli.py
```diff
@@ -134,7 +134,7 @@ def deploy_cli(api_client, spec_arg, spec, allow_duplicate_names, pipeline_id):
 @provide_api_client
 def delete_cli(api_client, pipeline_id):
     """
-    Stops a delta pipeline and deletes its associated Databricks resources. The pipeline can be
+    Stops the pipeline and deletes its associated Databricks resources. The pipeline can be
```
anew, line 137: I think we should say: "Stops the pipeline by cancelling any active update". It is not true anymore that 
- this deletes the associated resources (e..g, the maintenance jobs stays around, and we keep the cluster alive for a while)
- you don't need to redeploy to restart the pipeline. 
So let's just be brief here and say only the bare minimum.

databricks_cli/pipelines/cli.py
```diff
@@ -208,19 +215,47 @@ def reset_cli(api_client, pipeline_id):
 @provide_api_client
 def run_cli(api_client, pipeline_id):
     """
-    Starts the execution of a delta pipelines run by starting the cluster and processing data.
+    [Deprecated] Use the "start" command instead.
+
+    Starts a pipeline update by starting the cluster and processing data.
```
anew, line 220: similar comment. It does not always start a cluster. I think we should not speak about whether/when clusters are started, terminated, or reused, because is a detail that will change in the future as the product evolves.

databricks_cli/sdk/service.py
```diff
@@ -1060,6 +1060,14 @@ def run(self, pipeline_id=None, headers=None):
         return self.client.perform_query('POST', '/pipelines/{pipeline_id}/run'.format(pipeline_id=pipeline_id),
                                          data=_data, headers=headers)
 
+    def start_update(self, pipeline_id=None, full_refresh=False, headers=None):
+        _data = {}
+        if full_refresh:
```
anew, line 1065: minor: is this if statement needed? It defaults to false.

anew (approved):
I think the code changes here are good (see one minor comment).

I also think that all the messages need to be reworded. They reflect legacy terminology and an outdated version of the product. This should be done separately, by taking a pass over all messages and making them consistent, for example:
- we don't have runs but updates now
- we don't (always) delete all resources when stopping
- we don't require redeploy in order to restart a pipeline
- etc.

anew:
@jaseidman see my comment above. I think we need to make a pass over all messages in this part of the CLI. What do you think?

jaseidman:
@anew I definitely support making the messages consistent and technically correct. Otherwise, I think the messages have gone through extensive proofreading and review for style, especially by @PaulCornellDB. I can definitely take another pass once we're comfortable with the technical content of the messages.

databricks_cli/pipelines/cli.py
pradeepgv-db: Updated, this message is same as `stop` command, behavior to user is the same in both the cases?

databricks_cli/pipelines/cli.py
pradeepgv-db: Updated to "Starts a pipeline update."
Created https://databricks.atlassian.net/browse/SC-93969 for cli message updates.

pradeepgv-db:
> @anew I definitely support making the messages consistent and technically correct. Otherwise, I think the messages have gone through extensive proofreading and review for style, especially by @PaulCornellDB. I can definitely take another pass once we're comfortable with the technical content of the messages.

@jaseidman @PaulCornellDB Just FYI, created this JIRA ticket (https://databricks.atlassian.net/browse/SC-93969)
to make the CLI messages more consistent with the backend behavior across all the commands (w.r.t what @anew was mentioning above). This is to avoid increasing scope of this PR and decouple that work into it's own task. Will try to figure out how/when to get that task done.

dylanstee-db (approved):
Thanks for keeping the CLI up-to-date! 🐥

databricks_cli/click_types.py
```diff
@@ -98,12 +98,12 @@ class SecretPrincipalClickType(ParamType):
...
 class PipelineSpecClickType(ParamType):
     name = 'SPEC'
-    help = 'The path to the pipelines deployment spec file'
+    help = 'The path to the pipelines deployment spec file.'
 
 
 class PipelineIdClickType(ParamType):
     name = 'PIPELINE_ID'
-    help = 'Delta Pipeline ID'
+    help = 'The Delta Pipeline ID.'
```
PaulCornellDB, line 106: Should be "The Delta Live Tables pipeline ID" or just "The pipeline ID."

pradeepgv-db: yeah just "The Pipeline ID" should be good here.

databricks_cli/click_types.py
```diff
@@ -98,12 +98,12 @@ class SecretPrincipalClickType(ParamType):
...
 class PipelineSpecClickType(ParamType):
     name = 'SPEC'
-    help = 'The path to the pipelines deployment spec file'
+    help = 'The path to the pipelines deployment spec file.'
 
 
 class PipelineIdClickType(ParamType):
     name = 'PIPELINE_ID'
-    help = 'Delta Pipeline ID'
+    help = 'The Pipeline ID.'
```
PaulCornellDB, line 106:
```suggestion
    help = 'The pipeline ID.'
```

databricks_cli/pipelines/cli.py
```diff
@@ -229,16 +263,27 @@ def run_cli(api_client, pipeline_id):
 @provide_api_client
 def stop_cli(api_client, pipeline_id):
     """
-    Stops the execution of a delta pipelines run by terminating the cluster. Processing of data can
-    be resumed by calling `run`.
+    Stops the pipelines by cancelling any active update.
```
PaulCornellDB, line 266:
```suggestion
    Stops the pipeline by cancelling any active update.
```

databricks_cli/pipelines/cli.py
```diff
@@ -274,7 +320,7 @@ def _write_spec(src, spec):
 
 def _validate_pipeline_id(pipeline_id):
     """
-    Checks if the pipeline_id is not empty and only contains -, _ and alphanumeric characters
+    Checks if the pipeline_id is not empty and only contains -, _ and alphanumeric characters.
```
PaulCornellDB, line 323:
```suggestion
    Checks if the pipeline ID is not empty and contains only hyphen (-), underscore (_), and alphanumeric characters.
```

databricks_cli/pipelines/cli.py
```diff
@@ -274,7 +320,7 @@ def _write_spec(src, spec):
...
     """
     if pipeline_id is None or len(pipeline_id) == 0:
         error_and_quit(u'Empty pipeline id provided')
```
PaulCornellDB, line 326: "id" -> "ID"

pradeepgv-db: done

databricks_cli/pipelines/cli.py
pradeepgv-db: done

databricks_cli/pipelines/cli.py
pradeepgv-db: done

databricks_cli/click_types.py
pradeepgv-db: done

PaulCornellDB (approved):
LGTM from an editorial (not technical) perspective.

anew:
oops. I had more comments but after submitting saw that this was already merged. Maybe address in a follow-up PR. Mot importantly, we should be consistent with the UI and call it "settings" instead of "spec".

pradeepgv-db:
> oops. I had more comments but after submitting saw that this was already merged. Maybe address in a follow-up PR. Mot importantly, we should be consistent with the UI and call it "settings" instead of "spec".

yeah will address these in the next PR, we can finalize on the commands/docs/messages here and then go through review with docs team. https://docs.google.com/document/d/1KXS9B66uWGaGXEaFDOiXt6gLcdPAlmDSqv0gPLw0vvo/edit#heading=h.9waxn8f8b8x

PR #425, opened 2022-02-03
[Jobs] Add `--python-named-params` to databricks jobs run-now

nordp:
This allows users to override the `python_named_parameters` in the `jobs/run-now` endpoint.

Usage example:
```sh
databricks jobs run-now --job-id 12345 --python-named-params '{"parameter": "value"}'
```

Corresponding API payload to `api/2.1/jobs/run-now`:
```
{
    "job_id": 12345,
    "python_named_params": {
        "parameter": "value"
    }
}  
```

### DIFF
--- a/databricks_cli/jobs/api.py
+++ b/databricks_cli/jobs/api.py
@@ -57,6 +57,5 @@ class JobsApi(object):
 
     def _list_jobs_by_name(self, name, headers=None):
         jobs = self.list_jobs(headers=headers)['jobs']
-        result = list(
-            filter(lambda job: job['settings']['name'] == name, jobs))
+        result = list(filter(lambda job: job['settings']['name'] == name, jobs))
         return result
--- a/databricks_cli/jobs/api.py
+++ b/databricks_cli/jobs/api.py
@@ -49,12 +49,14 @@ class JobsApi(object):
         return self.client.client.perform_query('POST', '/jobs/reset', data=json, headers=headers,
                                                 version=version)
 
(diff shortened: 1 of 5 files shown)


databricks_cli/jobs/api.py
```diff
@@ -49,12 +49,14 @@ def reset_job(self, json, headers=None, version=None):
...
         return self.client.run_now(job_id, jar_params, notebook_params, python_params,
-                                   spark_submit_params, headers=headers, version=version)
+                                   python_named_params, spark_submit_params, headers=headers,
+                                   version=version)
 
     def _list_jobs_by_name(self, name, headers=None):
         jobs = self.list_jobs(headers=headers)['jobs']
-        result = list(filter(lambda job: job['settings']['name'] == name, jobs))
+        result = list(
+            filter(lambda job: job['settings']['name'] == name, jobs))
```
nordp, line 61:
```suggestion
        result = list(filter(lambda job: job['settings']['name'] == name, jobs))
```

KoningJasper (approved):

skyxie:
This change added `python_named_params` as a required positional parameter to `JobsApi.run_now`, which breaks existing use of this function. Since `python_named_params=None` by default in `JobsService.run_now`, then I believe that should also be the case in `JobsApi.run_now`.

nordp:
@skyxie Thanks for bringing this to my attention. I followed your proposal in #429 which has been released as a bugfix version [0.16.4](https://github.com/databricks/databricks-cli/releases/tag/0.16.4)

PR #428, opened 2022-02-03
Better error handling in Repos

alexott:
In current version, we have "catch-all" error handling that efficiently hides the actual reason for the error. It's not possible to understand the reason for error, for example, when IP Access List is used. This patch improves the error reporting:

Without patch

```
Error: RuntimeError: Can't find repo ID for /Repos/user/repo
```

With patch:

```
Error: RuntimeError: Error fetching repo ID for /Repos/user/repo: HTTP code: 403, Unauthorized access to Org: 1234567890
```


a-cong (approved):

jaipreet-s:
Changes LGTM, however, it would be good to add unit tests for the new code paths before merging

alexott:
I've added a test, but I can't see what actual coverage change is

nfx (approved):
lgtm

Issue #415, opened 2021-12-30
Documentation of runs cli does not match implementation


epilif1017a:
Although the databricks documentation references we can append `--version=2.1` to `databricks runs` commands, the same throws an error when we put that option in. Moreover, by looking at the code, `--version` does not appear to be one of the click.options, so the actual implementation does not seem to match the documentation here: https://docs.databricks.com/dev-tools/cli/runs-cli.html 

Suggested change would be to put that as a click option and pass it downstream to the api calls.

epilif1017a:
Hi team, 

Any news on the issue?

Best regards,
Carlos

nordp:
Hi @epilif1017a,

Thanks for reporting this, please take a look at #434 and let me know if this PR would solve your issue.

epilif1017a:
Hi @nordp ,

LGTM. Really appreciated! I'll be waiting for the next release then. Thanks a lot!

Issue #405, opened 2021-12-02
Runs CLI does not support Jobs 2.1 API


juve:
It looks like the runs CLI does not support the new Jobs 2.1 API. There is no `--version` flag to specify the Jobs 2.1 API when using `databricks runs submit`, for example, and it doesn't look like the runs CLI uses the new `jobs-api-version` config parameter.

nordp:
Hi @juve,

Thanks for reporting this, I've put together a PR to address this #434. Feel free to take a look.
I've not yet done QA to test that the `jobs-api-version` is passed through properly.

PR #434, opened 2022-02-25
[JOBS-5871] Add version option to runs

nordp:
As reported in #405 and #415, the `--version` option to override the configured `JOBS_API_VERSION` is not available in the `databricks runs` command group.

This PR ensures the same `@api_version_option` is used on the `databricks jobs` and `databricks runs` commands.


bbezine-databricks (approved):
LGTM

PR #436, opened 2022-02-28
Add expand tasks flag to runs list command

nordp:
Adds support for `--expand-tasks` and `--all` options in the `databricks runs list` command.

From the Jobs API 2.1 [docs](https://docs.databricks.com/dev-tools/api/latest/jobs.html#operation/JobsRunsList):

> expand_tasks: boolean (Default: false) | Whether to include task and cluster details in the response.
> -- | --

The `--all` command is available in the `databricks jobs list` command, so I thought it could be useful for the runs list too.

### DIFF
--- a/databricks_cli/runs/cli.py
+++ b/databricks_cli/runs/cli.py
@@ -113,13 +113,18 @@ def _runs_to_table(runs_json):
 @click.option('--limit', default=None, type=int,
               help='The limit determines the number of runs listed. '
                    'Limit must be between 0 and 1000. Set to 20 runs by default.')
+@click.option('--expand-tasks', is_flag=True,
+              help='Expands the tasks array (only available in API 2.1).')
+@click.option('--all', '_all', is_flag=True,
+              help='Lists all jobs by executing sequential calls to the API ' +
+                   '(only available in API 2.1).')
 @click.option('--output', help=OutputClickType.help, type=OutputClickType())
 @api_version_option
 @debug_option
 @profile_option
 @eat_exceptions  # noqa
 @provide_api_client
-def list_cli(api_client, job_id, active_only, completed_only, offset, limit, output, version):  # noqa
+def list_cli(api_client, job_id, active_only, completed_only, offset, limit, expand_tasks, output, version, _all):  # noqa
     """
     Lists job runs.
 
@@ -137,8 +142,26 @@ def list_cli(api_client, job_id, active_only, completed_only, offset, limit, out
       - Result state (can be n/a)
     """
     check_version(api_client, version)
-    runs_json = RunsApi(api_client).list_runs(
-        job_id, active_only, completed_only, offset, limit, version=version)
+    api_version = version or api_client.jobs_api_version
+    if api_version != '2.1' and (expand_tasks or offset or limit or _all):
+        click.echo(click.style('ERROR', fg='red') + ': the options --expand-tasks, ' +
+                   '--offset, --limit, and --all are only available in API 2.1', err=True)
+        return
+    runs_api = RunsApi(api_client)
+
+    has_more = True
+    runs = []
+    if _all:
+        offset = 0
+        limit = 20
+    while has_more:
+        runs_json = runs_api.list_runs(
+            job_id, active_only, completed_only, offset, limit, version=version, expand_tasks=expand_tasks)
+        runs += runs_json['runs'] if 'runs' in runs_json else []
+        has_more = runs_json.get('has_more', False) and _all
+        if has_more:
+            offset = offset + \
+                (len(runs_json['runs']) if 'runs' in runs_json else limit)
     if OutputClickType.is_json(output):
         click.echo(pretty_format(runs_json))
     else:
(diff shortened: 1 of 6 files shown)


databricks_cli/runs/cli.py
```diff
@@ -106,12 +112,32 @@ def list_cli(api_client, job_id, active_only, completed_only, offset, limit, out
...
+                   '--offset, --limit, and --all are only available in API 2.1', err=True)
+        return
+    runs_api = RunsApi(api_client)
+
+    has_more = True
+    runs = []
+    if _all:
+        offset = 0
+        limit = 20
+    while has_more:
```
KoningJasper, line 127: Should we make this more generic since `jobs/list` does the same thing.

databricks_cli/sdk/service.py
```diff
@@ -176,7 +176,7 @@ def run_now(self, job_id=None, jar_params=None, notebook_params=None, python_par
         return self.client.perform_query('POST', '/jobs/run-now', data=_data, headers=headers, version=version)
 
     def list_runs(self, job_id=None, active_only=None, completed_only=None, offset=None,
-                  limit=None, headers=None, version=None):
+                  limit=None, expand_tasks=None, headers=None, version=None):
```
KoningJasper, line 179: Won't this break somebody who relies on positional parameters?

KoningJasper (changes requested):

pietern:
@nordp Ping on this. Do you want to continue this PR or abort it?

databricks_cli/runs/cli.py
```diff
@@ -106,12 +112,32 @@ def list_cli(api_client, job_id, active_only, completed_only, offset, limit, out
...
+        runs_json = runs_api.list_runs(
+            job_id, active_only, completed_only, offset, limit,
+            version=version, expand_tasks=expand_tasks)
+        runs += runs_json['runs'] if 'runs' in runs_json else []
+        has_more = runs_json.get('has_more', False) and _all
+        if has_more:
+            offset = offset + \
+                (len(runs_json['runs']) if 'runs' in runs_json else limit)
```
pietern, line 135: @nordp The command itself has offset and limit arguments. I think you'd want to make `--all` mutually exclusive with specifying `--offset` and `--limit`, and then introduce a separate branch for `--all` here to put the logic for repeated calls to the API. Alternatively, right after L131, something like:

```py
if not _all:
  break
else:
  has_more = runs_json.get('has_more', False)
  # ...
```

databricks_cli/runs/cli.py
```diff
@@ -106,12 +112,32 @@ def list_cli(api_client, job_id, active_only, completed_only, offset, limit, out
       - Result state (can be n/a)
     """
     check_version(api_client, version)
-    runs_json = RunsApi(api_client).list_runs(
-        job_id, active_only, completed_only, offset, limit, version=version)
+    api_version = version or api_client.jobs_api_version
+    if api_version != '2.1' and (expand_tasks or offset or limit or _all):
+        click.echo(click.style('ERROR', fg='red') + ': the options --expand-tasks, ' +
+                   '--offset, --limit, and --all are only available in API 2.1', err=True)
+        return
```
pietern, line 119: Why does the user need to care about the API version in the first place? If every workspace has both APIs, why don't we only use the 2.1 API in the CLI?

pietern (commented):
I now see you're basically copying the code from the command to list jobs to the command to list runs, and as such copy the structure. Can you comment on my Q's regardless? Thanks!

databricks_cli/runs/cli.py
nordp: We provide the option of choosing API version due to slight differences in the response structure.

pietern: Of course, understood.

databricks_cli/runs/cli.py
pietern: @nordp Do you want to address this here or in a follow up PR?

pietern: That's fine. Can you rebase?

databricks_cli/runs/cli.py
nordp: I don't think I want to make it mutually exclusive, because a user might want to ask for a higher limit than what is supported by the API, but still not fetch ALL jobs/runs.

databricks_cli/runs/cli.py
```diff
@@ -113,13 +113,19 @@ def _runs_to_table(runs_json):
 @click.option('--limit', default=None, type=int,
               help='The limit determines the number of runs listed. '
                    'Limit must be between 0 and 1000. Set to 20 runs by default.')
+@click.option('--expand-tasks', is_flag=True,
+              help='Expands the tasks array (only available in API 2.1).')
+@click.option('--all', '_all', is_flag=True,
```
KoningJasper, line 118: This is potentially a problem because runs/list is essentially unbounded. So this would result in nearly infinite requests or a DOS

databricks_cli/runs/cli.py
```diff
@@ -137,12 +143,33 @@ def list_cli(api_client, job_id, active_only, completed_only, offset, limit, out
...
     check_version(api_client, version)
-    runs_json = RunsApi(api_client).list_runs(
-        job_id, active_only, completed_only, offset, limit, version=version)
+    api_version = version or api_client.jobs_api_version
+    if api_version != '2.1' and (expand_tasks or offset or limit or _all):
+        click.echo(click.style('ERROR', fg='red') + ': the options --expand-tasks, ' +
+                   '--offset, --limit, and --all are only available in API 2.1', err=True)
+        return
+    runs_api = RunsApi(api_client)
+    runs = walk_paginated_api(
```
KoningJasper, line 152: We should bound this I think

KoningJasper (changes requested):

PR #439, opened 2022-03-09
handle NotADirectoryError exception when attempting to create directory

### LINKED ISSUE #438: NotADirectoryError when running workspace export_dir on Windows
If I run `databricks workspace export_dir` on Windows, I get a `NotADirectoryError` error when attempting to download directories with invalid Windows names (e.g., all the databricks_automl subdirectories, all of which contain a colon).

Example:
```bash
> databricks workspace export_dir /Users/me@example.com .

*snip*

Error: NotADirectoryError: [WinError 267] The directory name is invalid: '.\\databricks_automl\\22-01-05-17:00-automl_test_1-deadb33f'
```

Not only does it fail to download all the folders in /databricks_automl, it also bails out of the loop so none of the other files get downloaded.

snosrap:
Fixes #438

### DIFF
--- a/databricks_cli/workspace/api.py
+++ b/databricks_cli/workspace/api.py
@@ -171,7 +171,11 @@ class WorkspaceApi(object):
                        .format(target_path, source_path))
             return
         if not os.path.isdir(target_path):
-            os.makedirs(target_path)
+            try:
+                os.makedirs(target_path)
+            except NotADirectoryError:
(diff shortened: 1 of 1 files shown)


nfx (approved):
lgtm

pietern:
PR #448 fixed the same issue by replacing incompatible characters with `_`.

Could you give the latest version of the CLI a try and see if it works for you?

PR #450, opened 2022-04-05
Add devcontainer configuration

fjakobs:
Allows working with VSCode remote containers and Github CodeSpaces.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,23 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.231.3/containers/python-3/.devcontainer/base.Dockerfile
+
+# [Choice] Python version (use -bullseye variants on local arm64/Apple Silicon): 3, 3.10, 3.9, 3.8, 3.7, 3.6, 3-bullseye, 3.10-bullseye, 3.9-bullseye, 3.8-bullseye, 3.7-bullseye, 3.6-bullseye, 3-buster, 3.10-buster, 3.9-buster, 3.8-buster, 3.7-buster, 3.6-buster
+ARG VARIANT="3.10-bullseye"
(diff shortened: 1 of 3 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,54 @@
...
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.231.3/containers/python-3
+{
+	"name": "Python 3",
+	"build": {
+		"dockerfile": "Dockerfile",
+		"context": "..",
+		"args": { 
+			// Update 'VARIANT' to pick a Python version: 3, 3.10, 3.9, 3.8, 3.7, 3.6
+			// Append -bullseye or -buster to pin to an OS version.
+			// Use -bullseye variants on local on arm64/Apple Silicon.
```
lennartkats-db, line 11: How does that work exactly for non-ARM?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.231.3/containers/python-3/.devcontainer/base.Dockerfile
```
lennartkats-db, line 1: Could you extend the README.md with a small paragraph about how this is supported for development?

.devcontainer/devcontainer.json
fjakobs: This is fully transparent. The ARM enabled docker images are multi-platform and docker will automatically pull the correct platform image. Some of the older images are x86 only. Those work on M1 Macs but only in slow emulation mode.

.devcontainer/Dockerfile
fjakobs: Absolutely. Will add it.

fjakobs: done

nfx (approved):
LGTM

Issue #341, opened 2020-10-13
Add `jobs run-script` command


nfx:
Use case `databricks jobs run-script metastore-export.py` as equivalent for:

```
databricks workspace import metastore-export.py /tmp/metastore-export_$RANDOM --language python --overwrite
# creating Spark Submit job - https://docs.databricks.com/dev-tools/api/latest/jobs.html#runs-submit
# waiting for completion of the job, perhaps showing status update every 30 seconds
databricks fs rm dbfs:/tmp/metastore-export_$RANDOM
```

TBD: cluster parameters

fjakobs:
With #455 you could do this:

`databricks execution-context command-execute-once --cluster-id <CLUSTER_ID> --command "$(cat metastore-export.py)" --wait`

nfx:
@fjakobs how can we make that invocation simpler?

fjakobs:
Just sketching:

We could
- move the command under the cluster group
- make `--wait=True` the default
- add an argument that reads the command from a file
- add an option to reference a cluster by name

`databricks cluster execute --cluster-name <CLUSTER_NAME> --command-file metastore-export.py`

nfx:
@fjakobs that looks a lot simpler!

how the results are going to be exported, e.g. if there's `return` in the end of the script?

fjakobs:
We could have different modes. The default could just be informative text based output like I have already implemented:

```
$ python -m databricks_cli.cli --profile AWS-DOGFOOD execution-context command-execute-once --cluster-id 0819-204509-hill72  --command "$(cat spark.py)" --wait=True 
Status: Queued
Status: Running
Status: Finished


Command ID: c98edf8a-418a-4fa0-b69c-dcfae9db917e

output > +---------+----------+--------+----------+------+------+
output > |firstname|middlename|lastname|       dob|gender|salary|
output > +---------+----------+--------+----------+------+------+
output > |    James|          |   Smith|1991-04-01|     M|  3000|
output > |  Michael|      Rose|        |2000-05-19|     M|  4000|
output > |   Robert|          |Williams|1978-09-05|     M|  4000|
output > |    Maria|      Anne|   Jones|1967-12-01|     F|  4000|
output > |      Jen|      Mary|   Brown|1980-02-17|     F|    -1|
output > +---------+----------+--------+----------+------+------+
```

For use in shell scripts we can also just return the last status result with the embedded data as JSON:

```
[...]

Issue #122, opened 2018-05-11
This is more of a question Can databricks-cli be used to run a notebook


OtisHarrison24:
I have a python script that pulls a ton of data into s3 and I created a notebook in databricks to turn it into a parquet file. I was wondering if I can use the cli to run a job after i finish pulling in all my data?
Thanks

rogelioamancisidor:
suppose --job-id 55 runs a .py file that contains <print> statements? how can I print those to my console when I run the command <databricks jobs run-now --job-id 55>? some sort of 'follow' mode.

luzlab:
It would be awesome the CLI had a synchronous method for executing notebooks. Right now, we have to poll the job to see when it's done. Even if the CLI returned only the notebook exit value at the end, it would be very useful when orchestrating databricks operations.

nfx:
Might be fixed by #455

chrisst:
**This repository has been archived.**

The legacy `databricks/databricks-cli` is deprecated and has been fully replaced by the new Databricks CLI at **https://github.com/databricks/cli**.

We are closing all open issues here as part of archiving this repository. If your issue still affects the new CLI, we'd appreciate you re-opening it at https://github.com/databricks/cli/issues so the team can track and address it there.

PR #455, opened 2022-05-04
Add execution context API

fjakobs:
This PR adds the still supported functionality from the REST API 1.2.

It allows users to create execution contexts and run Python, SQL, and R code on a cluster.

Note that his PR is still missing tests and can't be merged as is.


nfx (changes requested):
https://github.com/databricks/databricks-cli/commit/6c429d4f581fc99f54cbf9239879434310037cc7#diff-72d81b5e87625d187fbc145e395e1981ef31119974e19e2b34dfa3ab0a9b983bR32-R51 Command/Context API support was added along with Tunnels PoC

i'd also suggest to move .devcontainer into separate PR

fjakobs:
I moved the dev container code out into https://github.com/databricks/databricks-cli/pull/450 and made sure the linter is happy.

databricks_cli/execution_context/api.py
```diff
@@ -0,0 +1,75 @@
...
+            self.id = self.client.create_context(cluster_id=self.cluster_id)["id"]
+
+        return self
+
+    def __exit__(self, _type, _value, _traceback):
+        self.client.delete_context(cluster_id=self.cluster_id, context_id=self.id)
+        self.id = None
+
+
+class ExecutionContextApi(object):
```
nfx, line 49: Could you reuse the basic structure for command/context APIs from Tunnel CLI, so that it won't break, once merged? :) 
https://github.com/databricks/databricks-cli/commit/6c429d4f581fc99f54cbf9239879434310037cc7

it's removed, so probably it has to be re-added
https://github.com/databricks/databricks-cli/commit/0296117f531e430cd2f819652ec0837dce14ae98

databricks_cli/execution_context/cli.py
```diff
@@ -0,0 +1,251 @@
...
+@click.option('--cluster-id', required=True, type=ClusterIdClickType(),
+              help=ClusterIdClickType.help)
+@click.option('--language', required=False, default="python", 
+              type=click.Choice(['python', 'scala', 'sql'], case_sensitive=False),
+              help="The language for the context.")
+@debug_option
+@profile_option
+@eat_exceptions
+@provide_api_client
+def create_context_cli(api_client, cluster_id, language): #  NOQA
```
nfx, line 50: this command makes no sense in isolation within the CLI from end-user perspective. what makes more sense from practical perspective is "run _this python_ code on cluster with _this name_ and return it's results".

databricks_cli/execution_context/cli.py
```diff
@@ -0,0 +1,251 @@
...
+               short_help="Gets information about an execution context.")
+@click.option('--cluster-id', required=True, type=ClusterIdClickType(),
+              help=ClusterIdClickType.help)
+@click.option('--context-id', required=True, type=ExecutionContextIdClickType(),
+              help=ExecutionContextIdClickType.help)
+@debug_option
+@profile_option
+@eat_exceptions
+@provide_api_client
+def get_context_status_cli(api_client, cluster_id, context_id): #  NOQA
```
nfx, line 68: this command makes no sense in isolation within the CLI from end-user perspective.

databricks_cli/execution_context/cli.py
```diff
@@ -0,0 +1,251 @@
...
+               short_help="Deletes an execution context.")
+@click.option('--cluster-id', required=True, type=ClusterIdClickType(),
+              help=ClusterIdClickType.help)
+@click.option('--context-id', required=True, type=ExecutionContextIdClickType(),
+              help=ExecutionContextIdClickType.help)
+@debug_option
+@profile_option
+@eat_exceptions
+@provide_api_client
+def delete_context_cli(api_client, cluster_id, context_id): #  NOQA
```
nfx, line 86: this command makes no sense in isolation within the CLI from end-user perspective.

databricks_cli/execution_context/cli.py
```diff
@@ -0,0 +1,251 @@
...
+              help="If true wait for command completion, otherwise schedule "
+                   "the command and return immediately.")
+@click.option('--language', required=False, default="python", 
+              type=click.Choice(['python', 'scala', 'sql'], case_sensitive=False),
+              help="The language for the context.")
+@debug_option
+@profile_option
+@eat_exceptions
+@provide_api_client
+def execute_command_cli(api_client, output, cluster_id, context_id, command, language, wait): #  NOQA
```
nfx, line 113: from end-user perspective, this command is better be exposed under "clusters" command group. e.g. `databricks cluster command this_thing.py --cluster-name "Shared Autoscaling"`. Language could be parsed from file argument extension (`this_thing.py`). In most of the cases, users don't want to copy-paste cluster_id and they want to supply names instead. If cluster is not running - start it. also need to thing about stopping cluster after command is finished, if cluster was not running before that.

databricks_cli/execution_context/cli.py
```diff
@@ -0,0 +1,251 @@
...
+              type=click.Choice(['python', 'scala', 'sql'], case_sensitive=False),
+              help="The language for the context.")
+@click.option('--wait', required=False, default=False, type=click.BOOL,
+              help="If true wait for command completion, otherwise schedule"
+                   " the command and return immediately.")
+@debug_option
+@profile_option
+@eat_exceptions
+@provide_api_client
+def execute_command_once_cli(api_client, output, cluster_id, command, language, wait): #  NOQA
```
nfx, line 145: probably is better to merge with the previous command. context creation/deletion is cheap enough to be done for every command.

databricks_cli/execution_context/cli.py
```diff
@@ -0,0 +1,251 @@
...
+        if not OutputClickType.is_json(output):
+            click.echo("Status: " + status["status"])
+        if status["status"] in ["Cancelled", "Error", "Finished"]:
+            click.echo()
+            _print_command_result(output, status)
+            break
+        time.sleep(1)
+
+
+def _print_command_result(output, status):
```
nfx, line 174: this code misses error handling. please take a look at terraform provider implementation of command result parsing.

databricks_cli/execution_context/cli.py
```diff
@@ -0,0 +1,251 @@
...
+        result = client.execute_command(cluster_id, context.id, command, language)
+        if not wait:
+            click.echo(pretty_format(result))
+            return
+
+        command_id = result["id"]
+        _wait_for_command(client, output, cluster_id, context.id, command_id)
+
+
+def _wait_for_command(client, output, cluster_id, context_id, command_id):
```
nfx, line 162: and what would happen if user executes `while True: time.sleep(1)` as the command? All retryable executions must have a timeout specified. 20 minutes is a reasonable default timeout. Is must be possible to override that timeout by user controlled parameter.

databricks_cli/execution_context/cli.py
```diff
@@ -0,0 +1,251 @@
...
+              help=ClusterIdClickType.help)
+@click.option('--context-id', required=True, type=ExecutionContextIdClickType(),
+              help=ExecutionContextIdClickType.help)
+@click.option('--command-id', required=True, type=CommandIdClickType(),
+              help=CommandIdClickType.help)
+@debug_option
+@profile_option
+@eat_exceptions
+@provide_api_client
+def cancel_command_cli(api_client, cluster_id, context_id, command_id): #  NOQA
```
nfx, line 206: this command makes no sense in isolation within the CLI from end-user perspective.

databricks_cli/execution_context/cli.py
```diff
@@ -0,0 +1,251 @@
...
+              help=ClusterIdClickType.help)
+@click.option('--context-id', required=True, type=ExecutionContextIdClickType(),
+              help=ExecutionContextIdClickType.help)
+@click.option('--command-id', required=True, type=CommandIdClickType(),
+              help=CommandIdClickType.help)
+@debug_option
+@profile_option
+@eat_exceptions
+@provide_api_client
+def get_command_status_cli(api_client, cluster_id, context_id, command_id): #  NOQA
```
nfx, line 225: this command makes no sense in isolation within the CLI from end-user perspective.

databricks_cli/execution_context/cli.py
```diff
@@ -0,0 +1,251 @@
...
+              expose_value=False, is_eager=True, help=version)
+@debug_option
+@profile_option
+@eat_exceptions
+def execution_context_group():  # pragma: no cover
+    """Utility to interact with execution contexts."""
+    pass
+
+
+execution_context_group.add_command(create_context_cli, name='create')
```
nfx, line 245: i think that logically this belongs to a cluster command group instead

databricks_cli/sdk/api_client.py
```diff
@@ -105,6 +105,12 @@ def __init__(self, user=None, password=None, host=None, token=None,
         self.api_version = api_version
         self.jobs_api_version = jobs_api_version
 
+    def get_v1_client(self):
```
nfx, line 108: this will make OAuth (and AAD, and OIDC) implementation fragile for long-running commands, where token might have to be refreshed couple of times to keep working. could we have another way of propagating `/api/1.2` path prefix? kwarg in `perform_query` method, for example.

databricks_cli/sdk/service.py
```diff
@@ -1114,3 +1117,80 @@ def delete_repo(self, id, headers=None):
         _data = {}
     
         return self.client.perform_query('DELETE', '/repos/{id}'.format(id=id), data=_data, headers=headers)
+
+class ExecutionContextService(object):
```
nfx, line 1121: Could you reuse the basic structure for command/context APIs from Tunnel CLI, so that it won't break, once merged? :)
https://github.com/databricks/databricks-cli/commit/6c429d4f581fc99f54cbf9239879434310037cc7

it's removed, so probably it has to be re-added
https://github.com/databricks/databricks-cli/commit/0296117f531e430cd2f819652ec0837dce14ae98

nfx (changes requested):
Left detailed review on proposed changes

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,54 @@
...
+{
+	"name": "Python 3",
+	"build": {
+		"dockerfile": "Dockerfile",
+		"context": "..",
+		"args": { 
+			// Update 'VARIANT' to pick a Python version: 3, 3.10, 3.9, 3.8, 3.7, 3.6
+			// Append -bullseye or -buster to pin to an OS version.
+			// Use -bullseye variants on local on arm64/Apple Silicon.
+			"VARIANT": "3.6-bullseye",
```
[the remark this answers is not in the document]
fjakobs: `tox` is still using 3.6 and will fail with any other version. I think bumping the python version should be a separate task.

databricks_cli/execution_context/cli.py
fjakobs: It does make sense when you want to run multiple commands within the same context. What you describe is possible with `command-execute-once`.

databricks_cli/execution_context/cli.py
fjakobs: It does make sense when you start a command async.

databricks_cli/execution_context/cli.py
fjakobs: It does make sense when you start a command async.

databricks_cli/sdk/api_client.py
fjakobs: I'll change it to use kwargs, that's indeed cleaner

databricks_cli/execution_context/cli.py
nfx: why people need to start individual notebook command async if they have run-submit jobs, that also have a snapshot of notebook being executed?

databricks_cli/execution_context/cli.py
nfx: because commands are possible only in scope of a running interactive cluster

databricks_cli/execution_context/cli.py
nfx: it's just not clear for me which end user journey it simplifies.

databricks_cli/execution_context/cli.py
nfx: i've never heard end users talking about command contexts :)

databricks_cli/execution_context/cli.py
nfx: `execute_command_cli` can fit into "workspace initialization" journey with implemented additions, as i see requests for something like that from time to time. current status quo is a run-submit a notebook job, which is also cancellable.

my concern is keeping new command surface to minimum.

databricks_cli/execution_context/cli.py
nfx: this command actually fixes this issue - https://github.com/databricks/databricks-cli/issues/341 - and solves certain operational workflow.

fjakobs:
@nfx what do you think about the following changes:

1. Keep the execution-context group but only keep the 1:1 mapping of commands to the API there. This are the low level commands for advanced usage. e.g. When running multiple commands, potentially in different languages, in the same context. This allows us to simulate notebooks from the CLI.
2. Add a single `execute` command to the cluster group that is smart and convenient and covers the 80% used case. This could look like described in https://github.com/databricks/databricks-cli/issues/341#issuecomment-1118302184

nfx:
@fjakobs it's just that i've literally never heard of any customer asking for interactive notebook individual commands to be exposed in CLI in any of workflows. 

It's quite the opposite for SDK for command execution - yes, we need to expose a python context manager for command context in the SDK, it has plenty of real-world scenarios. Some of the applications are https://github.com/databrickslabs/jupyterlab-integration and https://github.com/databrickslabs/migrate

and of course, there were high-level workflow requests, like #122 (10 votes since 2018) / #341 (back then there were too many issues, so i followed the broken window theory and created more).

chrisst:

PR #477, opened 2022-06-14
Use correct string method in create_repo function

molkuva:
Use correct string method `strip()` as python doesn't have `trim()` method for a string.
Bug Introduced in https://github.com/databricks/databricks-cli/pull/385


pietern:
We just released a patch version 0.16.8 that includes this fix.

PR #478, opened 2022-06-14
Fix typo in service.py - replacing trim() with strip()

VincentTNR:
replacing trim() with strip()


pietern:
@VincentTNR Thanks for posting the fix! I added it to #480 to ensure we have test coverage for this path.

pietern:
We just released a patch version 0.16.8 that includes this fix.

PR #480, opened 2022-06-15
Fix trim -> strip

### LINKED ISSUE #479: create_repo failing with AttributeError: 'str' object has no Attribute
It seems the latest release 0.16.7 a validation was added for provider arguement https://github.com/databricks/databricks-cli/blob/master/databricks_cli/sdk/service.py#L1137-L1138, where in it will attempt to do provider.trim(), it should be provider.strip() as trim is not a function in python @alexott

pietern:
This backfills tests to ensure that the fix is correct.

Closes #478.
Closes #479.


alexott (approved):
lgtm

PR #481, opened 2022-06-15
Add workflow that tags a version bump

pietern:


fjakobs (approved):
Nice, didn't know you can run python as a shell here.

PR #484, opened 2022-06-15
Restrict package to include only databricks_cli

pietern:
Because `setup.py` uses an exclude pattern rather than an include pattern we inadvertently include the `integration` directory as a top level package. This change modifies that to ensure we only package `databricks_cli`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ version = imp.load_source(
 setup(
     name='databricks-cli',
     version=version,
-    packages=find_packages(exclude=['tests', 'tests.*']),
+    packages=find_packages(include=['databricks_cli*']),
     install_requires=[
         # Note: please keep this in sync with `requirements.txt`.
         'click>=6.7',


pietern:
To test this, I ran `python setup.py egg_info` before and after this change and saw the following diff:
```diff
diff --git a/databricks_cli.egg-info/SOURCES.txt b/databricks_cli.egg-info/SOURCES.txt
index f9db6bd..80f4185 100644
--- a/databricks_cli.egg-info/SOURCES.txt
+++ b/databricks_cli.egg-info/SOURCES.txt
@@ -69,10 +69,4 @@ databricks_cli/tokens/cli.py
 databricks_cli/workspace/__init__.py
 databricks_cli/workspace/api.py
 databricks_cli/workspace/cli.py
-databricks_cli/workspace/types.py
-integration/__init__.py
-integration/conftest.py
-integration/dbfs/__init__.py
-integration/dbfs/test_integration.py
-integration/workspace/__init__.py
-integration/workspace/test_integration.py
\ No newline at end of file
+databricks_cli/workspace/types.py
\ No newline at end of file
```

fjakobs (approved):

Issue #475, opened 2022-06-13
Add --wait option to `databricks runs submit` CLI command 


smurching:
In addition to submitting a one-time job run through the `databricks runs submit` CLI, it would be great to additionally support an experimental `--wait` option to await its success (polling job status every ~30s). This would greatly simplify integration testing use cases, allowing us to generalize plugins like https://github.com/databricks/run-notebook across CI/CD providers.

In the future, we might consider:
* Extending this parameter to the `databricks jobs run-now` API to allow awaiting the result of a run of an existing job as well
* Making the polling interval configurable, with some fixed minimum to avoid DOSing the jobs `/runs/get` REST API endpoint used to poll for run status.

The `--wait` option suggestion is based off of a similar API change proposed [in this PR](https://github.com/databricks/run-notebook). This functionality is also similar to that provided by `dbx launch --trace`, but has the added flexibility of being able to run jobs that are not declared in DBX's YAML format & without requiring the caller to install `dbx` and invoke `dbx configure`

smurching:
I just tried implementing this via shell scripts around existing CLI commands - it was pretty tricky. See https://gist.github.com/smurching/aff4e7ed25f8ff2eb9d920cd837f48f8 for details

pietern:
Added in #487.

edmondop:
How about adding it also to the `databricks job run-now`?

PR #487, opened 2022-06-15
Add --wait option to databricks runs submit CLI command

jerrylian-db:
## Description

This PR follows https://github.com/databricks/databricks-cli/issues/475 to implement a --wait option for the runs submit CLI. When this option is enabled, the CLI submits a one-time job run and continues to poll for the job run state until the job run completes.

## Testing

* I added unit tests to cover different branches in the added code.
* I manually tested submitting and waiting for a job run (for both API version 2.0 and 2.1). See screenshot:
Screen Shot 2022-06-15 at 11 11 03 AM
Finally, the job run successfully completes and the terminal output looks like below:
Screen Shot 2022-06-15 at 11 12 57 AM

### DIFF
--- a/databricks_cli/runs/cli.py
+++ b/databricks_cli/runs/cli.py
@@ -31,7 +31,7 @@ from tabulate import tabulate
 from databricks_cli.click_types import OutputClickType, JsonClickType, RunIdClickType
 from databricks_cli.jobs.cli import check_version
 from databricks_cli.utils import eat_exceptions, CONTEXT_SETTINGS, pretty_format, truncate_string, \
-    error_and_quit
+    error_and_quit, backoff_with_jitter
 from databricks_cli.configure.config import provide_api_client, profile_option, debug_option, \
     api_version_option
 from databricks_cli.runs.api import RunsApi
@@ -66,6 +66,8 @@ def submit_cli(api_client, json_file, json, wait, version):
     if wait:
         run_id = submit_res['run_id']
         completed_states = set(['TERMINATED', 'SKIPPED', 'INTERNAL_ERROR'])
+        prev_life_cycle_state = ""
+        attempt = 0
         # Wait for run to complete
         while True:
             run = RunsApi(api_client).get_run(run_id, version=version)
@@ -76,9 +78,13 @@ def submit_cli(api_client, json_file, json, wait, version):
                 else:
                     error_and_quit('Run failed with state ' + run_state['result_state'] +
                                    ' and state message ' + run_state['state_message'])
-            click.echo('Run is still active, with lifecycle state ' +
-                       run_state['life_cycle_state'] + '. URL: ' + run['run_page_url'], err=True)
-            time.sleep(5)
+            if prev_life_cycle_state != run_state['life_cycle_state']:
+                prev_life_cycle_state = run_state['life_cycle_state']
+                click.echo('Waiting on run to complete. Current state: ' +
+                           run_state['life_cycle_state'] + '. URL: ' +
+                           run['run_page_url'], err=True)
+            time.sleep(backoff_with_jitter(attempt))
+            attempt += 1
 
 
 def _runs_to_table(runs_json):
--- a/databricks_cli/runs/cli.py
+++ b/databricks_cli/runs/cli.py
@@ -52,7 +52,7 @@ from databricks_cli.version import print_version_callback, version as cli_versio
 @provide_api_client
 def submit_cli(api_client, json_file, json, wait, version):
     """
-    Submits a one-time run.
+    Submits a one-time run and optionally waits for its completion.
 
     The specification for the request json can be found
     https://docs.databricks.com/api/latest/jobs.html#runs-submit
@@ -74,9 +74,9 @@ def submit_cli(api_client, json_file, json, wait, version):
                 if run_state['result_state'] == 'SUCCESS':
                     sys.exit(0)
                 else:
-                    error_and_quit('job failed with state ' + run_state['result_state'] +
+                    error_and_quit('Run failed with state ' + run_state['result_state'] +
                                    ' and state message ' + run_state['state_message'])
-            click.echo('Job still running with lifecycle state ' + run_state['life_cycle_state'] +
+            click.echo('Run is still active, with lifecycle state ' + run_state['life_cycle_state'] +
                         '. URL: ' + run['run_page_url'], err=True)
             time.sleep(5)
 
--- a/databricks_cli/runs/cli.py
+++ b/databricks_cli/runs/cli.py
@@ -44,7 +44,7 @@ from databricks_cli.version import print_version_callback, version as cli_versio
 @click.option('--json', default=None, type=JsonClickType(),
               help=JsonClickType.help('/api/2.*/jobs/runs/submit'))
 @click.option('--wait', is_flag=True, default=False,
-              help='If specified, wait for the submitted run to complete.')
+              help='If specified, waits for the submitted run to complete.')
 @api_version_option
 @debug_option
 @profile_option
--- a/databricks_cli/runs/cli.py
+++ b/databricks_cli/runs/cli.py
@@ -21,6 +21,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
 import time
 from json import loads as json_loads
 
@@ -29,7 +30,8 @@ from tabulate import tabulate
 
 from databricks_cli.click_types import OutputClickType, JsonClickType, RunIdClickType
 from databricks_cli.jobs.cli import check_version
-from databricks_cli.utils import eat_exceptions, CONTEXT_SETTINGS, pretty_format, truncate_string
+from databricks_cli.utils import eat_exceptions, CONTEXT_SETTINGS, pretty_format, truncate_string, \
+    error_and_quit
 from databricks_cli.configure.config import provide_api_client, profile_option, debug_option, \
     api_version_option
 from databricks_cli.runs.api import RunsApi
@@ -71,14 +73,12 @@ def submit_cli(api_client, json_file, json, wait, version):
             run_state = run['state']
             if run_state['life_cycle_state'] in completed_states:
                 if run_state['result_state'] == 'SUCCESS':
-                    click.echo('Job completed successfully.')
+                    sys.exit(0)
                 else:
-                    click.echo(click.style('ERROR', fg='red') + ': job failed with state ' +
-                                run_state['result_state'] + ' and state message ' +
-                                run_state['state_message'], err=True)
-                return
+                    error_and_quit('job failed with state ' + run_state['result_state'] +
+                                   ' and state message ' + run_state['state_message'])
             click.echo('Job still running with lifecycle state ' + run_state['life_cycle_state'] +
-                        '. URL: ' + run['run_page_url'])
+                        '. URL: ' + run['run_page_url'], err=True)
             time.sleep(5)
 
 
--- a/databricks_cli/runs/cli.py
+++ b/databricks_cli/runs/cli.py
@@ -57,12 +57,6 @@ def submit_cli(api_client, json_file, json, wait, version):
     https://docs.databricks.com/api/latest/jobs.html#runs-submit
     """
     check_version(api_client, version)
-    api_version = version or api_client.jobs_api_version
-    if api_version != '2.1' and wait:
-        click.echo(click.style('ERROR', fg='red') + ': the option --wait is only available ' +
-                   'in API 2.1', err=True)
-        return
-
     if json_file:
         with open(json_file, 'r') as f:
             json = f.read()
--- a/databricks_cli/runs/cli.py
+++ b/databricks_cli/runs/cli.py
@@ -44,8 +44,7 @@ from databricks_cli.version import print_version_callback, version as cli_versio
 @click.option('--json', default=None, type=JsonClickType(),
               help=JsonClickType.help('/api/2.*/jobs/runs/submit'))
 @click.option('--wait', is_flag=True,
-              help='If specified, the CLI will wait for the submitted run to complete ' +
-                   '(only available in API 2.1).')
+              help='If specified, the CLI will wait for the submitted run to complete.')
 @api_version_option
 @debug_option
 @profile_option
--- a/databricks_cli/runs/cli.py
+++ b/databricks_cli/runs/cli.py
@@ -22,15 +22,14 @@
 # limitations under the License.
 
 import time
-from json import dumps as json_dumps, loads as json_loads
+from json import loads as json_loads
 
 import click
 from tabulate import tabulate
 
 from databricks_cli.click_types import OutputClickType, JsonClickType, RunIdClickType
 from databricks_cli.jobs.cli import check_version
-from databricks_cli.utils import eat_exceptions, CONTEXT_SETTINGS, pretty_format, json_cli_base, \
-    truncate_string
+from databricks_cli.utils import eat_exceptions, CONTEXT_SETTINGS, pretty_format, truncate_string
 from databricks_cli.configure.config import provide_api_client, profile_option, debug_option, \
     api_version_option
 from databricks_cli.runs.api import RunsApi
--- a/databricks_cli/runs/cli.py
+++ b/databricks_cli/runs/cli.py
@@ -72,17 +72,17 @@ def submit_cli(api_client, json_file, json, wait, version):
         while True:
             run = RunsApi(api_client).get_run(run_id, version=version)
             run_state = run['state']
-            if run_state['life_cycle_state'] in completed_states:
+            life_cycle_state = run_state['life_cycle_state']
+            if life_cycle_state in completed_states:
                 if run_state['result_state'] == 'SUCCESS':
                     sys.exit(0)
                 else:
                     error_and_quit('Run failed with state ' + run_state['result_state'] +
                                    ' and state message ' + run_state['state_message'])
-            if prev_life_cycle_state != run_state['life_cycle_state']:
-                prev_life_cycle_state = run_state['life_cycle_state']
-                click.echo('Waiting on run to complete. Current state: ' +
-                           run_state['life_cycle_state'] + '. URL: ' +
-                           run['run_page_url'], err=True)
+            if prev_life_cycle_state != life_cycle_state:
+                click.echo('Waiting on run to complete. Current state: ' + life_cycle_state +
+                           '. URL: ' + run['run_page_url'], err=True)
+                prev_life_cycle_state = life_cycle_state
             time.sleep(backoff_with_jitter(attempt))
             attempt += 1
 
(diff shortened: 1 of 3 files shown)


databricks_cli/runs/cli.py
```diff
@@ -39,21 +41,51 @@
...
     The specification for the request json can be found
     https://docs.databricks.com/api/latest/jobs.html#runs-submit
     """
     check_version(api_client, version)
-    json_cli_base(json_file, json, lambda json: RunsApi(
-        api_client).submit_run(json, version=version))
+    api_version = version or api_client.jobs_api_version
+    if api_version != '2.1' and wait:
+        click.echo(click.style('ERROR', fg='red') + ': the option --wait is only available ' +
+                   'in API 2.1', err=True)
```
pietern, line 63: Why?

databricks_cli/runs/cli.py
```diff
@@ -39,21 +41,51 @@
...
+        while True:
+            run = RunsApi(api_client).get_run(run_id, version=version)
+            run_state = run['state']
+            if run_state['life_cycle_state'] in completed_states:
+                if run_state['result_state'] == 'SUCCESS':
+                    click.echo('Job completed successfully.')
+                else:
+                    click.echo(click.style('ERROR', fg='red') + ': job failed with state ' +
+                                run_state['result_state'] + ' and state message ' +
+                                run_state['state_message'], err=True)
```
pietern, line 84: This should exit with status 1.

databricks_cli/runs/cli.py
```diff
@@ -39,21 +41,51 @@
...
+    if api_version != '2.1' and wait:
+        click.echo(click.style('ERROR', fg='red') + ': the option --wait is only available ' +
+                   'in API 2.1', err=True)
+        return
+
+    if json_file:
+        with open(json_file, 'r') as f:
+            json = f.read()
+    submit_res = RunsApi(api_client).submit_run(json_loads(json), version=version)
+    click.echo(pretty_format(submit_res))
```
pietern, line 70: Does it make sense to echo this?

databricks_cli/runs/cli.py
```diff
@@ -39,21 +41,51 @@
...
+    if wait:
+        run_id = submit_res['run_id']
+        completed_states = set(['TERMINATED', 'SKIPPED', 'INTERNAL_ERROR'])
+        # Wait for run to complete
+        while True:
+            run = RunsApi(api_client).get_run(run_id, version=version)
+            run_state = run['state']
+            if run_state['life_cycle_state'] in completed_states:
+                if run_state['result_state'] == 'SUCCESS':
+                    click.echo('Job completed successfully.')
```
pietern, line 80: Success can be implied by exiting with status 0, no need to echo anything.

databricks_cli/runs/cli.py
```diff
@@ -39,21 +41,51 @@
...
+                else:
+                    click.echo(click.style('ERROR', fg='red') + ': job failed with state ' +
+                                run_state['result_state'] + ' and state message ' +
+                                run_state['state_message'], err=True)
+                return
+            click.echo('Job still running with lifecycle state ' + run_state['life_cycle_state'] +
+                        '. URL: ' + run['run_page_url'])
```
pietern, line 87: This seems like debug output. WDYT?

jerrylian-db: This is not. We are following the behavior of the run-notebook open source GitHub action to continuous print the status and link of the job run.

databricks_cli/runs/cli.py
jerrylian-db: For this, I am just following the original behavior of the run submit CLI, which is this:
Screen Shot 2022-06-15 at 11 37 35 AM

databricks_cli/runs/cli.py
smurching: @jerrylian-db in the CLI I agree with @pietern that this should be debug output - in the run-notebook Action, we continuously print the URL due to a limitation in GitHub Action output, which is that only output generated after the UI is loaded is shown (so if we don't continuously print output users won't see it if they load UI after the page loads).

databricks_cli/runs/cli.py
jerrylian-db: I was concerned that runs-get call will have a different response structure between `2.1` and `2.0`. However, I tested the `--wait` option on `2.0` and it still worked. So I will remove this check.

databricks_cli/runs/cli.py
pietern: Either way, please redirect to stderr. If you want to use the `run_id` from the JSON body printed above, you won't be able to if we echo other stuff to stdout.

jerrylian-db: I've redirected this echo to stderr

databricks_cli/runs/cli.py
```diff
@@ -39,21 +43,42 @@
...
+        completed_states = set(['TERMINATED', 'SKIPPED', 'INTERNAL_ERROR'])
+        # Wait for run to complete
+        while True:
+            run = RunsApi(api_client).get_run(run_id, version=version)
+            run_state = run['state']
+            if run_state['life_cycle_state'] in completed_states:
+                if run_state['result_state'] == 'SUCCESS':
+                    sys.exit(0)
+                else:
+                    error_and_quit('job failed with state ' + run_state['result_state'] +
```
jerrylian-db, line 77: QQ: `error_and_quit` currently doesn't not print to stderr. See definition below. Should I change it to do so?
```
def error_and_quit(message):
    ctx = click.get_current_context()
    context_object = ctx.ensure_object(ContextObject)
    if context_object.debug_mode:
        traceback.print_exc()
    click.echo(u'Error: {}'.format(message))
    sys.exit(1)
```

smurching, line 77:
```suggestion
                    error_and_quit('Job failed with state ' + run_state['result_state'] +
```

Nit: capitalization consistency?

databricks_cli/runs/cli.py
```diff
@@ -39,21 +43,42 @@
...
+                                   ' and state message ' + run_state['state_message'])
+            click.echo('Job still running with lifecycle state ' + run_state['life_cycle_state'] +
```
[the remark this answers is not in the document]
smurching: Switching to "Run" instead of "Job" and using terminology from our docs around "active runs"

tests/runs/test_cli.py
```diff
@@ -55,6 +74,38 @@ def test_submit_cli_json(runs_api_mock):
...
+def test_submit_wait_success(runs_api_mock):
+    runs_api_mock.submit_run.return_value = SUBMIT_RETURN
+    runs_api_mock.get_run.return_value = RUNS_GET_RETURN_SUCCESS
+    runner = CliRunner()
+    result = runner.invoke(cli.submit_cli, ['--json', SUBMIT_JSON, '--wait'])
+    runs_api_mock.get_run.assert_called_once()
+    assert result.exit_code == 0
+
+@provide_conf
+def test_submit_wait_failure(runs_api_mock):
```
smurching, line 87: Should we also add a test for eventual failure, and for other terminal states besides `TERMINATED` (i.e. test the `INTERNAL_ERROR` and `SKIPPED` life_cycle_state cases)? You can use the `pytest.mark.parametrize` API to parametrize your test across different terminal states ([see docs](https://docs.pytest.org/en/6.2.x/parametrize.html)).

smurching (commented):
Just minor comments, but mostly looks good!

databricks_cli/runs/cli.py
```diff
@@ -39,21 +43,42 @@
...
+    if json_file:
+        with open(json_file, 'r') as f:
+            json = f.read()
+    submit_res = RunsApi(api_client).submit_run(json_loads(json), version=version)
+    click.echo(pretty_format(submit_res))
+    if wait:
+        run_id = submit_res['run_id']
+        completed_states = set(['TERMINATED', 'SKIPPED', 'INTERNAL_ERROR'])
+        # Wait for run to complete
+        while True:
```
vladimirk-db, line 70: I wonder if we should have a time-out? So it's not perpetually waiting if something goes wrong.

databricks_cli/runs/cli.py
```diff
@@ -39,21 +43,42 @@
...
+              help='If specified, the CLI will wait for the submitted run to complete.')
 @api_version_option
 @debug_option
 @profile_option
 @eat_exceptions
 @provide_api_client
-def submit_cli(api_client, json_file, json, version):
+def submit_cli(api_client, json_file, json, wait, version):
     """
     Submits a one-time run.
```
vladimirk-db, line 55: Should we update documentation ("... and waits for its completion") ?

tests/runs/test_cli.py
jerrylian-db: Done

databricks_cli/runs/cli.py
jerrylian-db: Leaning towards not having a time-out for now. I believe that the submitted run themselves have an internal timeouts in Databricks. Users can also force exit on their own.

vladimirk-db:
Successful test run output against a test shard:

```
bazel run //mlflow/src/test/e2e:DefaultRepoNotebookExperimentIntegrationSuite_test_shard -- --context dev-aws-eu-west-1 --test-shard-name test-shard-vladimir-kolovski --dev-mode true
...
...bunch of INFO output...
...
Iteration 1 succeeded
2022/06/14 17:44:15 INFO DefaultRepoNotebookExperimentIntegrationSuite[class com.databricks.mlflow.e2e.DefaultRepoNotebookExperimentIntegrationSuite-0] [tenant=1382500038054285 root=DustSuite-51a815007a1c0045 parent=Thread-1 op=DustSuite-51a815007a1c0045 traceId=86efc248a99b52fbb53f962096dfd0d0 spanId=e503827fcadc28c7]: Iteration 1 succeeded
All 1 iterations succeeded
2022/06/14 17:44:15 INFO DefaultRepoNotebookExperimentIntegrationSuite[class com.databricks.mlflow.e2e.DefaultRepoNotebookExperimentIntegrationSuite-0] [tenant=1382500038054285 root=DustSuite-51a815007a1c0045 parent=Thread-1 op=DustSuite-51a815007a1c0045 traceId=86efc248a99b52fbb53f962096dfd0d0 spanId=e503827fcadc28c7]: All 1 iterations succeeded
Run completed in 6 minutes, 35 seconds.
Total number of tests run: 5
Suites: completed 1, aborted 0
Tests: succeeded 5, failed 0, canceled 0, ignored 0, pending 0
All tests passed.

```

tests/runs/test_cli.py
```diff
@@ -55,6 +67,45 @@ def test_submit_cli_json(runs_api_mock):
...
+    result = runner.invoke(cli.submit_cli, ['--json', SUBMIT_JSON, '--wait'])
+    runs_api_mock.get_run.assert_called_once()
+    assert result.exit_code == 0
+
+@pytest.mark.parametrize('bad_life_cycle_state', ['TERMINATED', 'SKIPPED', 'INTERNAL_ERROR'])
+@provide_conf
+def test_submit_wait_failure(runs_api_mock, bad_life_cycle_state):
+    with mock.patch('click.echo') as echo_mock:
+        runs_api_mock.submit_run.return_value = SUBMIT_RETURN
+        runs_api_mock.get_run.return_value = {
```
smurching, line 84: Nit: Would you mind mocking a PENDING / RUNNING request before job failure just to make this test case a bit more realistic & cover a few more run states? 

e.g. something like

```
runs_get_failed = {
            "state": {
                "life_cycle_state": bad_life_cycle_state,
                "result_state": "FAILED",
                "state_message": "OH NO!",
            },
}
runs_api_mock.get_run.side_effect = [RUNS_GET_RETURN_PENDING, RUNS_GET_RETURN_PENDING, RUNS_GET_RETURN_RUNNING, runs_get_failed]
```

tests/runs/test_cli.py
```diff
@@ -55,6 +67,45 @@ def test_submit_cli_json(runs_api_mock):
...
+        result = runner.invoke(cli.submit_cli, ['--json', SUBMIT_JSON, '--wait'])
+        assert 'Run failed with state FAILED and state message OH NO!' in echo_mock.call_args[0][0]
+        assert result.exit_code == 1
+
+@provide_conf
+def test_submit_wait_eventually_succeeds(runs_api_mock):
+    with mock.patch('databricks_cli.runs.cli.click.echo') as echo_mock, \
+         mock.patch('time.sleep') as sleep_mock:
+        runs_api_mock.submit_run.return_value = SUBMIT_RETURN
+        runs_api_mock.get_run.side_effect = [RUNS_GET_RETURN_RUNNING, RUNS_GET_RETURN_SUCCESS]
```
smurching, line 101: Similar comment here, it might be nice to add a `PENDING` response before the `RUNNING` response

smurching (commented):
Last few comments, after that LGTM, thanks @jerrylian-db !

smurching (approved):

databricks_cli/runs/cli.py
```diff
@@ -39,21 +43,42 @@
               help='File containing JSON request to POST to /api/2.*/jobs/runs/submit.')
 @click.option('--json', default=None, type=JsonClickType(),
               help=JsonClickType.help('/api/2.*/jobs/runs/submit'))
+@click.option('--wait', is_flag=True,
+              help='If specified, the CLI will wait for the submitted run to complete.')
```
pietern, line 47: Nit: `the CLI will` is superfluous. Please also add a `default=False` to the option for completeness.

databricks_cli/runs/cli.py
```diff
@@ -39,21 +43,42 @@
...
+            run_state = run['state']
+            if run_state['life_cycle_state'] in completed_states:
+                if run_state['result_state'] == 'SUCCESS':
+                    sys.exit(0)
+                else:
+                    error_and_quit('job failed with state ' + run_state['result_state'] +
+                                   ' and state message ' + run_state['state_message'])
+            click.echo('Job still running with lifecycle state ' + run_state['life_cycle_state'] +
+                        '. URL: ' + run['run_page_url'], err=True)
+            time.sleep(5)
```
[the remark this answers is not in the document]
pietern: I think this is fine. The same is used elsewhere (e.g. dbx and your GH action).

Beware that there are uses for this beyond integration tests.

pietern (commented):
One final nit. Everything else LGTM.

We can release this as 0.16.10.

databricks_cli/runs/cli.py
fjakobs: +1 Users can always CTRL-C

fjakobs (changes requested):
It would be great if we could also print the stdout/stderr of the job after it has finished. If you e.g. run the job in order to run some unit tests you will be interested in the outcome without having to go to the Databricks web UI.

databricks_cli/runs/cli.py
```diff
@@ -39,21 +43,42 @@
...
+            run_state = run['state']
+            if run_state['life_cycle_state'] in completed_states:
+                if run_state['result_state'] == 'SUCCESS':
+                    sys.exit(0)
+                else:
+                    error_and_quit('Run failed with state ' + run_state['result_state'] +
+                                   ' and state message ' + run_state['state_message'])
+            click.echo('Run is still active, with lifecycle state ' +
+                       run_state['life_cycle_state'] + '. URL: ' + run['run_page_url'], err=True)
+            time.sleep(5)
```
lennartkats-db, line 81: 5 seconds is fine as a starting point but we do want exponential backoff and jitter as discussed in the thread. For longer-running jobs (and they can take many hours) I would bias towards backoff of up to 20-30s.

lennartkats-db (changes requested):

databricks_cli/runs/cli.py
```diff
@@ -39,21 +43,42 @@
...
+        while True:
+            run = RunsApi(api_client).get_run(run_id, version=version)
+            run_state = run['state']
+            if run_state['life_cycle_state'] in completed_states:
+                if run_state['result_state'] == 'SUCCESS':
+                    sys.exit(0)
+                else:
+                    error_and_quit('Run failed with state ' + run_state['result_state'] +
+                                   ' and state message ' + run_state['state_message'])
+            click.echo('Run is still active, with lifecycle state ' +
```
lennartkats-db, line 79: This seems to print text with every single ping? Is it not more user friendly to just print on state changes as seen in the Airbreeze version?

lennartkats-db (changes requested):

jerrylian-db:
@fjakobs wondering if it would be best to ask users to call `runs get_output` after the run completes rather than we print the output in `runs submit --wait`. Right now, the only stdout print of this `runs submit --wait` is the run-id in JSON format, which is easy for users to parse in workflows. Printing more stuff to stout would complicate that.

databricks_cli/runs/cli.py
jerrylian-db: I agree that it is a lot of text. However, maybe given my experience with the GitHub, I also find printing the ping results as comforting. It gives me a sense of transparency that the CLI is still working and pinging.

jerrylian-db: I can envision a fancier user experience where there is a countdown text and spinner in the CLI that shows how many seconds until the next ping and what the latest state result is.

jerrylian-db: @lennartkats-db how would you feel about going with this rudimentary printing for now?

jerrylian-db: After some discussion with @smurching, I realize that users can simply click on the run page link to see if the run is still active. Will work on making the print text more concise.

jerrylian-db: Sounds good. I've implemented a more concise printing now.

databricks_cli/runs/cli.py
```diff
@@ -39,21 +43,48 @@
...
+    if wait:
+        run_id = submit_res['run_id']
+        completed_states = set(['TERMINATED', 'SKIPPED', 'INTERNAL_ERROR'])
+        prev_life_cycle_state = ""
+        attempt = 0
+        # Wait for run to complete
+        while True:
+            run = RunsApi(api_client).get_run(run_id, version=version)
+            run_state = run['state']
+            if run_state['life_cycle_state'] in completed_states:
```
smurching, line 75: Super minor nit: maybe do `life_cycle_state = run_state['life_cycle_state']` before this so that we can just do the dictionary access once (helps prevent typos etc with the hardcoded 'life_cycle_state' string)

databricks_cli/runs/cli.py
lennartkats-db: 👍 Ok, thanks for changing this. I do like the new version better!

Note that getting this printing right is important for the developer experience. As a non-blocking comment, I'd consider tinkering a bit more with it even to get it just right:
* I wouldn't print the very long URL with every message printed
* I'd consider printing a "." without a newline every now and then to show progress. A spinner would be even better, but then you do need to make sure that no one is capturing the output (all the ANSI codes would otherwise mess up CI/CD logs). And I'm not sure you want to build that yourself, and pulling in dependencies is also something we'd rather avoid at this point since we install in the global Python namespace.

fjakobs:
@jerrylian-db I've created a separate ticket for the output handling. https://github.com/databricks/databricks-cli/issues/489

When customers provide `--wait` I would expect the CLI to print the response from the last "get run" API call, which is a superset of the "submit job" response.

lennartkats-db (approved):

PR #490, opened 2022-06-16
addition of git source to create_job

mattias-de-coninck:
Addition of git_source parameter for the create_job function


pietern (approved):
Thanks for the PR, @mattias-de-coninck!

PR #496, opened 2022-06-20
Add uc_api_version to ApiClient, UC CLI

adamcain-db:
* Adds `uc_api_version` to `ApiClient`, defaulting it to `2.1`.
* Adds `unity-catalog configure` CLI with `--version` option, to allow user to configure a profile to use `2.0` as UC API version.
* Unit tests for the above.


adamcain-db:
@pietern: could you review?  (I am unable to add any reviewers to this PR, for some reason)

PR #498, opened 2022-06-21
Use 2.1 API version for UC paths

adamcain-db:
Replaces PR #496

* Use 2.1 for API version with UC API paths (`/unity-catalog` and `/lineage-tracking`)
* Simple test for the above.
* Tested with various `unity-catalog` commands with `--debug` option.


pietern (commented):
Do any of the files under `/examples/unity_catalog` need to be updated as well?

adamcain-db:
> Do any of the files under /examples/unity_catalog need to be updated as well?

Good point.  We will soon want to change the permissions in `update-permissions1.json` and rename it to clarify which object it applies to.  However, the UC service doesn't yet support those new permissions on input.  So that will need to be in a follow-up change.

pietern (approved):

PR #503, opened 2022-06-28
Fix external-location 'lists' command

adamcain-db:
* Correct unity-catalog `external-locations lists` -> `external-locations list` CLI


pietern (approved):
Thanks for fixing!

PR #505, opened 2022-06-30
Regenerate service.py

pietern:
First pass to ensure better completeness of the arguments to the methods in `service.py`.

### DIFF
--- a/databricks_cli/sdk/service.py
+++ b/databricks_cli/sdk/service.py
@@ -33,11 +33,28 @@ class JobsService(object):
     def __init__(self, client):
         self.client = client
 
-    def create_job(self, name=None, existing_cluster_id=None, new_cluster=None, libraries=None,
-                   email_notifications=None, timeout_seconds=None, max_retries=None,
-                   min_retry_interval_millis=None, retry_on_timeout=None, schedule=None,
-                   notebook_task=None, spark_jar_task=None, spark_python_task=None,
-                   spark_submit_task=None, max_concurrent_runs=None, tasks=None, headers=None, version=None, git_source=None):
+    def create_job(
+        self,
+        name=None,
+        existing_cluster_id=None,
+        new_cluster=None,
+        libraries=None,
+        email_notifications=None,
+        timeout_seconds=None,
+        max_retries=None,
+        min_retry_interval_millis=None,
+        retry_on_timeout=None,
+        schedule=None,
+        notebook_task=None,
+        spark_jar_task=None,
+        spark_python_task=None,
+        spark_submit_task=None,
+        max_concurrent_runs=None,
+        tasks=None,
+        headers=None,
+        version=None,
+        git_source=None,
+    ):
         _data = {}
         if name is not None:
             _data['name'] = name
@@ -52,7 +69,9 @@ class JobsService(object):
         if email_notifications is not None:
             _data['email_notifications'] = email_notifications
             if not isinstance(email_notifications, dict):
-                raise TypeError('Expected databricks.JobEmailNotifications() or dict for field email_notifications')
+                raise TypeError(
+                    'Expected databricks.JobEmailNotifications() or dict for field email_notifications'
+                )
         if timeout_seconds is not None:
             _data['timeout_seconds'] = timeout_seconds
         if max_retries is not None:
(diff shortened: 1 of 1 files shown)


databricks_cli/sdk/service.py
```diff
@@ -202,19 +202,17 @@ def get_job(self, job_id, headers=None, version=None):
         )
 
     def list_jobs(
-        self, job_type=None, expand_tasks=None, limit=None, offset=None, headers=None, version=None
+        self, job_type=None, offset=None, limit=None, expand_tasks=None, headers=None, version=None
```
pietern, line 205: Order of these kwargs has changed. I doubt there are call sites where all arguments are specified sequentially and where this change proves to be incompatible.

Regardless, it justifies a minor bump.

alexott: I saw people using databricks-cli as SDK, plus it's used in package like https://pypi.org/project/databricks-api/, so it could be a breaking change

databricks_cli/sdk/service.py
```diff
@@ -170,16 +238,26 @@ def run_now(self, job_id=None, jar_params=None, notebook_params=None, python_par
             _data['notebook_params'] = notebook_params
         if python_params is not None:
             _data['python_params'] = python_params
-        if python_named_params is not None:
-            _data['python_named_params'] = python_named_params
```
alexott, line 174: Why we need to move it? having python params close to each other is easier to read

databricks_cli/sdk/service.py
```diff
@@ -232,15 +322,34 @@ def __init__(self, client):
...
+    def create_cluster(
+        self,
+        num_workers=None,
+        autoscale=None,
+        cluster_name=None,
+        spark_version=None,
+        spark_conf=None,
+        aws_attributes=None,
+        azure_attributes=None,
+        gcp_attributes=None,
```
alexott, line 336: What about adding them after headers? Because it's again change in the signature

databricks_cli/sdk/service.py
```diff
@@ -232,15 +322,34 @@ def __init__(self, client):
...
+        cluster_log_conf=None,
+        spark_env_vars=None,
+        autotermination_minutes=None,
+        enable_elastic_disk=None,
+        cluster_source=None,
+        instance_pool_id=None,
+        policy_id=None,
+        enable_local_disk_encryption=None,
+        driver_instance_pool_id=None,
+        apply_policy_default_values=None,
```
alexott, line 350: same here...

databricks_cli/sdk/service.py
```diff
@@ -323,12 +459,33 @@ def resize_cluster(self, cluster_id, num_workers=None, autoscale=None, headers=N
...
+    def edit_cluster(
+        self,
+        cluster_id,
+        num_workers=None,
+        autoscale=None,
+        cluster_name=None,
+        spark_version=None,
+        spark_conf=None,
+        aws_attributes=None,
+        azure_attributes=None,
```
alexott, line 471: also signature change

alexott (commented):
Should we minimize the changes in the signatures? Or we need to bump version & declare it as a breaking change

databricks_cli/sdk/service.py
pietern: That would mean we can never add fields to the existing functions. Only way to circumvent is by using some kind of builder, but that's too big of a departure from what we're working with today.

alexott: no, it's ok to add them - just at the end, so it won't break existing things if people are using positional parameters

databricks_cli/sdk/service.py
pietern: Fixed in the latest iteration.

databricks_cli/sdk/service.py
pietern: In the latest iteration of the PR all new arguments are appended to the list.

pietern:
@alexott

I did some more investigation and found a couple call sites that use positional parameters and would be broken by this change. Unfortunately we cannot start forcing keyword arguments (e.g. with `*` separator argument) or we'd end up breaking these call sites. To address this I added a unit test to explicitly test for positional argument ordering in #507 and use the output of this test to generate `service.py`.

All new keyword arguments are appended so none of the existing call sites should break.

databricks_cli/sdk/service.py
```diff
@@ -52,7 +72,9 @@ def create_job(self, name=None, existing_cluster_id=None, new_cluster=None, libr
         if email_notifications is not None:
             _data['email_notifications'] = email_notifications
             if not isinstance(email_notifications, dict):
-                raise TypeError('Expected databricks.JobEmailNotifications() or dict for field email_notifications')
+                raise TypeError(
+                    'Expected databricks.JobEmailNotifications() or dict for field email_notifications'
```
nfx, line 76: `databricks` package does not exist, neither does `JobEmailNotifications` class, so this message has to be just `Expected dict for field email_notifications`. Same for other `TypeError`s, so that it doesn't confuse developers.

databricks_cli/sdk/service.py
```diff
@@ -106,24 +161,42 @@ def submit_run(self, run_name=None, existing_cluster_id=None, new_cluster=None,
...
+        if idempotency_token is not None:
+            _data['idempotency_token'] = idempotency_token
+        if job_clusters is not None:
+            _data['job_clusters'] = job_clusters
+        if git_source is not None:
+            _data['git_source'] = git_source
+            if not isinstance(git_source, dict):
+                raise TypeError('Expected databricks.GitSource() or dict for field git_source')
+        return self.client.perform_query(
+            'POST', '/jobs/runs/submit', data=_data, headers=headers, version=version
```
nfx, line 198: `version` here should really be propagated as explicit path, e.g. `/api/2.1/jobs/runs/submit`...

databricks_cli/sdk/service.py
```diff
@@ -106,24 +161,42 @@ def submit_run(self, run_name=None, existing_cluster_id=None, new_cluster=None,
...
         if tasks is not None:
             _data['tasks'] = tasks
-        return self.client.perform_query('POST', '/jobs/runs/submit', data=_data, headers=headers, version=version)
+        if idempotency_token is not None:
+            _data['idempotency_token'] = idempotency_token
+        if job_clusters is not None:
+            _data['job_clusters'] = job_clusters
+        if git_source is not None:
+            _data['git_source'] = git_source
+            if not isinstance(git_source, dict):
```
nfx, line 195: perhaps throw error _before_ assigning to `_data`?

nfx (changes requested):

databricks_cli/sdk/service.py
pietern: I filed https://github.com/databricks/databricks-cli/issues/508 to track.

I want to keep this PR focused on regeneration alone.

databricks_cli/sdk/service.py
pietern: Perhaps, only that's not the way it's done in the CLI today.

I agree that we should have version scoped services. I.e. you instantiate the service object with the version, not define it on a per method basis. But that's a much larger refactor than should happen here. In fact, I don't think it's worth to do that work here but rather deal with it in vnext.

databricks_cli/sdk/service.py
pietern: Yeah, doesn't make a difference. I kept everything consistent with existing code in this PR.

databricks_cli/sdk/service.py
```diff
@@ -325,12 +532,33 @@ def resize_cluster(self, cluster_id, num_workers=None, autoscale=None, headers=N
...
-    def edit_cluster(self, cluster_id, num_workers=None, autoscale=None, cluster_name=None,
-                     spark_version=None, spark_conf=None, aws_attributes=None, node_type_id=None,
-                     driver_node_type_id=None, ssh_public_keys=None, custom_tags=None,
-                     cluster_log_conf=None, spark_env_vars=None, autotermination_minutes=None,
-                     enable_elastic_disk=None, cluster_source=None, instance_pool_id=None,
-                     headers=None):
+    def edit_cluster(
+        self,
+        cluster_id,
+        num_workers=None,
```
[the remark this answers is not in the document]
pietern: Unfortunately no, because of positional arguments. It would break call sites that use those, so it's a breaking change.

databricks_cli/sdk/service.py
```diff
@@ -572,7 +849,9 @@ def put_test(self, path, contents=None, overwrite=None, headers=None, src_path=N
             headers = {'Content-Type': None}
             filename = os.path.basename(src_path)
             _files = {'file': (filename, open(src_path, 'rb'), 'multipart/form-data')}
-        return self.client.perform_query('POST', '/dbfs/put', data=_data, headers=headers, files=_files)
+        return self.client.perform_query(
```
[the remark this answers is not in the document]
pietern: I agree. Filed #509.

pietern:
@nfx Filed follow ups for the issues not in scope here.

nfx (approved):

PR #507, opened 2022-07-01
Add tests to ensure API compatibility when using positional arguments

pietern:
This test requires the environment variable `DATABRICKS_CLI_MAIN_CHECKOUT` to be set to the path of a reference checkout of this repository. It will dynamically load the argspecs from both the reference repository and the current repository and compare them. The test fails if the current repository contains incompatible modifications of function signatures.

In CI we can run this test with `DATABRICKS_CLI_MAIN_CHECKOUT` pointing to a checkout of the main branch. It will test compatibility of the public API in the PR against the public API in the main branch, and fail on error.


tests/test_compat.py
```diff
@@ -0,0 +1,125 @@
...
+        databricks_cli.runs.api,
+        databricks_cli.secrets.api,
+        databricks_cli.tokens.api,
+        databricks_cli.unity_catalog.api,
+        databricks_cli.workspace.api,
+    ]
+
+    current_argspecs = collect_argspecs([databricks_cli.sdk.service] + api_packages)
+
+    test_compatibility_file = os.getenv('TEST_COMPATIBILITY_FILE', default=None)
```
nfx, line 116: Why not determining this dynamically? it looks like a hack and pain to maintain.

Why not checking out `main` branch in a test and discovering argspecs from there? 

* you can either use `ast` package, though `collect_argspecs` would have to be rewritten.
* or make that checkout quack like a python sub-package, though `sys.path.modules` or something, so that `collect_argspecs` is only slightly modified.

nfx (changes requested):

tests/test_compat.py
pietern: I like option 2. I'll give it a shot.

nfx (approved):
It'll also be great if you add a shell script to run this locally, not only on gha

pietern:
Added a runner script at `scripts/test_compatibility.sh`. It works out of the box.

PR #511, opened 2022-07-07
Add flag for UC get-effective-permissions

adamcain-db:
* Add --effective / -e flag to unity-catalog `get-permissions` CLI which causes it to use UC's `getEffectivePermissions` endpoint.

### DIFF
--- a/databricks_cli/unity_catalog/perms_cli.py
+++ b/databricks_cli/unity_catalog/perms_cli.py
@@ -67,7 +67,8 @@ def _get_perm_securable_name_and_type(catalog_name, schema_full_name, table_full
 @click.option('--external-location', cls=OneOfOption, default=None,
               one_of=PERMISSIONS_OBJ_TYPES,
               help='Name of the external location of interest')
-@click.option('--effective', '-e', is_flag=True, default=False)
+@click.option('--effective', is_flag=True, default=False,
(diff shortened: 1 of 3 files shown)


cpettitt-databricks (approved):

BoopBoopBeepBoop (approved):

databricks_cli/unity_catalog/perms_cli.py
```diff
@@ -67,19 +67,23 @@ def _get_perm_securable_name_and_type(catalog_name, schema_full_name, table_full
 @click.option('--external-location', cls=OneOfOption, default=None,
               one_of=PERMISSIONS_OBJ_TYPES,
               help='Name of the external location of interest')
+@click.option('--effective', '-e', is_flag=True, default=False)
```
pietern, line 70: The short form option doesn't seem like a good idea, since it naturally conflicts with `--external-location`.

WDYT?

adamcain-db: Good point. I removed the `-e` flag (and added some help text) for that command.

pietern (approved):
Thanks, @adamcain-db. For future PRs, having a couple of tests to confirm the intended behavior would be nice. There are no tests for the entire UC subcommand at the moment.

Issue #512, opened 2022-07-08
Add `-R` recursive option to `fs ls` command


justinTM:
currently i'm just looping results of `fs ls` and appending results of command to the end of the root path before running the command again until i reach the file I'm looking for across all subdirectories.

what would be nice is a `-R` option to recursively list subdirectories of a DBFS path. 

```console
$ databricks fs ls -R --absolute "dbfs:/job-logs"
dbfs:/job-logs/0708-112403-f1tsd6as8/executor/app-20220708042345808-0000/0/stderr
dbfs:/job-logs/0708-112403-f1tsd6as8/executor/app-20220708042345808-0000/0/stdout
dbfs:/job-logs/0708-112403-f1tsd6as8/executor/app-20220708042345808-0000/1/stdout
dbfs:/job-logs/0708-112403-f1tsd6as8/executor/app-20220708042345808-0000/1/stderr
```

justinTM:
https://github.com/databricks/databricks-cli/pull/513

chrisst:
**This repository has been archived.**

The legacy `databricks/databricks-cli` is deprecated and has been fully replaced by the new Databricks CLI at **https://github.com/databricks/cli**.

We are closing all open issues here as part of archiving this repository. If your issue still affects the new CLI, we'd appreciate you re-opening it at https://github.com/databricks/cli/issues so the team can track and address it there.

PR #513, opened 2022-07-08
Add `--recursive` option to `fs ls` command

justinTM:


pietern (commented):
Thanks for your PR, this looks useful.

Two general comments:
* It seems to me that recursive implies absolute, or at least relative to the specified path. As is, it would allow for recursively showing the basename only, which doesn't make a lot of sense. Can you make the behavior of non-absolute recursive listing show the relative paths?
* Please add a couple tests for this behavior.

databricks_cli/dbfs/cli.py
```diff
@@ -53,10 +55,20 @@ def ls_cli(api_client, l, absolute, dbfs_path): #  NOQA
...
+        table = tabulate([f.to_row(is_long_form=l, is_absolute=absolute) for f in files],
+                         tablefmt='plain')
+        click.echo(table)
+    
+    def recursive_echo(this_dbfs_path):
+      files = DbfsApi(api_client).list_files(this_dbfs_path)
+      echo_path(files)
+      for f in files:
+          if f.is_dir:
+              recursive_echo(this_dbfs_path.join(f.basename))
```
stormwindy, line 69: Do we have a rough estimation how much more requests this would generate? Do we have customers who already manually doing this and we are just giving them a shortcut?

databricks_cli/dbfs/cli.py
```diff
@@ -53,10 +55,20 @@ def ls_cli(api_client, l, absolute, dbfs_path): #  NOQA
...
+
+    recursive_echo(dbfs_path) if recursive else echo_path(dbfs_path)
```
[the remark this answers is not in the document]
justinTM: hi @pietern thank you for your helpful feedback! i believe i fixed this in the latest commit: https://github.com/databricks/databricks-cli/blob/3330fafbb231a23731b8ad5a52ed9aacb17930c1/databricks_cli/dbfs/api.py#L94

databricks_cli/dbfs/cli.py
```diff
@@ -53,10 +55,20 @@ def ls_cli(api_client, l, absolute, dbfs_path): #  NOQA
...
     else:
         error_and_quit('ls can take a maximum of one path.')
-    files = DbfsApi(api_client).list_files(dbfs_path)
-    table = tabulate([f.to_row(is_long_form=l, is_absolute=absolute) for f in files],
-                     tablefmt='plain')
-    click.echo(table)
+    
+    def echo_path(files):
+        table = tabulate([f.to_row(is_long_form=l, is_absolute=absolute) for f in files],
+                         tablefmt='plain')
```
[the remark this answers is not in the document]
justinTM: @pietern thanks again. i believe tabulate will only be called once now after the latest commit, and the max file width should be a single value.

databricks_cli/dbfs/cli.py
justinTM: hi @stormwindy thanks for your review. i truly have no idea how many more requests it will generate; totally dependent on customer file structure. will be 1 extra request for each directory in the path and all of their subdirectories.

stormwindy: I see. What are the use cases for this? Unless there is a crazy deep folder structure this shouldn't cause any problems. Overall I have no other concern.

justinTM: getting filepaths of all worker logs in a cluster: https://github.com/databricks/databricks-cli/issues/512

databricks_cli/dbfs/cli.py
```diff
@@ -53,7 +55,10 @@ def ls_cli(api_client, l, absolute, dbfs_path): #  NOQA
         dbfs_path = dbfs_path[0]
     else:
         error_and_quit('ls can take a maximum of one path.')
-    files = DbfsApi(api_client).list_files(dbfs_path)
+
+    DbfsApi(api_client).list_files(dbfs_path, is_recursive=recursive)
+    absolute = absolute or recursive
+
     table = tabulate([f.to_row(is_long_form=l, is_absolute=absolute) for f in files],
```
stormwindy, line 62: I think you have meant to define a files variable here since we are trying to loop over it? Is there a definition I have missed?

stormwindy (commented):
Overall the idea to add this option is good on our end. I am also happy with the overall implementation direction. I have left one not comment about the code. After that @pietern can decide if/when to approve the PR. Thanks a lot for the contribution.

databricks_cli/dbfs/cli.py
justinTM: fixed with b0fb9cfa9360a87eae5d2c91e6cc076a9a184205

databricks_cli/dbfs/api.py
```diff
@@ -91,8 +91,18 @@ class DbfsApi(object):
     def __init__(self, api_client):
         self.client = DbfsService(api_client)
 
-    def list_files(self, dbfs_path, headers=None):
-        list_response = self.client.list(dbfs_path.absolute_path, headers=headers)
+    def _recursive_list(self, *args, **kwargs):
+        paths = self.client.list_files(*args, **kwargs)
+        files = [p for p in paths if not p.is_dir]
+        for p in paths:
+            files = files + self._recursive_list(p) if p.is_dir else files
```
pietern, line 98: I had to read this a couple times to understand what you're doing. I think it should be simplified to iterate over `paths` just once and then have an `if p.is_dir` in there with both dealing with a file or a directory.

pietern: Separate comment: this also need to pass along the `headers` from the `**kwargs` to be consistent with the `list_files` API today. Otherwise you use it on the first call but not later calls.

tests/dbfs/test_api.py
```diff
@@ -89,15 +102,26 @@ def dbfs_api():
...
+        json = {
+            'files': [TEST_FILE_JSON0, TEST_DIR_JSON, TEST_FILE_JSON1]
+        }
+        dbfs_api.client.list.return_value = json
+        files = dbfs_api.list_files("dbfs:/")
+
+        assert len(files) == 2
+        assert TEST_FILE_INFO0 == files[0]
+        assert TEST_FILE_INFO1 == files[1]
+
```
pietern, line 115: There doesn't seem to be a test for a real recursive call.

The variable `TEST_FILE_JSON2` is unused. When you make a test that uses it and does a real recursive call, I expect it to fail for the reason I commented about above; the mismatch between `FileInfo` and `DbfsPath`.

pietern (commented):
@justinTM Thanks for addressing my earlier comment. You refactored a bunch of the PR and I think there are correctness issues. Please have a look.

chrisst:

PR #448, opened 2022-04-02
Sample - Fix exports from Workspace to Windows FS

stormwindy:

Issue #366, opened 2021-03-12
export_workspace doesn't work with a pipe in the notebook name on Windows


znanfelt:
Steps to reproduce: 
1. create notebook with a "|" character in it in Databricks via the website, for example: "/Shared/sub-folder/sample|notebook"
2. run the following command locally: 

> databricks workspace export_dir /Shared/ c:/temp/

3. Error occurs: 

>ERROR: OSError: [Errno 22] Invalid argument: 'c:/temp\\sub-folder\\sample|notebook.sql'

chrisst:
**This repository has been archived.**

The legacy `databricks/databricks-cli` is deprecated and has been fully replaced by the new Databricks CLI at **https://github.com/databricks/cli**.

We are closing all open issues here as part of archiving this repository. If your issue still affects the new CLI, we'd appreciate you re-opening it at https://github.com/databricks/cli/issues so the team can track and address it there.

PR #516, opened 2022-07-13
Revert "Sample - Fix exports from Workspace to Windows FS (#448)"

pietern:
This reverts commit c695b8db92917f18bf3d6a6a22acc88fe743a3e2.

See https://github.com/databricks/databricks-cli/pull/448#issuecomment-1181692483.

### DIFF
--- a/databricks_cli/workspace/api.py
+++ b/databricks_cli/workspace/api.py
@@ -22,7 +22,6 @@
 # limitations under the License.
 
 import os
-import re
 from base64 import b64encode, b64decode
 
 import click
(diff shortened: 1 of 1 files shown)


pietern:
Thanks, @stormwindy. Is there an issue for this already? If not, could you create one with the exact incompatibility you're thinking about? E.g. is this just about forward vs backward slash or more incompatible characters?

stormwindy:
@pietern There used to be an issue from a contributor. However, I couldn't find it again (it was about downloading job results from what I remember). There is another one recently created issue about characters not being compatible: https://github.com/databricks/databricks-cli/issues/366

stormwindy:
@pietern I created this for following the issue https://github.com/databricks/databricks-cli/issues/517

pietern:
@stormwindy Thank you!

PR #524, opened 2022-07-29
Allow 'new_clusters' key in 'jobs reset' option to mimic JSON API pay…

janvandervegt:
Added check for 'new_clusters' key so that the regular JSON payload for the reset endpoint is also accepted

### DIFF
--- a/databricks_cli/jobs/cli.py
+++ b/databricks_cli/jobs/cli.py
@@ -87,10 +87,14 @@ def reset_cli(api_client, json_file, json, job_id, version):
         with open(json_file, 'r') as f:
             json = f.read()
     deser_json = json_loads(json)
-    request_body = {
-        'job_id': job_id,
-        'new_settings': deser_json
-    }
+    if 'new_settings' in deser_json:
+        request_body = deser_json
+        request_body['job_id'] = job_id
+    else:
+        request_body = {
+            'job_id': job_id,
+            'new_settings': deser_json
+        }
     JobsApi(api_client).reset_job(request_body, version=version)
 
 
--- a/tests/jobs/test_cli.py
+++ b/tests/jobs/test_cli.py
@@ -74,15 +74,24 @@ def test_create_cli_json_file(jobs_api_mock, tmpdir):
 
 RESET_JSON = '{"job_name": "test_job"}'
 
+CORRECT_RESPONSE = {'job_id': 1,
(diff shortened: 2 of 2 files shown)


tests/jobs/test_cli.py
```diff
@@ -74,15 +74,24 @@ def test_create_cli_json_file(jobs_api_mock, tmpdir):
 
 RESET_JSON = '{"job_name": "test_job"}'
 
+CORRECT_RESPONSE = {'job_id': 1,
+                    'new_settings': json.loads(RESET_JSON)}
```
pietern, line 78: This should be `CORRECT_REQUEST_PAYLOAD` if I'm reading the test correctly.

tests/jobs/test_cli.py
```diff
@@ -74,15 +74,24 @@ def test_create_cli_json_file(jobs_api_mock, tmpdir):
...
     runner = CliRunner()
     runner.invoke(cli.reset_cli, ['--json', RESET_JSON, '--job-id', 1])
-    assert jobs_api_mock.reset_job.call_args[0][0] == {
-        'job_id': 1,
-        'new_settings': json.loads(RESET_JSON)
-    }
+    assert jobs_api_mock.reset_job.call_args[0][0] == CORRECT_RESPONSE
+
+
+RESET_JSON_NEW_SETTINGS = '{"new_settings": {"job_name": "test_job"}}'
```
pietern, line 87: Can use `RESET_JSON` to avoid duplication.

databricks_cli/jobs/cli.py
```diff
@@ -87,10 +87,14 @@ def reset_cli(api_client, json_file, json, job_id, version):
         with open(json_file, 'r') as f:
             json = f.read()
     deser_json = json_loads(json)
-    request_body = {
-        'job_id': job_id,
-        'new_settings': deser_json
-    }
+    if 'new_settings' in deser_json:
+        request_body = deser_json
+        request_body['job_id'] = job_id
```
pietern, line 92: What about checking for `new_clusters` as the title implies? If you specify a dict with only that key, it would be wrapped in the `new_settings` field in the else branch. I suspect you may need to check for more than just the `new_settings` key here to determine what is actually specified.

pietern (commented):
@janvandervegt Thanks for the PR, looks reasonable. Please check out the comments.

pietern:
Resolved by #582.

PR #526, opened 2022-07-29
Don't let .netrc take precedence in ApiClient

kartikgupta-db:
- Python requests library automatically picks up `.netrc` configurations for authentication. (https://github.com/psf/requests/issues/2773)
- This PR disables this behaviour to make it consistent with the docs (https://docs.databricks.com/dev-tools/cli/index.html#set-up-authentication) 
- The fix uses a workaround which should be stable until the API for authentication in requests library changes.

### DIFF
--- a/databricks_cli/sdk/api_client.py
+++ b/databricks_cli/sdk/api_client.py
@@ -41,7 +41,6 @@ from . import version
 from requests.adapters import HTTPAdapter
 from requests.utils import get_netrc_auth
 from requests.auth import HTTPBasicAuth
-from requests import PreparedRequest
 from six.moves.urllib.parse import urlparse
 
 try:
@@ -77,7 +76,7 @@ class FallbackNetrcAuth(requests.auth.AuthBase):
         request.Session().auth=FallbackNetrcAuth()
     '''
 
-    def __call__(self, r: PreparedRequest):
+    def __call__(self, r):
         if "Authorization" in r.headers:
             return r
         
--- a/databricks_cli/sdk/api_client.py
+++ b/databricks_cli/sdk/api_client.py
@@ -39,6 +39,9 @@ import pprint
 from . import version
 
 from requests.adapters import HTTPAdapter
+from requests.utils import get_netrc_auth
+from requests.auth import HTTPBasicAuth
+from requests import PreparedRequest
 from six.moves.urllib.parse import urlparse
 
 try:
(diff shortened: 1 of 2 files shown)


fjakobs (commented):
Fix looks reasonable but I'd like to get @pietern's opinion before merging since it would be a breaking change.

pietern:
We can't merge this as is. Reading through the linked issues, there's an opportunity to do this in a non-breaking way. The default auth handler checks the host and overrides a token from `.netrc` if present, regardless of existing authentication being configured. What we can do instead is fallback to reading from `.netrc` if nothing is configured. Then the case where a user has a `.netrc` and only specifies a hostname, the CLI will continue to work.

pietern:
This is a nice workaround. In the body we could check for the `Authorization` header and only defer to `.netrc` if it isn't set.

kartikgupta-db:
> We can't merge this as is. Reading through the linked issues, there's an opportunity to do this in a non-breaking way. The default auth handler checks the host and overrides a token from `.netrc` if present, regardless of existing authentication being configured. What we can do instead is fallback to reading from `.netrc` if nothing is configured. Then the case where a user has a `.netrc` and only specifies a hostname, the CLI will continue to work.

The cli checks that there should be atleast 1 form of accepted configuration methods (spark config, environment variables or databrickscfg) are present. So there should be no cli user directly dependent on a `.netrc`.

For users using the api client directly from the sdk, the fallback should provide them the similar experience as they have now, with the caveat that now other forms of configurations will take precedence (earlier `.netrc` had the highest precedence).  But this is an undocumented use case.

databricks_cli/sdk/api_client.py
```diff
@@ -63,6 +65,28 @@ class TlsV1HttpAdapter(HTTPAdapter):
...
 
+# https://github.com/psf/requests/issues/2773#issuecomment-174312831
+class FallbackNetrcAuth(requests.auth.AuthBase):
+    '''Force requests to ignore the ``.netrc`` if other authentication 
+    methods have been setup. Fallback to ``.netrc`` if not. 
+
+    Use with::
+
+        requests.get(url, auth=FallbackNetrcAuth())
+        request.Session().auth=FallbackNetrcAuth()
```
pietern, line 76: This second example isn't valid Python.

pietern (commented):
Please add a couple tests:

* `.netrc` exists with a valid entry, but auth is specified elsewhere.
* `.netrc` exists with a valid entry, auth is not specified.
* `.netrc` exists with an invalid entry
* `.netrc` does not exist

As is, I'm worried what would happen all cases but the first.

pietern:
> For users using the api client directly from the sdk, the fallback should provide them the similar experience as they have now, with the caveat that now other forms of configurations will take precedence (earlier .netrc had the highest precedence). But this is an undocumented use case.

It's one we need to call out in the changelog / release notes though ;-)

pietern (commented):
What happens if you have a netrc file with garbage / invalid content?

I expect the default behavior in the requests library is to ignore it. Can you confirm this implementation does the same?

pietern (approved):
Looks great, thank you!

PR #527, opened 2022-07-31
Add 'metastores get-assignment' CLI for UC

adamcain-db:
* Adds `unity-catalog metastores get-assignment` (a.k.a. `unity-catalog get-metastore-assignment`) for exercising new UC `getCurrentMetastoreAssignment` API endpoint

### DIFF
--- a/databricks_cli/unity_catalog/metastore_cli.py
+++ b/databricks_cli/unity_catalog/metastore_cli.py
@@ -132,6 +132,19 @@ def metastore_summary_cli(api_client):
     summary_json = UnityCatalogApi(api_client).get_metastore_summary()
     click.echo(mc_pretty_format(summary_json))
 
+@click.command(context_settings=CONTEXT_SETTINGS,
+               short_help='Get metastore assignment for workspace.')
+@debug_option
+@profile_option
+@eat_exceptions
+@provide_api_client
+def get_metastore_assignment_cli(api_client):
+    """
+    Get current metastore assignment for workspace.
+    """
+    assign_json = UnityCatalogApi(api_client).get_current_metastore_assignment()
+    click.echo(mc_pretty_format(assign_json))
+
 
 @click.command(context_settings=CONTEXT_SETTINGS,
                short_help='Assign a metastore to a workspace.')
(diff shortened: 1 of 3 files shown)


databricks_cli/unity_catalog/metastore_cli.py
```diff
@@ -193,6 +206,8 @@ def register_metastore_commands(cmd_group):
     cmd_group.add_command(hide(metastore_summary_cli), name='metastore-summary')
     cmd_group.add_command(hide(assign_metastore_cli), name='assign-metastore')
     cmd_group.add_command(hide(unassign_metastore_cli), name='unassign-metastore')
+    cmd_group.add_command(hide(get_metastore_assignment_cli),
```
wchau, line 209: Do we need this? Based on the comment, these are "deprecated" commands

adamcain-db: No, I just kinda liked the `get-metastore-assignment` version.  Can remove it, if you prefer.

wchau (approved):

adamcain-db:
I'll wait for @pietern to have a chance to review before I merge this.

pietern (approved):
LGTM

@adamcain-db Can you add one or two UC tests to set a precedent? Would be good to ensure new functionality gets tested. For this one it would be good to have a mocked response payload and expectations on the output. It looks like the current approach just dumps the JSON returned by the server, but it might be more in line with expectations to only print the metastore ID.

pietern:
@adamcain-db Please enable commit signing and repush so we can merge.

See https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

adamcain-db:
@pietern:

> It looks like the current approach just dumps the JSON returned by the server, but it might be more in line with expectations to only print the metastore ID.

I think we should keep dumping the full JSON (as done with the other `unity-catalog` commands).  Turns out, the `default_catalog_name` part of the assignment is of major interest to some users (e.g., the Partner Connect folks).

I'll add unit tests in a separate PR.

pietern:
Understood, thanks for clarifying.

PR #530, opened 2022-08-02
Full update options for Delta Sharing shares, recipients, and providers

wchau:
Added support for all updates to shares, recipients, and providers.
Allow option to either specify individual fields to update or JSON.

Tested locally.

### DIFF
--- a/databricks_cli/unity_catalog/api.py
+++ b/databricks_cli/unity_catalog/api.py
@@ -22,6 +22,7 @@
 # limitations under the License.
 
 from databricks_cli.unity_catalog.uc_service import UnityCatalogService
+from databricks_cli.unity_catalog.utils import mc_pretty_format
 
 
 class UnityCatalogApi(object):
@@ -221,9 +222,20 @@ class UnityCatalogApi(object):
     def get_provider(self, name):
         return self.client.get_provider(name)
 
-    def update_provider(self, name, new_name=None, comment=None, recipient_profile=None):
-        return self.client.update_provider(name, new_name, comment, recipient_profile)
-
+    def update_provider(self, name, new_name=None, comment=None,
+                        recipient_profile=None, provider_spec=None):
+        # If spec is provided, override all legacy parameters.
+        if provider_spec is not None:
+            return self.client.update_provider(name, provider_spec)
+        _data = {}
+        if new_name is not None:
+            _data['name'] = new_name
+        if recipient_profile is not None:
+            _data['recipient_profile_str'] = mc_pretty_format(recipient_profile)
+        if comment is not None:
+            _data['comment'] = comment
+        return self.client.update_provider(name, _data)
+        
     def delete_provider(self, name):
         return self.client.delete_provider(name)
 
--- a/databricks_cli/unity_catalog/delta_sharing_cli.py
+++ b/databricks_cli/unity_catalog/delta_sharing_cli.py
@@ -28,7 +28,7 @@ from databricks_cli.configure.config import provide_api_client, profile_option, 
 from databricks_cli.unity_catalog.api import UnityCatalogApi
 from databricks_cli.unity_catalog.utils import hide, json_file_help, json_string_help, \
     mc_pretty_format
-from databricks_cli.utils import eat_exceptions, CONTEXT_SETTINGS, json_cli_base
+from databricks_cli.utils import eat_exceptions, CONTEXT_SETTINGS, json_cli_base, merge_dicts
 
 
 ##############  Share Commands  ##############
@@ -127,6 +127,11 @@ def shared_data_object(name):
                short_help='Update a share.')
 @click.option('--name', required=True,
               help='Name of the share to update.')
+@click.option('--new-name', default=None, help='New name of the share.')
+@click.option('--comment', default=None, required=False,
+              help='Free-form text description.')
+@click.option('--owner', default=None, required=False,
+              help='Owner of the share.')
 @click.option('--add-table', default=None, multiple=True,
               metavar='NAME',
               help='Full name of table to add to share (can be specified multiple times).')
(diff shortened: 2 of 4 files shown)


databricks_cli/unity_catalog/api.py
```diff
@@ -221,9 +222,20 @@ def list_providers(self):
...
+        # If spec is provided, override all legacy parameters.
+        if provider_spec is not None:
+            return self.client.update_provider(name, provider_spec)
+        _data = {}
+        if new_name is not None:
+            _data['name'] = new_name
+        if recipient_profile is not None:
+            _data['recipient_profile_str'] = mc_pretty_format(recipient_profile)
+        if comment is not None:
+            _data['comment'] = comment
```
adamcain-db, line 236: Are we keeping the `new_name`, `comment`, `recipient_profile` args for backward compatibility (in case other apps call this function)?  Looks like our CLI code only calls it with `provider_spec`.

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -141,20 +146,24 @@ def shared_data_object(name):
...
+        if (json_file is not None) or (json is not None):
+            raise ValueError('Cannot specify JSON if any other update flags are specified')
         updates = []
         for a in add_table:
             updates.append({'action': 'ADD', 'data_object': shared_data_object(a)})
         for r in remove_table:
             updates.append({'action': 'REMOVE', 'data_object': shared_data_object(r)})
-        d = {'updates': updates}
-        share_json = UnityCatalogApi(api_client).update_share(name, d)
+        data = {'name': new_name, 'comment': comment, 'owner': owner, 'updates': updates}
```
adamcain-db, line 165: I'm wondering if we should filter out keys that have value `None`.  Probably not needed now, since the server won't see those fields defined. We'll need to revisit this once the UC 'update' APIs allow clearing fields (with FieldMasks or whatever).

adamcain-db (commented):
Basically LGTM.  Just a couple questions.

databricks_cli/unity_catalog/delta_sharing_cli.py
wchau: Yeah, we can do that later, since all the other APIs do something similar along the lines.

databricks_cli/unity_catalog/api.py
wchau: Yes. The API is "public", so the test has been complaining about backwards incompatibility. This is to maintain the existing API.

adamcain-db (approved):
LGTM.  Might want to wait til @pietern or team member can have a look before merging.

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -188,7 +197,7 @@ def delete_share_cli(api_client, name):
 @click.option('--allowed_ip_address', default=None, required=False, multiple=True,
```
pietern, line 197: @wchau Could you change this to use dashes to be consistent with everything else and check if other commands use the same? We shouldn't end up in a situation where folks depend on it.

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -236,6 +245,16 @@ def get_recipient_cli(api_client, name):
                short_help='Update a recipient.')
 @click.option('--name', required=True,
               help='Name of the recipient who needs to be updated.')
+@click.option('--new-name', default=None, help='New name of the recipient.')
+@click.option('--comment', default=None, required=False,
+              help='Free-form text description.')
+@click.option('--owner', default=None, required=False,
+              help='Owner of the recipient.')
+@click.option('--allowed_ip_address', default=None, required=False, multiple=True,
```
pietern, line 253: Same.

databricks_cli/utils.py
```diff
@@ -161,3 +161,14 @@ def for_profile(profile):
              'Please configure by entering '
              '`{argv} configure --profile {profile}`').format(
                 profile=profile, argv=sys.argv[0]))
+
+
+def merge_dicts(*dicts):
+    """
+    Merges dicts throw shallow copy.
```
pietern, line 168: *through

Please rename this to `merge_dicts_shallow` to reflect the shallow nature at the call site.

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -369,34 +401,53 @@ def get_provider_cli(api_client, name):
...
+        if (recipient_profile_json_file is None) and (recipient_profile_json is None):
+            provider_json = UnityCatalogApi(api_client).update_provider(name, provider_spec=data)
+            click.echo(mc_pretty_format(provider_json))
+        else:
+            json_cli_base(recipient_profile_json_file, recipient_profile_json,
+                          lambda profile_json: UnityCatalogApi(api_client).update_provider(
+                              name,
+                              provider_spec=merge_dicts(
+                                  data,
+                                  {'recipient_profile_str': mc_pretty_format(profile_json)})),
```
pietern, line 443: I think you can just set the `recipient_profile_str` key prior to this call. No need for a `merge_dicts` call.

databricks_cli/utils.py
pietern: No need for this function at all, per the other comment.

databricks_cli/unity_catalog/delta_sharing_cli.py
wchau: Done

databricks_cli/unity_catalog/delta_sharing_cli.py
wchau: Done

databricks_cli/unity_catalog/delta_sharing_cli.py
wchau: We can't set this easily without a nested def because it's scoped within a lambda function. With `merge_dicts_shallow`, it can be reused in the future in a easier fashion.

databricks_cli/utils.py
wchau: Per other comment

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -236,6 +246,16 @@ def get_recipient_cli(api_client, name):
...
+@click.option('--new-name', default=None, help='New name of the recipient.')
+@click.option('--comment', default=None, required=False,
+              help='Free-form text description.')
+@click.option('--owner', default=None, required=False,
+              help='Owner of the recipient.')
+@click.option('--allowed-ip-address', default=None, required=False, multiple=True,
+              help=(
+                  'IP address in CIDR notation that is allowed to use delta sharing '
+                  '(can be specified multiple times). Specify a single empty string to disable '
+                  'IP allowlist.'))
```
pietern, line 258: This is a bad API, IMO. If the intent is to clear the allowed IP addresses field to delete them, there should be a dedicated option, e.g. `--clear-allowed-ip-addresses` to do so. Using a magic empty string can lead to incorrect usage where the list is cleared accidentally (e.g. bad interpolation in a script).

pietern (approved):
Thanks for the update. Please consider addressing the concern w.r.t. `--allowed-ip-address` in the update operation in a follow up PR. No need to block a release for this to be addressed.

PR #534, opened 2022-08-11
Change Create Recipient CLI to use global metastore ID

wchau:
Sharing code is not used anymore.

Tested locally.

### DIFF
--- a/databricks_cli/unity_catalog/delta_sharing_cli.py
+++ b/databricks_cli/unity_catalog/delta_sharing_cli.py
@@ -183,8 +183,8 @@ def delete_share_cli(api_client, name):
 @click.option('--name', required=True, help='Name of new recipient.')
 @click.option('--comment', default=None, required=False,
               help='Free-form text description.')
-@click.option('--sharing-code', default=None, required=False,
-              help='A one-time sharing code shared by the data recipient offline.')
+@click.option('--global-metastore-id', default=None, required=False,
+              help='The global UC metastore id provided by the data recipient offline.')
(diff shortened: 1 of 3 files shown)


wchau:
Do we still need to fix public API even if sharing code is useless today?

adamcain-db (commented):
The databricks-cli changes LGTM, though I know nothing of the underlying CreateRecipient API change. Should a delta-sharing person also review?

zhuansunxt (commented):
In the UI we use the term `sharing identifier`. I think the rationale behind that is we want to be more abstract so we can change the format later. Let's keep it consistent in the CLI. (probably too late to migrate REST API)

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -183,8 +183,8 @@ def delete_share_cli(api_client, name):
 @click.option('--name', required=True, help='Name of new recipient.')
 @click.option('--comment', default=None, required=False,
               help='Free-form text description.')
-@click.option('--global-metastore-id', default=None, required=False,
-              help='The global UC metastore id provided by the data recipient offline.')
+@click.option('--sharing-id', default=None, required=False,
+              help='The sharing id provided by the data recipient offline.')
```
zhuansunxt, line 187: nit: "The sharing identifier ...", to be exactly consistent as UI.

zhuansunxt (approved):
Thanks for the fix!

pietern (approved):
We can ignore the API compatibility test as these functions are unlikely to already be in use.

PR #535, opened 2022-08-12
[wip] Run CI on Windows

pietern:
This is a test.


pietern:
Per the logs, Windows only works with Python >= 3.7.9.

PR #546, opened 2022-08-19
Add workflow to automatically bump version

pietern:
This has been a tedious and manual process that should be automated.


pietern:
This is tricky due to not being able to sign automated commits.

PR #549, opened 2022-08-25
Fix bug that disallowed JSON flags in unity-catalog recipients update command

wchau:


pietern:
The flag defaults to `None` though (in the `@click.option` above). Do you see something else?

pietern:
It would be great if you can add a test for this. Looking at the implementation is also does a couple other unexpected things; like update the comment and and owner to nothing if they aren't specified. With a couple tests this could have been caught.

cc @adamcain-db

wchau:
> It would be great if you can add a test for this. Looking at the implementation is also does a couple other unexpected things; like update the comment and and owner to nothing if they aren't specified. With a couple tests this could have been caught.
> 
> cc @adamcain-db

Add tests for all delta sharing CLI commands. Those things are expected; the nulls will not negate the fields.

pietern (approved):
Thanks for adding tests!

> Those things are expected; the nulls will not negate the fields.

Understood. This is not what I would expect though, and what other APIs implement. Typically `null` means remove the field, per  https://datatracker.ietf.org/doc/html/rfc7386.

PR #547, opened 2022-08-25
Add optional region arg to "unity-catalog metastores create"

andrewli81:
Add optional region arg to `databricks unity-catalog metastores create`
DOC ticket: https://databricks.atlassian.net/browse/DOC-5742

The old API was:
```
databricks --profile uc1 unity-catalog metastores create --name andrew_test --storage-root s3://us-east-1-extdev-managed-catalog-test-bucket-1/e280cd28-dced-4e26-b37a-2d14272e55a9
```
User could not specify region when creating a metastore, and region of metastore is always the same region as the workspace's shard. This means user could never create a metastore in a region that doesn't have a CP shard, for example: `us-west-1`, `az-australiasoutheast`. If user has workspace in these regions, they could never assign these workspaces to any metastore due to the metastore <> region matching constraint.

The new API is:
```
databricks --profile uc1 unity-catalog metastores create --name andrew_test --storage-root s3://us-east-1-extdev-managed-catalog-test-bucket-1/e280cd28-dced-4e26-b37a-2d14272e55a9 --region us-west-2
{
  "name": "andrew_test",
  "storage_root": "s3://us-east-1-extdev-managed-catalog-test-bucket-1/e280cd28-dced-4e26-b37a-2d14272e55a9/1abdb18b-1694-42f4-8fdf-82a4ab4aec3e",
  "delta_sharing_scope": "INTERNAL",
  "owner": "andrew.li+admin@databricks.com",
  "privilege_model_version": "1.0",
  "region": "us-west-2",
  "metastore_id": "1abdb18b-1694-42f4-8fdf-82a4ab4aec3e",
  "created_at": 1661398103199,
  "created_by": "andrew.li+admin@databricks.com",
  "updated_at": 1661398103199,
  "updated_by": "andrew.li+admin@databricks.com",
  "cloud": "aws",
  "global_metastore_id": "aws:us-west-2:1abdb18b-1694-42f4-8fdf-82a4ab4aec3e"
}
```
User can specify arbitrary region in the same cloud when creating metastore, and the created metastore will be located in the region user wants. If user doesn't specify the region, it will be the same as the old behavior. 

This was tested by building the CLI locally and creating a metastore with the region field specified. I also tested without region (old behavior):
```
[...]

PR #550, opened 2022-08-26
Disable linting for 'missing-timeout'

pietern:
This was added in pylint==2.15.0 released on 2022-08-26.

It fails in databricks_cli/oauth/oauth.py.


pietern:
This caused linting to fail in checks on #547.

Per https://pypi.org/project/pylint/#history, pylint==2.15.0 was released just 2 hours ago.

PR #565, opened 2022-10-14
Add UPDATE action to delta sharing update share CLI

andyl-db:
This PR adds UPDATE action that was recently added to the Delta Sharing backend.

### DIFF
--- a/databricks_cli/unity_catalog/delta_sharing_cli.py
+++ b/databricks_cli/unity_catalog/delta_sharing_cli.py
@@ -136,6 +136,9 @@ def shared_data_object(name):
 @click.option('--add-table', default=None, multiple=True,
               metavar='NAME',
               help='Full name of table to add to share (can be specified multiple times).')
+@click.option('--update-table', default=None, multiple=True,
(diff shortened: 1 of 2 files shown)


databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -136,6 +136,9 @@ def shared_data_object(name):
 @click.option('--add-table', default=None, multiple=True,
               metavar='NAME',
               help='Full name of table to add to share (can be specified multiple times).')
+@click.option('--update-table', default=None, multiple=True,
```
wchau, line 139: You're not updating anything here, though...

Note that the other add/remove table flags are basic flags without CDF, partitioning, and aliasing options.

wchau (changes requested):

databricks_cli/unity_catalog/delta_sharing_cli.py
andyl-db: Ah oops. I assumed that it would create the object automatically!

Issue #566, opened 2022-10-16
API Client is pointing to non-existent version (v2.1) of "lineage-tracking" API


liamperritt:
API Client is pointing to a non-existent version (v2.1) of the "lineage-tracking" API, although the latest version of this API is still 2.0. This is causing the following error when trying to use the "lineage-tracking" API: `Bad Target: /api/2.1/lineage-tracking/table-lineage/get`

The bug is in this line of code: https://github.com/databricks/databricks-cli/blob/c62d8613b22cbc16f6314a6cdda956dfda84b383/databricks_cli/sdk/api_client.py#L189

The API Client is classifying "lineage-tracking" as a Unity Catalog API (which does have a v2.1), when it is actually a separate API.

liamperritt:
Created Pull Request to fix the above issue: https://github.com/databricks/databricks-cli/pull/567

chrisst:
**This repository has been archived.**

The legacy `databricks/databricks-cli` is deprecated and has been fully replaced by the new Databricks CLI at **https://github.com/databricks/cli**.

We are closing all open issues here as part of archiving this repository. If your issue still affects the new CLI, we'd appreciate you re-opening it at https://github.com/databricks/cli/issues so the team can track and address it there.

PR #567, opened 2022-10-17
fix lineage-tracking api version to point to v2.0

liamperritt:
Fixing bug associated with the following issue: https://github.com/databricks/databricks-cli/issues/566


pietern:
@liamperritt Thanks for posting a fix. This was likely part of the original PR introducing UC support.

@adamcain-db @wchau FYI

PR #582, opened 2022-11-07
Jobs 9434

afberg:


databricks_cli/jobs/cli.py
```diff
@@ -87,9 +87,14 @@ def reset_cli(api_client, json_file, json, job_id, version):
         with open(json_file, 'r') as f:
             json = f.read()
     deser_json = json_loads(json)
+    # If the payload is defined using the API definition rather than the CLI
+    # extract the settings data.
+    json_settings = deser_json['new_settings'] if (
```
KoningJasper, line 92: Should we perhaps also take their `job_id` if provided and make it optional in the `click` annotation?

KoningJasper (approved):

databricks_cli/jobs/cli.py
afberg: Yeah sure, I could add that as well, seems like the right thing to do here

afberg:
Added @mengxr as a reviewer here following [this doc](https://databricks.atlassian.net/wiki/spaces/UN/pages/439517219/Contributing+to+Databricks+CLI), but I imagine the info there might be outdated.

tests/jobs/test_cli.py
```diff
@@ -74,15 +74,33 @@ def test_create_cli_json_file(jobs_api_mock, tmpdir):
...
+
+
+RESET_JSON_NEW_SETTINGS = '{ "new_settings": %s, "job_id": 1 }' % RESET_JSON
+
+
+@provide_conf
+def test_reset_cli_json_new_settings(jobs_api_mock):
+    runner = CliRunner()
+    runner.invoke(cli.reset_cli, [
+                  '--json', RESET_JSON_NEW_SETTINGS, '--job-id', 1])
```
pietern, line 95: This doesn't test if the argument takes precedence or not. It has the same behavior as the test below so both would pass even if you ignore the command line argument altogether. Could you update this to use a different job ID to ensure it doesn't use the job ID from the JSON payload?

tests/jobs/test_cli.py
```diff
@@ -313,7 +331,8 @@ def test_get_job_21(jobs_api_mock):
         runner = CliRunner()
         runner.invoke(cli.get_cli, ['--job-id', '1', '--version', '2.1'])
         assert jobs_api_mock.get_job.call_args == mock.call('1', version='2.1')
-        assert echo_mock.call_args[0][0] == pretty_format(LIST_21_RETURN['jobs'][0])
+        assert echo_mock.call_args[0][0] == pretty_format(
+            LIST_21_RETURN['jobs'][0])
```
pietern, line 335: Can be undone? Or did you see linting problems?

pietern (commented):
Minor comments. LGTM otherwise.

tests/jobs/test_cli.py
afberg: I just ran the linter and I think it updated this, I'll undo it

pietern (approved):
@afberg Thanks!

PR #584, opened 2022-11-10
CLI support for setting Delta Sharing recipient properties

zhuansunxt:
Add `--property` param to both recipient create and update command following the pattern of `allowed_ip_address`

### DIFF
--- a/databricks_cli/unity_catalog/delta_sharing_cli.py
+++ b/databricks_cli/unity_catalog/delta_sharing_cli.py
@@ -471,10 +471,10 @@ def get_recipient_cli(api_client, name):
                   'IP allowlist.'))
 @click.option('--property', 'custom_property', default=None, required=False, multiple=True,
               help=(
-                  'Custom properties of the recipient. Key and value should be provided '
+                  'Properties of the recipient. Key and value should be provided '
                   'at the same time separated by an equal sign. '
-                  'Example: --custom-property country=US. '
-                  'Specify a single empty string to remove all custom properties.'))      
+                  'Example: --property country=US. '
+                  'Specify a single empty string to remove all properties.'))
 @click.option('--json-file', default=None, type=click.Path(),
               help=json_file_help(method='PATCH', path='/recipients/{name}'))
 @click.option('--json', default=None, type=JsonClickType(),
--- a/databricks_cli/unity_catalog/delta_sharing_cli.py
+++ b/databricks_cli/unity_catalog/delta_sharing_cli.py
@@ -394,22 +394,25 @@ def delete_share_cli(api_client, name):
               help=(
                   'IP address in CIDR notation that is allowed to use delta sharing. '
                   '(can be specified multiple times).'))
-@click.option('--property', default=None, type=JsonClickType(), required=False, multiple=True,
+@click.option('--custom-property', default=None, required=False, multiple=True,
               help=(
-                  'Custom properties of the recipient. For format should be '
-                  '{"key": "<property-key>", "value": "<property-value>"}. '
-                  'Example: {"key": "country", "value": "US"}'))                  
+                  'Custom properties of the recipient. Key and value should be provided '
+                  'at the same time separated by an equal sign. '
+                  'Example: --custom-property country=US'))                 
 @debug_option
 @profile_option
 @eat_exceptions
 @provide_api_client
-def create_recipient_cli(api_client, name, comment, sharing_id, allowed_ip_address, property):
+def create_recipient_cli(api_client, name, comment, sharing_id,
+                         allowed_ip_address, custom_property):
     """
     Create a new recipient.
     """
-    properties = [json_loads(property_str) for property_str in property]
+    parsed_custom_properties = [property_str.split('=') for property_str in custom_property]
+    custom_properties = [{"key": property_key, "value": property_value}
+                            for (property_key, property_value) in parsed_custom_properties]
     recipient_json = UnityCatalogApi(api_client).create_recipient(
-        name, comment, sharing_id, allowed_ip_address, properties)
+        name, comment, sharing_id, allowed_ip_address, custom_properties)
     click.echo(mc_pretty_format(recipient_json))
 
 
@@ -457,12 +460,12 @@ def get_recipient_cli(api_client, name):
                   'IP address in CIDR notation that is allowed to use delta sharing '
                   '(can be specified multiple times). Specify a single empty string to disable '
                   'IP allowlist.'))
-@click.option('--property', default=None, type=JsonClickType(), required=False, multiple=True,
+@click.option('--custom-property', default=None, required=False, multiple=True,
               help=(
-                  'Custom properties of the recipient. The format should be '
-                  '{"key": "<property-key>", "value": "<property-value>"}. '
-                  'Example: {"key": "country", "value": "US"}. '
-                  'Specify a single empty object `{}` to remove all properties.'))               
+                  'Custom properties of the recipient. Key and value should be provided '
+                  'at the same time separated by an equal sign. '
+                  'Example: --custom-property country=US'
+                  'Speficy a single empty string to remove all customer properties.'))      
 @click.option('--json-file', default=None, type=click.Path(),
               help=json_file_help(method='PATCH', path='/recipients/{name}'))
 @click.option('--json', default=None, type=JsonClickType(),
@@ -472,14 +475,14 @@ def get_recipient_cli(api_client, name):
 @eat_exceptions
 @provide_api_client
 def update_recipient_cli(api_client, name, new_name, comment, owner,
-                         allowed_ip_address, property, json_file, json):
+                         allowed_ip_address, custom_property, json_file, json):
     """
     Update a recipient.
 
     The public specification for the JSON request is in development.
     """
     if ((new_name is not None) or (comment is not None) or (owner is not None) or
-        len(allowed_ip_address) or len(property) > 0):
+        len(allowed_ip_address) or len(custom_property) > 0):
         if (json_file is not None) or (json is not None):
             raise ValueError('Cannot specify JSON if any other update flags are specified')
         data = {'name': new_name, 'comment': comment, 'owner': owner}
@@ -487,10 +490,14 @@ def update_recipient_cli(api_client, name, new_name, comment, owner,
             data['ip_access_list'] = {}
             if len(allowed_ip_address) != 1 or allowed_ip_address[0] != '':
                 data['ip_access_list']['allowed_ip_addresses'] = allowed_ip_address
-        if len(property) > 0:
+        if len(custom_property) > 0:
             data['properties_kvpairs'] = {}
-            if len(property) != 1 or json_loads(property[0]) != {}:
-                data['properties_kvpairs']['properties'] = [json_loads(property_str) for property_str in property]
+            if len(custom_property) != 1 or custom_property[0] != '':
+                parsed_custom_properties = [property_str.split('=')
+                    for property_str in custom_property]
+                data['properties_kvpairs']['properties'] = [{"key": property_key,
+                                                            "value": property_value}
+                    for (property_key, property_value) in parsed_custom_properties]
         recipient_json = UnityCatalogApi(api_client).update_recipient(name, data)
         click.echo(mc_pretty_format(recipient_json))
     else:
--- a/databricks_cli/unity_catalog/delta_sharing_cli.py
+++ b/databricks_cli/unity_catalog/delta_sharing_cli.py
@@ -405,7 +405,7 @@ def parse_recipient_custom_properties(custom_property_list):
               help=(
                   'IP address in CIDR notation that is allowed to use delta sharing. '
                   '(can be specified multiple times).'))
-@click.option('--custom-property', default=None, required=False, multiple=True,
+@click.option('--property', 'custom_property', default=None, required=False, multiple=True,
               help=(
                   'Custom properties of the recipient. Key and value should be provided '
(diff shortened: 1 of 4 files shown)


zhuansunxt:
@wchau do you know who else should I send for review or is it enough to get reviewed within our team?

wchau:
> @wchau do you know who else should I send for review or is it enough to get reviewed within our team?

Added @pietern

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -394,16 +394,22 @@ def delete_share_cli(api_client, name):
               help=(
                   'IP address in CIDR notation that is allowed to use delta sharing. '
                   '(can be specified multiple times).'))
+@click.option('--property', default=None, type=JsonClickType(), required=False, multiple=True,
```
andyl-db, line 397: Maybe `--property key value` or `--property key=value` would be easier for users than specifying a JSON each time.

https://stackoverflow.com/questions/51164033/python-click-multiple-key-value-pair-arguments

Though the benefit of the JSON is supporting multiple future properties, but are we planning on other properties to the JSON?

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -451,6 +457,12 @@ def get_recipient_cli(api_client, name):
                   'IP address in CIDR notation that is allowed to use delta sharing '
                   '(can be specified multiple times). Specify a single empty string to disable '
                   'IP allowlist.'))
+@click.option('--property', default=None, type=JsonClickType(), required=False, multiple=True,
```
andyl-db, line 460: Same as comment above

databricks_cli/unity_catalog/delta_sharing_cli.py
zhuansunxt: Good idea. I changed it to be separated by the equal sign `key=value` with a caveat that it won't support setting `=` as part of the key or value which I think is probably fine.

andyl-db (approved):
LGTM

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -460,21 +475,29 @@ def get_recipient_cli(api_client, name):
...
             raise ValueError('Cannot specify JSON if any other update flags are specified')
         data = {'name': new_name, 'comment': comment, 'owner': owner}
         if len(allowed_ip_address) > 0:
             data['ip_access_list'] = {}
             if len(allowed_ip_address) != 1 or allowed_ip_address[0] != '':
                 data['ip_access_list']['allowed_ip_addresses'] = allowed_ip_address
+        if len(custom_property) > 0:
+            data['properties_kvpairs'] = {}
+            if len(custom_property) != 1 or custom_property[0] != '':
+                parsed_custom_properties = [property_str.split('=')
```
pietern, line 496: Please limit max splits to 2, so the tuple expansion below doesn't break with more (e.g. `x=y=z`).

pietern (changes requested):
Couple remarks:
* Please name this `properties` to match what it's called at the API level. Why custom?
* Please add a test for passing `x=y=z` to this argument to make it works.
* Nit: please trim trailing whitespace. It adds diffs to future commits.

zhuansunxt:
> Couple remarks:
> 
> * Please name this `properties` to match what it's called at the API level. Why custom?

Initially I named it `property` but that was a keyword redefine. And I thought the name should NOT be of plural form since user are specifying one property at a time? 

> * Please add a test for passing `x=y=z` to this argument to make it works.

Added a test.

> * Nit: please trim trailing whitespace. It adds diffs to future commits.

Dumb question: how do you identify all lines with trailing whitespaces? I can tell by either look at the github diff or the local code.

databricks_cli/unity_catalog/delta_sharing_cli.py
zhuansunxt: Done.

pietern:
Thanks!

Yeah singular is better. Can you elaborate on the keyword redefine issue?

zhuansunxt:
@pietern If I use the name `property` the linter will report an error complaining about `Redefining built-in 'property'`. It looks like some reserved keyword in python.

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -451,6 +460,12 @@ def get_recipient_cli(api_client, name):
                   'IP address in CIDR notation that is allowed to use delta sharing '
                   '(can be specified multiple times). Specify a single empty string to disable '
                   'IP allowlist.'))
+@click.option('--custom-property', default=None, required=False, multiple=True,
+              help=(
+                  'Custom properties of the recipient. Key and value should be provided '
+                  'at the same time separated by an equal sign. '
+                  'Example: --custom-property country=US'
+                  'Speficy a single empty string to remove all customer properties.'))
```
wchau, line 468: Specify

wchau, line 468: remove all custom properties.

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -394,16 +394,25 @@ def delete_share_cli(api_client, name):
...
 @eat_exceptions
 @provide_api_client
-def create_recipient_cli(api_client, name, comment, sharing_id, allowed_ip_address):
+def create_recipient_cli(api_client, name, comment, sharing_id,
+                         allowed_ip_address, custom_property):
     """
     Create a new recipient.
     """
+    parsed_custom_properties = [property_str.split('=', 2) for property_str in custom_property]
+    custom_properties = [{"key": property_key, "value": property_value}
```
wchau, line 412: Should add check that the split actually created 2 entries and respond with appropriate error message if not (eg someone passing in a parameter that does not contain =)

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -460,21 +475,29 @@ def get_recipient_cli(api_client, name):
...
             raise ValueError('Cannot specify JSON if any other update flags are specified')
         data = {'name': new_name, 'comment': comment, 'owner': owner}
         if len(allowed_ip_address) > 0:
             data['ip_access_list'] = {}
             if len(allowed_ip_address) != 1 or allowed_ip_address[0] != '':
                 data['ip_access_list']['allowed_ip_addresses'] = allowed_ip_address
+        if len(custom_property) > 0:
+            data['properties_kvpairs'] = {}
+            if len(custom_property) != 1 or custom_property[0] != '':
+                parsed_custom_properties = [property_str.split('=', 2)
```
wchau, line 496: Also add split check here.

databricks_cli/unity_catalog/delta_sharing_cli.py
zhuansunxt: Added extra check.

databricks_cli/unity_catalog/delta_sharing_cli.py
zhuansunxt: Done.

zhuansunxt:
@wchau @pietern can you take another look?

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -460,21 +479,32 @@ def get_recipient_cli(api_client, name):
...
             raise ValueError('Cannot specify JSON if any other update flags are specified')
         data = {'name': new_name, 'comment': comment, 'owner': owner}
         if len(allowed_ip_address) > 0:
             data['ip_access_list'] = {}
             if len(allowed_ip_address) != 1 or allowed_ip_address[0] != '':
                 data['ip_access_list']['allowed_ip_addresses'] = allowed_ip_address
+        if len(custom_property) > 0:
+            data['properties_kvpairs'] = {}
+            if len(custom_property) != 1 or custom_property[0] != '':
+                data['properties_kvpairs']['properties'] = []
```
pietern, line 500: This logic is duplicated between create and update. Please move it to a function, e.g. parse_properties.

tests/unity_catalog/test_delta_sharing_cli.py
```diff
@@ -527,6 +550,12 @@ def test_update_recipient_cli(api_mock, echo_mock):
                 '8.8.8.8',
                 '8.8.4.4'
             )
+        },
+        'properties_kvpairs': {
+            'properties': [
+                {"key": "k1", "value": "v1"}, 
+                {"key": "k2", "value": "v2"}
+            ]
```
pietern, line 558: For my info, do you know why this is double nested?

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -451,6 +464,12 @@ def get_recipient_cli(api_client, name):
                   'IP address in CIDR notation that is allowed to use delta sharing '
                   '(can be specified multiple times). Specify a single empty string to disable '
                   'IP allowlist.'))
+@click.option('--custom-property', default=None, required=False, multiple=True,
+              help=(
+                  'Custom properties of the recipient. Key and value should be provided '
+                  'at the same time separated by an equal sign. '
+                  'Example: --custom-property country=US'
```
pietern, line 471: This displays as:

```
  --custom-property TEXT     Custom properties of the recipient. Key and value
                             should be provided at the same time separated by
                             an equal sign. Example: --custom-property
                             country=USSpecify a single empty string to remove
                             all custom properties.
```

pietern (changes requested):
I understand the issue with the keyword, but that doesn't mean we have to use "custom" in the name of the argument and the help texts. We should stick to the naming we use in the documentation. The variable can be named whatever works that isn't `property`.

Otherwise a user may have questions such as:
* why can't I set properties, and only custom properties?
* what is the difference between properties and custom properties?

zhuansunxt:
> The variable can be named whatever works that isn't property

@pietern With the constraints of 1) can not override `property` 2) must be singular, `custom-property` is the best name I can think of as the product owner. A recipient can have pre-defined recipients and custom properties, and only the latter can be set by the end users (pre-defined properties are system-generated and can't be mutated). It WILL align with the public documentation we publish at a later point.

databricks_cli/unity_catalog/delta_sharing_cli.py
zhuansunxt: Done.

databricks_cli/unity_catalog/delta_sharing_cli.py
zhuansunxt: Re-formatted.

tests/unity_catalog/test_delta_sharing_cli.py
zhuansunxt: It is because we reuse a common proto message definition on unity catalog service but we need to un-apply `(json_inline) = true` setting in order to allow the client to delete all existing properties. See more in the REST API PR: https://github.com/databricks/universe/pull/227235/files#diff-9d05ccad757e3c0f1a3d894d62996a64b3ab4df3704c0f55290f21f7e0fb2549L85

wchau (approved):

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -383,6 +383,17 @@ def delete_share_cli(api_client, name):
 
 ##############  Recipient Commands  ##############
 
+def parse_recipient_custom_properties(custom_property_list):
+    custom_properties = []
+    for property_str in custom_property_list:
+        tokens = property_str.split('=', 2)
+        if len(tokens) != 2:
+            raise ValueError('Invalid format of custom property. '
```
pietern, line 391:
```suggestion
            raise ValueError('Invalid format for property. '
```

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -394,16 +405,23 @@ def delete_share_cli(api_client, name):
               help=(
                   'IP address in CIDR notation that is allowed to use delta sharing. '
                   '(can be specified multiple times).'))
+@click.option('--property', 'custom_property', default=None, required=False, multiple=True,
+              help=(
+                  'Custom properties of the recipient. Key and value should be provided '
+                  'at the same time separated by an equal sign. '
+                  'Example: --custom-property country=US.'))
```
pietern, line 412:
```suggestion
                  'Properties of the recipient. Key and value should be provided '
                  'at the same time separated by an equal sign. '
                  'Example: --property country=US.'))
```

databricks_cli/unity_catalog/delta_sharing_cli.py
```diff
@@ -451,6 +469,12 @@ def get_recipient_cli(api_client, name):
                   'IP address in CIDR notation that is allowed to use delta sharing '
                   '(can be specified multiple times). Specify a single empty string to disable '
                   'IP allowlist.'))
+@click.option('--property', 'custom_property', default=None, required=False, multiple=True,
+              help=(
+                  'Custom properties of the recipient. Key and value should be provided '
+                  'at the same time separated by an equal sign. '
+                  'Example: --custom-property country=US. '
+                  'Specify a single empty string to remove all custom properties.'))
```
pietern, line 477:
```suggestion
                  'Properties of the recipient. Key and value should be provided '
                  'at the same time separated by an equal sign. '
                  'Example: --property country=US. '
                  'Specify a single empty string to remove all properties.'))
```

pietern (commented):
Argument help was lagging the argument name. Replaced in suggestions.

pietern (approved):

PR #595, opened 2022-12-27
Make use of the new name filtering in calls to list jobs.

dinowernli:
Tried this out manually:

```
> databricks jobs list --name=test_continuous --version=2.1
371294009417906  test_continuous
```

And with 2.0 instead of 2.1:

```
> databricks jobs list --name=test_continuous
WARN: Your CLI is configured to use Jobs API 2.0. In order to use the latest Jobs features please upgrade to 2.1: 'databricks jobs configure --version=2.1'. Future versions of this CLI will default to the new Jobs API. Learn more at https://docs.databricks.com/dev-tools/cli/jobs-cli.html
ERROR: the options --expand-tasks, --offset, --limit, --all, and --name are only available in API 2.1
```

### DIFF
--- a/databricks_cli/jobs/api.py
+++ b/databricks_cli/jobs/api.py
@@ -32,9 +32,10 @@ class JobsApi(object):
                                                 version=version)
 
     def list_jobs(self, job_type=None, expand_tasks=None, offset=None, limit=None, headers=None,
-                  version=None):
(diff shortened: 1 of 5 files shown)


KoningJasper (approved):

lennartkats-db (approved):
Please note that you need to enable commit signing for this repository.

PR #614, opened 2023-03-13
Add missing fields to JobsService

pietern:
Adds the `webhook_notifications` field to `JobsService`.

This change also renames the argument `name_filter` to `name` to match the API.

FYI @dinowernli


dinowernli (approved):
LGTM

Issue #636, opened 2023-05-04
DEFAULT_METHOD_WHITELIST does not exist anymore


kparaju:
DEFAULT_METHOD_WHITELIST does not exist anymore and it causes the following issue with `dbx deploy`
```
│ /home/circleci/.pyenv/versions/3.9.16/lib/python3.9/site-packages/databricks │
│ _cli/sdk/api_client.py:106 in __init__                                       │
│                                                                              │
│   103 │   │   │   total=6,                                                   │
│   104 │   │   │   backoff_factor=1,                                          │
│   105 │   │   │   status_forcelist=[429],                                    │
│ ❱ 106 │   │   │   method_whitelist=set({'POST'}) | set(Retry.DEFAULT_METHOD_ │
│   107 │   │   │   respect_retry_after_header=True,                           │
│   108 │   │   │   raise_on_status=False # return original response when retr │
│   109 │   │   )                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
AttributeError: type object 'Retry' has no attribute 'DEFAULT_METHOD_WHITELIST'
```

This codebase [references](https://github.com/databricks/databricks-cli/blob/5c3879188322c0dfe94701db1ec8299bad7ccd5f/databricks_cli/sdk/api_client.py#LL106C56-L106C80
) `Retry.DEFAULT_METHOD_WHITELIST` but it doesn't exist anymore. It was [removed](https://github.com/urllib3/urllib3/commit/382ab32f23795c44faae83b4e8b18a16fb605a0a) from urllib3.

> Using 'Retry.DEFAULT_METHOD_WHITELIST' is deprecated and 
> will be removed in v2.0. Use 'Retry.DEFAULT_METHODS_ALLOWED' instead

Locking the version for urllib3 fixes the issue:
```
pip install dbx urllib3==1.26.15
```

bigdatamoore:
I am getting the same error.

jonsnowseven:
Me too. I solved it using `pip install 'urllib3<2'`.

Requests 2.30.0 release specifies possible breaking changes: https://requests.readthedocs.io/en/latest/community/updates/#id1

chrimaho:
Same error for me too.

The error is occurring in the line 106 of the [databricks_cli/sdk/api_client.py](https://github.com/databricks/databricks-cli/blob/5c3879188322c0dfe94701db1ec8299bad7ccd5f/databricks_cli/sdk/api_client.py) file.

See:

[...]

PR #637, opened 2023-05-05
Fix api client by switching to replacements for deprecated urllib3 methods

shreyas-goenka:
Recently with the release of urllib3 v2.0.0 some deprecated attributes were removed. This PR uses there new replacements and pins the urllib3 version to when the replacements became available

We need to bound the urllib3 version because this issue has not been resolved on the `requests` package side as well and the requests package enforces `urllib3 v1.21 or greater` which does not contain the methods we need

This should fix https://github.com/databricks/databricks-cli/issues/636

Tested manually by:

1. pip3 install requirements.txt
2. python3 databricks_cli/cli.py fs ls dbfs:/

Before:
```
shreyas.goenka@THW32HFW6T databricks-cli % python3 databricks_cli/cli.py fs ls dbfs:/
Error: AttributeError: type object 'Retry' has no attribute 'DEFAULT_METHOD_WHITELIST'
```

After (expected response):
```
shreyas.goenka@THW32HFW6T databricks-cli % python3 databricks_cli/cli.py fs ls dbfs:/
dir1
dir2
dir3
...
```

### DIFF
--- a/databricks_cli/sdk/api_client.py
+++ b/databricks_cli/sdk/api_client.py
@@ -103,7 +103,7 @@ class ApiClient(object):
             total=6,
             backoff_factor=1,
             status_forcelist=[429],
-            method_whitelist=set({'POST'}) | set(Retry.DEFAULT_METHOD_WHITELIST),
+            allowed_methods=set({'POST'}) | set(Retry.DEFAULT_ALLOWED_METHODS),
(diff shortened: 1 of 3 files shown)


mgyucht (approved):

ViktorDronov:
this fix does not solve the issue, because no new releases are published

PR #650, opened 2023-06-16
Add trampoline into newer CLI versions if found in $PATH

pietern:


pietern:
The try/except body in `main()` was inspired by https://github.com/pallets/click/blob/d0af32d8e7d78f4fdaf9c3d1559e1ffaac1b2569/src/click/core.py#L1070-L1116.

databricks_cli/cli.py
```diff
@@ -70,5 +74,119 @@ def cli():
...
 cli.add_command(unity_catalog_group, name='unity-catalog')
 
+
+def _trampoline_into_new_cli():
+    trampoline_disable_env_var = 'DATABRICKS_CLI_DO_NOT_EXECUTE_NEWER_VERSION'
+    if os.environ.get(trampoline_disable_env_var) is not None:
+        return
+
+    # Try to trampoline only if we're running the CLI as 'databricks'.
+    if os.path.basename(sys.argv[0]) != 'databricks':
```
andrewnester, line 84: Do we want to different between running as a full path and just command name?

pietern: I'd like to but unfortunately, this is always set to the full path if it was resolved by looking at $PATH.

We can add an additional check to see if the path is absolute as well.

Then you can still call it as `bin/databricks` or something and it doesn't show the warning.

pietern: TBH, for consistency it's probably fine to keep the check as is. If folks intend to run it they can use the environment variable to force it to and silence the error.

andrewnester (approved):

databricks_cli/cli.py
```diff
@@ -70,5 +74,119 @@ def cli():
...
+        if not self:
+            self = exec_path
+
+        # The new CLI is a single binary larger than 1MB.
+        # We use this as heuristic to tell if the new CLI is installed.
+        # Because this version of the CLI is much smaller, we do not
+        # need to dedup our own path to avoid an exec loop.
+        stat = os.stat(exec_path)
+        if stat.st_size < (1024 * 1024):
+            continue
```
mgyucht, line 110: I hope we don't break this...

If we're already loading the version of the CLI below, why don't we just call `version` anyways and get the result, breaking if we find something with version >= "0.100"?

databricks_cli/cli.py
```diff
@@ -70,5 +74,119 @@ def cli():
...
+    e("", nl=True)
+
+    e("Executing CLI ")
+    e("v{}".format(candidate_version), highlight=True)
+    e("...", nl=True)
+    e("-" * (len("Executing CLI v{}...".format(candidate_version))), nl=True)
+
+    # The new CLI is at least 1MB in size. This is a good heuristic to
+    # determine if the new CLI is installed.
+    os.execv(exec_path, sys.argv)
```
mgyucht, line 160: This is still probably safe, but maybe either use `candidate` here or get rid of `candidate` everywhere and assume that if we haven't returned, `exec_path` is the candidate? Otherwise, no need for two variables with the same value.

mgyucht (approved):
Some small nits/suggestions but LGTM

databricks_cli/cli.py
pietern: The version output in JSON only works on the new CLI, so if we remove this check we could be exec'ing into ourselves. The binary comes in at a decent 20MB so I think it's safe.

Are you thinking of another failure mode?

databricks_cli/cli.py
pietern: Good catch, yeah it should be candidate, even though they happen to be the same b/c of the break.

databricks_cli/cli.py
mgyucht: Ah, I see... I was suggesting replacing this check with the version check. The other failure mode is if the python binary grows in size somehow.

PR #673, opened 2023-10-04
Remove imp dependency and test on python 3.12

mgyucht:
To enable users to use the databricks-cli library with Python 3.12, we must remove our use of the `imp` library. This is done in this PR as well as adding 3.12 to unit test coverage to ensure the package can be installed & used with Python 3.12.


.github/workflows/push.yml
```diff
@@ -62,6 +62,7 @@ jobs:
       matrix:
         python-version:
           - '3.7'
+          - '3.12'
```
pietern, line 65: We could also run this on all versions we support to be sure.

pietern (approved):

pietern:
@mgyucht We might need to update the runner image to `ubuntu-22.04` to make it work.

jmahlik (commented):
Currently there are only wheels for python 2 on pypi. 

https://pypi.org/project/databricks-cli/#files

Can the [universal](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#wheels) option be added to the `setup.py`? This will produce a python 2 and 3 compatible wheel.

```python
setup(
...,
options={'bdist_wheel': {'universal': True}},
)
```

Current -- users get the sdist, which causes setup.py invocation, leading to this issue.

```

databricks_cli-0.17.8-py2-none-any.whl
databricks-cli-0.17.9.dev0.tar.gz

```

With universal - python 2 and 3 users get a wheel which is faster to install and doesn't invoke the setup.py.

```
databricks_cli-0.17.9.dev0-py2.py3-none-any.whl
databricks-cli-0.17.9.dev0.tar.gz
```



Also related to https://github.com/databricks/databricks-cli/issues/631.

mgyucht:
@jmahlik Yes, we've discussed and will do that in a follow-up PR.

PR #674, opened 2023-10-04
Require Python 3.7 or later

pietern:
* Update every mention of a Python version to >= 3.7
* Remove `dev-requirements-py2.txt`
* Run actions on all supported Python versions
* Update GH actions runners to `ubuntu-22.04`


mgyucht (approved):
Great, thanks for carefully going through this!

PR #675, opened 2023-10-04
Add deprecation warning

mgyucht:
This PR adds a deprecation warning with a call-to-action for users to move to https://github.com/databricks/databricks-sdk-py.

### DIFF
--- a/databricks_cli/__init__.py
+++ b/databricks_cli/__init__.py
@@ -24,6 +24,13 @@
 # pylint:disable=import-error
 # pylint:disable=bare-except
 
+import warnings
+
+warnings.warn("the databricks-cli module is deprecated in favor of databricks-sdk-py. Python 3.12 will be the last "
+              "version of Python supported by databricks-cli. Please migrate to databricks-sdk-py as documented in "
+              "the migration guide: https://docs.databricks.com/en/dev-tools/cli/migrate.html",
+              DeprecationWarning, stacklevel=2)
+
 
 def initialize_cli_for_databricks_notebooks():
     import IPython
--- a/databricks_cli/__init__.py
+++ b/databricks_cli/__init__.py
@@ -28,7 +28,12 @@ import warnings
 
 
 def issue_deprecation_warning():
-    warnings.warn("the databricks-cli module is deprecated in favor of databricks-sdk-py. Python "
+    # Don't print deprecation warning when running the CLI itself.
+    import sys
+    import os
+    if sys.argv and os.path.basename(sys.argv[0]) == 'databricks':
+        return
+    warnings.warn("the databricks_cli module is deprecated in favor of databricks-sdk-py. Python "
                   "3.12 will be the last version of Python supported by databricks-cli. Please "
                   "migrate to databricks-sdk-py as documented in the migration guide: "
                   "https://docs.databricks.com/en/dev-tools/cli/migrate.html",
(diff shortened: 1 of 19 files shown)


databricks_cli/__init__.py
```diff
@@ -24,6 +24,17 @@
 # pylint:disable=import-error
 # pylint:disable=bare-except
 
+import warnings
+
+
+def issue_deprecation_warning():
+    warnings.warn("the databricks-cli module is deprecated in favor of databricks-sdk-py. Python "
```
pietern, line 31:
```suggestion
    warnings.warn("the databricks_cli module is deprecated in favor of databricks-sdk-py. Python "
```

databricks_cli/__init__.py
```diff
@@ -24,6 +24,17 @@
...
+                  "3.12 will be the last version of Python supported by databricks-cli. Please "
+                  "migrate to databricks-sdk-py as documented in the migration guide: "
+                  "https://docs.databricks.com/en/dev-tools/cli/migrate.html",
+                  DeprecationWarning, stacklevel=3)
```
pietern, line 35: What does the stacklevel mean?

pietern: Ideally, if folks use the Databricks CLI itself, they don't see these.

mgyucht: stacklevel changes the output of warnings.warn. Normally, the warning message starts with the line that the warning.warn call is on, but increasing stacklevel allows the warning to point to the place where the caller uses the library/function in question. In this case, the warning will point to the line with `import databricks_cli`:
Screenshot 2023-10-05 at 10 41 20

mgyucht:
When I run this databricks binary, I see that the deprecation warning is not printed, but it is when imported.
Screenshot 2023-10-05 at 10 41 50

databricks_cli/__init__.py
```diff
@@ -24,6 +24,22 @@
...
+def issue_deprecation_warning():
+    # Don't print deprecation warning when running the CLI itself.
+    import sys
+    import os
+    if sys.argv and os.path.basename(sys.argv[0]) == 'databricks':
+        return
+    warnings.warn("the databricks_cli module is deprecated in favor of databricks-sdk-py. Python "
+                  "3.12 will be the last version of Python supported by databricks-cli. Please "
+                  "migrate to databricks-sdk-py as documented in the migration guide: "
+                  "https://docs.databricks.com/en/dev-tools/cli/migrate.html",
```
pietern, line 39: This migration guide doesn't mention the SDK, only the CLI.

We should link to SDK docs here, assuming callers are trying to leverage this code as SDK.

databricks_cli/__init__.py
```diff
@@ -24,6 +24,22 @@
...
 
+import warnings
+
+
+def issue_deprecation_warning():
+    # Don't print deprecation warning when running the CLI itself.
+    import sys
+    import os
+    if sys.argv and os.path.basename(sys.argv[0]) == 'databricks':
+        return
```
pietern, line 35: Could we add a check here to only fire the warning once?

I ran `python3 -m databricks_cli.cli` and got a spew of a dozen identical warnings.

databricks_cli/__init__.py
mgyucht: Hm, we never made a migration guide for the Python SDK...

chrisst:

PR #676, opened 2023-10-05
Add bdist universal wheel

mgyucht:
We can support a universal wheel for the build. This means that users won't need to install a source distribution & run setup.py; they can just download the wheel artifact for use.


mgyucht:
cc @jmahlik

pietern (approved):

